PR #65, opened 2026-05-03
feat: AES-256-GCM column-level encryption for user PII

AndresL230:
## Summary

Adds AES-256-GCM column-level encryption for user PII, OAuth tokens, assignment notes/points, document summaries/concept notes, session summaries, and chat messages. Includes encrypt-on-write + decrypt-on-read across all affected routes, a schema migration that retypes encrypted columns to TEXT, and an idempotent backfill script for existing rows.

### 1. What to apply to Supabase

Run one SQL file in the Supabase SQL editor: `backend/db/migration_encryption_text_columns.sql`. It contains four `ALTER TABLE ... ALTER COLUMN ... TYPE TEXT USING ::TEXT` statements.

Apply it once, before deploying the new code. The casts preserve existing rows as TEXT (e.g., `87.5` becomes the string `"87.5"`); the `decrypt_if_present` / `decrypt_numeric` fallbacks keep those legacy rows readable until you backfill.

### 2. Values you need to set

Just one new env var. `ENCRYPTION_KEY` = 64 hex chars (32 bytes). Generate it with:

```bash
python -c "import secrets; print(secrets.token_hex(32))"
```

Set it in:

- `backend/.env` for local dev (`backend/.env.example` already has the placeholder + the generation command).
- Wherever your backend actually runs in deploy (Cloudflare Workers/Pages env vars, Vercel, Fly, Railway, whatever — the same place `SUPABASE_SERVICE_KEY` is set today). `docker-compose.yml` already passes it through from `backend/.env`.

Two non-negotiables:

- **The same key must be used everywhere your backend runs.** Different keys = ciphertext written by one instance is unreadable by another.
- **Never rotate or lose this key.** Losing it permanently bricks every encrypted column. If you need to rotate, write a script that decrypts with the old key and re-encrypts with the new one in one transaction.

The backend will refuse to boot if `ENCRYPTION_KEY` is missing, not 64 chars, or not valid hex (intentional fail-fast).

### 3. The backfill script

`backend/db/backfill_encryption.py`. For each encrypted column on each table (`users.name`/`email`/`bio`/..., `oauth_tokens.access_token`/`refresh_token`, `assignments.notes`/`points_*`, `documents.summary`/`concept_notes`, `sessions.summary_json`, `messages.content`, `room_messages.text`), it:

1. Selects every row.
2. For each cell, calls `decrypt(value)` once.
3. If it succeeds → already encrypted → skip.
4. If it raises → the cell is still legacy plaintext → encrypt it and write back.

Prints counts per column. Idempotent — running it a second time changes nothing. Defaults to dry-run (counts only); `--apply` to actually write. `--table users` to do it in stages.

## How to actually run all of this, in order

```bash
# 1. Generate and set the key
python -c "import secrets; print(secrets.token_hex(32))"
# paste output as ENCRYPTION_KEY=... in backend/.env
# AND set the same value in your deploy environment

# 2. Apply the schema migration in Supabase
#    Dashboard → SQL Editor → paste contents of:
#    backend/db/migration_encryption_text_columns.sql → Run

# 3. Deploy the new backend code (encryption wiring is on this branch)

# 4. Backfill — first dry-run to see what will change:
cd backend
.\venv\Scripts\python.exe -m db.backfill_encryption

# 5. Then actually apply:
.\venv\Scripts\python.exe -m db.backfill_encryption --apply
```

After the backfill completes successfully, the `decrypt_if_present` fallback-to-raw-value warnings should stop appearing in logs. That's the signal it worked.

## Test plan

- [x] Apply `migration_encryption_text_columns.sql` to a Supabase branch DB
- [ ] Set `ENCRYPTION_KEY` (64-hex) and confirm backend boots; confirm it refuses to boot when the var is missing/invalid
- [x] Run backend tests: `cd backend && python -m pytest tests/ -q` (includes new `test_encryption.py`)
- [ ] Sign in via Google OAuth → confirm `users.name`/`email` written as ciphertext, `/me` returns plaintext
- [ ] Upload a document → confirm `documents.summary`/`concept_notes` ciphertext at rest, decrypt on library/study-guide/flashcards reads
- [ ] Sync calendar → confirm `oauth_tokens.access_token`/`refresh_token` and `assignments.notes`/`points_*` encrypted
- [ ] Send a room message + a learn-session message → confirm `room_messages.text` and `messages.content` encrypted
- [ ] Run backfill in dry-run on a copy of prod data, verify counts, then `--apply` and re-run to confirm idempotency (zero changes)
- [ ] Confirm `decrypt_if_present` fallback warnings disappear from logs after backfill

## Summary by CodeRabbit

* **New Features**
  * End-to-end encryption for sensitive data (profiles, documents, sessions, messages, tokens).

* **Improvements**
  * Decryption in responses and encryption on save for many user-facing flows.
  * Tighter per-request authorization across endpoints and stricter ownership checks.

* **Chores**
  * Added encryption key configuration and dependency updates; adjusted schema to support encrypted values.

* **Tests & Docs**
  * Added encryption test coverage and an implementation/rollout plan.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -9,6 +9,10 @@ FRONTEND_URL=http://localhost:3000
 SUPABASE_URL=https://your-project-ref.supabase.co
 SUPABASE_SERVICE_KEY=your-service-role-key-here
 
+# Column-level encryption (AES-256-GCM). 32 bytes as 64 hex chars.
+# Generate with: python -c "import secrets; print(secrets.token_hex(32))"
+ENCRYPTION_KEY=
+
 # OCR pipeline
 # OCR_ENGINE: "docling" (default, layout-aware), "auto" (Docling + GOT-OCR fallback for math), "tesseract" (legacy)
 OCR_ENGINE=docling
--- a/backend/db/backfill_encryption.py
+++ b/backend/db/backfill_encryption.py
@@ -1,0 +1,340 @@
+"""One-shot backfill: re-encrypt every legacy plaintext row across the
+columns wired for AES-256-GCM column-level encryption.
+
+Why this exists
+---------------
+The decrypt helpers (services.encryption.decrypt_if_present /
+decrypt_json / decrypt_numeric) silently fall back to the raw value when
+a column hasn't been encrypted yet. That keeps the app working through
+the rollout, but every legacy read logs a warning. This script walks
+every encrypted column, detects which rows are still plaintext, and
+rewrites them through the encrypt helpers. After it finishes once,
+fallback warnings should stop.
+
+Detection rule
+--------------
+For every cell, try `decrypt(value)`:
+    - succeeds → already encrypted, skip
+    - fails    → plaintext (or corrupted); encrypt and write
+
+This is idempotent: re-running it after a successful run is a no-op
+(every cell is already valid ciphertext, so every check skips).
+
+Usage
+-----
+Dry run (default — counts only, no writes):
+    python -m db.backfill_encryption
+
+Apply (writes to Supabase):
+    python -m db.backfill_encryption --apply
+
+Per-table only (useful for staged rollouts):
+    python -m db.backfill_encryption --apply --table users
+    python -m db.backfill_encryption --apply --table assignments
+
+Run from the `backend/` directory with the venv active so config + the
+encryption module + the supabase client all resolve correctly.
+"""
+from __future__ import annotations
+
+import argparse
+import sys
+from typing import Any, Callable
+
+from db.connection import table
+from services.encryption import (
+    decrypt,
+    decrypt_json,
+    encrypt,
+    encrypt_if_present,
+    encrypt_json,
+)
+
+
+# ── Detection ─────────────────────────────────────────────────────────────────
+
+def _is_already_encrypted(value: Any) -> bool:
+    """True if `value` is a valid base64 AES-GCM ciphertext under the
+    current ENCRYPTION_KEY. False for None, non-strings, plaintext,
+    or ciphertext under a different key."""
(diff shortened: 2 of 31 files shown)


AndresL230:
### Code review

Found 2 issues:

1. `resume_session` returns encrypted `messages.content` to the client without decrypting. `save_message` now writes `encrypt_if_present(content)` and `get_conversation_history` correctly applies `decrypt_if_present`, but `resume_session` selects `id,role,content,created_at` from the `messages` table and returns the rows as-is — users resuming a session will see base64 ciphertext instead of message text.

https://github.com/SaplingLearn/Sapling/blob/ba27384363f772d873a807c91bfcf6ac24ac794b/backend/routes/learn.py#L524-L533

2. The `/auth/me` endpoint signature changed to require a verified session (`get_session_user_id(request)` — accepts only an `auth_token` query param or `sapling_session` cookie), but `frontend/src/middleware.ts` still calls `${API_URL}/api/auth/me?user_id=...` from edge middleware without forwarding either credential. Edge `fetch()` does not auto-forward browser cookies, so every protected-route navigation will receive `401 Not authenticated` and fall into the `redirectToSignin(request, 'session_expired')` branch — full sign-in regression for approved users. The same `?user_id=...` pattern is also used by `frontend/src/app/api/auth/session/route.ts` (fallback path) and `frontend/src/context/UserContext.tsx`.

https://github.com/SaplingLearn/Sapling/blob/ba27384363f772d873a807c91bfcf6ac24ac794b/frontend/src/middleware.ts#L43-L62

https://github.com/SaplingLearn/Sapling/blob/ba27384363f772d873a807c91bfcf6ac24ac794b/backend/services/auth_guard.py#L16-L51

PR #94, opened 2026-05-05
fix(graph): resolve color fallback so SVG nodes don't render black

Jose-Gael-Cruz-Lopez:
## Summary
- SVG presentation attributes (`fill=`) don't resolve CSS custom properties — when the backend doesn't supply a per-course color, the `var(--c-sage)` fallback was rendering black.
- Adds a deterministic hex palette + `paletteFor(seed)` helper in `lib/data.ts` and wires it into the Dashboard, Tree, and KnowledgeGraph fallback paths.
- Bumps subject-root node radius from 22 → 28 so course family centers read distinctly from concept nodes.

## Test plan
- [x] `npx tsc --noEmit` in `frontend/` — clean
- [ ] Manual: load Dashboard / Tree with a course that has no `course_color` and confirm nodes render with a stable hue (not black)
- [ ] Manual: confirm subject-root nodes look noticeably larger than concept nodes
- [ ] Manual: same course id always maps to the same palette entry across screens

## Note
A sibling fallback exists in `screens/Learn.tsx#apiToGraphNode`. It's intentionally left untouched here because PR #93 is in flight on that file. Once both land, a one-liner can switch Learn's fallback to `paletteFor(...)` for consistency.

### DIFF
--- a/frontend/src/components/KnowledgeGraph.tsx
+++ b/frontend/src/components/KnowledgeGraph.tsx
@@ -219,10 +219,12 @@ function KnowledgeGraphImpl({
   // ── Helpers ──────────────────────────────────────────────────────────────
   const masteryOpacity = (tier: GraphNode["mastery_tier"]) =>
     ({ mastered: 1, learning: 0.78, struggling: 0.55, unexplored: 0.28 })[tier] || 0.6;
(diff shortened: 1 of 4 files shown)


Jose-Gael-Cruz-Lopez:
Folded into #93 — keeping rename feature and graph color fix in a single PR.

PR #241, opened 2026-06-18
Gradebook Feature

Darkest-Teddy:
## Description
Major Gradebook feature release. Adds a bell curve grading system, a grade predictor panel, category tab navigation, and a large set of UI polish across the course page.

## Changes Made
- **Bell curve grading**: `apply_curve` backend function, per-assignment class stats (entered via assignment modal), course-level curve policy (avg target + SD delta), Raw/Curved toggle persisted per-course, curved score chip on assignment rows
- **Grade predictor panel**: expandable panel below composition bar with per-assignment hypothetical sliders, Raw/Curved toggle inside predictor, class stats override for curve computation
- **Grade projector & composition bar**: instant update on Raw/Curved toggle, `isPredicted` visual mode, `computeCurrentGrade` mirrors backend logic exactly
- **Letter scale fix**: frontend DEFAULT_SCALE updated to match backend full 12-tier A+/A-/B+ scale; floating-point boundary fix (round to 1dp before comparisons)
- **Category navigation**: horizontal tab strip above assignment list with hover states and assignment counts
- **Assignment rows**: hover highlight extends full width, clean straight dividers, Dropped/Curved chips inline with title, tightened fraction display (30/34 format)
- **Category colors**: 10 distinct hues evenly spaced around the hue wheel, no repeats
- **Auth fix**: `SECURE_COOKIES` env flag, `SESSION_SECRET` wired to frontend, `secure=false` on localhost for OAuth cookies
- **Drag-to-reorder**: categories in Edit Weights modal support drag-and-drop reordering
- **Drop policy**: renamed "drops x/x" → "N Drops"
- **Course name**: no longer truncated with ellipsis

## Related Issues
Closes #

## Testing
- [ ] Tested locally
- [ ] Added/updated tests

## Notes for Reviewers
Schema changes ship as ordered migrations (`backend/db/migrations/0019`–`0021`) applied by the migration runner (`backend/db/migrate.py --apply`, run from `backend/`) — **no manual Supabase SQL-editor steps required**. The runner is idempotent and records applied versions in `schema_migrations`. The migrations cover the bell-curve columns (`assignments.curve_*`, `user_courses.curve_*`), `course_categories.drop_lowest`, and the Gradescope tables/columns.

_(The previous manual `ALTER TABLE` block here was stale — it omitted `drop_lowest` and the Gradescope schema — and has been removed in favor of the runner.)_

## Summary by CodeRabbit

* **New Features**
  * Added Gradescope account connection, course linking, and sync support from Settings.
  * Added course-level curve settings and per-category “drop lowest” grading support.
  * Introduced grade prediction tools for ungraded assignments and curved projections.
  * Added a dedicated Connected Accounts page.

* **Bug Fixes**
  * Improved cookie handling for login/session flows across secure and non-secure environments.
  * Refined grading calculations and grade display behavior for curved, dropped, and predicted scores.


backend/services/gradebook_service.py
```diff
@@ -41,38 +42,125 @@ class AssignmentRow(TypedDict, total=False):
...
+                    and item_avg is not None and item_sd_delta is not None):
+                raw_pct = apply_curve(
+                    float(raw_pct), float(item_mean), float(item_sd),
+                    float(item_avg), float(item_sd_delta)
+                )
+        scores.append(raw_pct)
+    if not scores:
         return None
-    return total_earned / total_possible
+    return sum(scores) / len(scores)
```
AndresL230: 🔴 **Breaking grade-math change (please confirm it's intended).** This is now an *unweighted mean of per-assignment percentages* (`sum(scores)/len(scores)`), but `origin/main` computed a *points-weighted* grade (`total_earned / total_possible`). For a category with a 100-pt exam (92/100) and a 10-pt quiz (5/10): old = `97/110 = 88.2%`, new = `mean(0.92, 0.50) = 71%`. Every student's grade in any category with unequal-point assignments shifts the moment this ships, with no migration or note — the updated test just codifies the new output. If this is intended, it needs to be called out as a breaking change.

frontend/src/components/screens/Gradebook/Course.tsx
```diff
@@ -1,154 +1,1631 @@
...
+  const ungradedAssignments = React.useMemo(
+    () => (data?.assignments ?? []).filter((a) => a.points_earned === null),
+    [data],
+  );
 
-  React.useEffect(() => { reload(); }, [reload]);
+  const augmentedAssignments = React.useMemo(() => {
+    if (!predictorOpen || !data) return data?.assignments ?? [];
+    const curvePolicy = predictorCurveEnabled && data.curve_mode === "curved" ? {
+      curve_avg_target: data.curve_avg_target ?? 0.83,
```
AndresL230: 🔴 **Frontend/backend curve divergence.** When `curve_mode='curved'` but the course policy is unset (`curve_avg_target`/`curve_sd_delta` NULL — e.g. after `handleToggleCurveMode`, which only sends `{curve_mode}`) and assignments carry no per-item targets, the backend's 4-way `is not None` guard in `category_grade` fails and it uses the **raw** score. Here the frontend falls back to hardcoded `?? 0.83` / `?? 0.10` and **applies the curve** anyway. The masthead `data.percent` (server, raw) then disagrees with the curved composition bar / chips for the same course.

backend/routes/gradescope.py
```diff
@@ -0,0 +1,521 @@
...
+            continue
+        title = a.get("name") or "(untitled)"
+        due_date = _due_to_date_string(a.get("due_date"))
+        points_earned = a.get("points_earned")
+        points_possible = a.get("points_possible")
+
+        record_write: dict[str, Any] = {
+            "title": title,
+            "due_date": due_date,
+            "points_earned": encrypt_if_present(points_earned),
```
AndresL230: 🔴 **Silent grade data-loss on re-sync.** If Gradescope renders a grade as `—`/blank (regrade pending, unreleased), `_parse_grade` returns `None`, `encrypt_if_present(None)` → `None`, and this UPDATE overwrites a previously-synced real grade (e.g. 85) with NULL — the assignment flips back to ungraded and the computed grade drops. Consider skipping the points columns when the incoming grade is `None` rather than writing NULL over an existing value.

backend/services/gradescope_service.py
```diff
@@ -0,0 +1,649 @@
...
+
+import logging
+import re
+import time
+from typing import Any
+
+import requests
+from bs4 import BeautifulSoup
+
+from gradescopeapi import DEFAULT_GRADESCOPE_BASE_URL
```
AndresL230: 🟠 **A missing dependency takes down the whole app, not just this feature.** `requests`, `bs4`, and `gradescopeapi` are imported unconditionally here (only Playwright is guarded below), and `main.py:24` imports `routes.gradescope` unconditionally. If any of these newly-added deps isn't installed in a deploy target, `import routes.gradescope` raises `ImportError` during router mount and **uvicorn never boots** — all of `/api` goes down. Guard these imports the way Playwright already is, or make the router mount degrade gracefully.

backend/routes/gradescope.py
```diff
@@ -0,0 +1,521 @@
...
+                    record_write,
+                    filters={"id": f"eq.{by_gs_id[gs_id]}"},
+                )
+                updated += 1
+            else:
+                table("assignments").insert({
+                    "id": str(uuid.uuid4()),
+                    "user_id": user_id,
+                    "course_id": sapling_course_id,
+                    "category_id": None,  # user can re-categorize later
```
AndresL230: 🟠 **Synced grades never affect the computed grade.** Inserted assignments get `category_id=None`, but `current_grade()` only buckets assignments whose `category_id` matches a known category — so freshly-synced grades contribute nothing to the percent until the user manually categorizes each one. A successful sync visibly changes the grade by zero, which reads as a broken sync.

backend/services/gradebook_service.py
```diff
@@ -41,38 +42,125 @@ class AssignmentRow(TypedDict, total=False):
...
+        e = a.get("points_earned")
+        aid = a.get("id")
+        if p is None or e is None or aid is None:
+            continue
+        if float(p) <= 0:
+            continue
+        graded.append((float(e) / float(p), float(p), str(aid)))
+    if not graded:
+        return []
+    graded.sort(key=lambda x: (x[0], -x[1], x[2]))
```
AndresL230, line 98: 🟠 **Drop-lowest is computed on RAW scores, but the curved display drops by CURVED score.** `dropped_assignment_ids` ranks by raw `earned/possible`, while the frontend `projectGrade`/curved path drops based on curved values and the 'dropped' badge (`droppedAssignmentIds(..., data.assignments)`) uses raw. Under a curve that reorders scores, the server and UI exclude *different* assignments, and the badge can mark one assignment dropped while the math drops another.

backend/routes/gradebook.py
```diff
@@ -142,10 +148,18 @@ def get_course(course_id: str, request: Request, user_id: str = Query(...)):
         if cid in by_cat:
             by_cat[cid].append(a)
     for c in cats:
-        c["category_grade"] = gradebook_service.category_grade(by_cat[c["id"]])
+        c["category_grade"] = gradebook_service.category_grade(
```
AndresL230, line 277: 🟠 **Per-category grade is always raw, even for curved courses.** This call passes no curve args (defaults to `curve_mode='raw'`), while the overall `percent` just below (`:155`) is curved. The per-category numbers shown to the student won't reconcile with the curved headline grade.

backend/config.py
```diff
@@ -15,6 +15,8 @@
 PORT = int(os.getenv("PORT", "5000"))
 FRONTEND_URL = os.getenv("FRONTEND_URL", "http://localhost:3000")
 SESSION_SECRET = os.getenv("SESSION_SECRET", "")
+_sc_env = os.getenv("SECURE_COOKIES")
+SECURE_COOKIES: bool = _sc_env.lower() == "true" if _sc_env is not None else FRONTEND_URL.startswith("https://")
```
AndresL230, line 19: 🟠 **OAuth cookie `Secure` flag is no longer fail-closed.** `origin/main` hardcoded `secure=True`; now it's `secure=SECURE_COOKIES`, which defaults to `FRONTEND_URL.startswith('https://')` when the env var is unset — and `FRONTEND_URL` itself defaults to `http://localhost:3000`. A TLS-fronted prod deploy that forgets to set `FRONTEND_URL`/`SECURE_COOKIES` will send the `sapling_oauth` cookie without `Secure` (used in auth.py:280/308/418/447). The frontend's `secure: NODE_ENV==='production'` for the session cookie (session/route.ts:101) is a second, independent mechanism that can disagree.

backend/routes/gradescope.py
```diff
@@ -0,0 +1,521 @@
...
+            "due_date": due_date,
+            "points_earned": encrypt_if_present(points_earned),
+            "points_possible": encrypt_if_present(points_possible),
+            "source": "gradescope",
+            "gradescope_assignment_id": gs_id,
+        }
+        try:
+            if gs_id in by_gs_id:
+                table("assignments").update(
+                    record_write,
```
AndresL230, line 514: 🟡 **Re-sync clobbers user edits.** `record_write` unconditionally rewrites `title`/`due_date`/`source` on every sync, discarding any manual rename or due-date fix the student made to a linked assignment. Sync should be authoritative only for the grade columns + the idempotency key (`gradescope_assignment_id`).

frontend/src/components/screens/Gradebook/Course.tsx
```diff
@@ -1,154 +1,1631 @@
...
+    return data.assignments.map((a) => {
+      const hyp = hypotheticals.get(a.id);
+      let result = a;
+      if (a.points_earned === null && hyp) {
+        result = {
+          ...a,
+          points_earned: hyp.earned,
+          points_possible: hyp.possible > 0 ? hyp.possible : a.points_possible,
+          // Predictor override takes priority over stored assignment class stats
+          curve_class_mean: hyp.curveClassMean !== null ? hyp.curveClassMean : a.curve_class_mean,
```
AndresL230, line 369: 🟡 **Predictor can't clear a per-assignment curve override.** `hyp.curveClassMean !== null ? ... : a.curve_class_mean` — when the user blanks the class-avg field, `GradePredictorPanel` emits `null`, so this falls back to the stored mean and the curve stays applied. (The panel itself reads with `!== undefined`, so the two layers disagree.) The 'what-if no curve' scenario is silently ignored.

backend/services/gradebook_service.py
```diff
@@ -102,13 +213,17 @@ def letter_for(percent: Optional[float], scale: Optional[list[dict]]) -> Optiona
...
+    # from arithmetic should still reach the 90.0 threshold) without promoting
+    # values that are semantically below a boundary (92.999 must stay A-, not A).
+    rounded = round(percent, 4)
     if scale:
         ordered = sorted(scale, key=lambda x: -float(x.get("min", 0)))
         for tier in ordered:
-            if percent >= float(tier["min"]):
+            if rounded >= float(tier["min"]):
                 return str(tier["letter"])
         return None
```
AndresL230, line 284: 🟡 **A custom letter scale without a 0-floor tier yields no letter for failing students.** `set_letter_scale` doesn't require an `F`/`min:0` tier, so a scale like `[{90,A},{80,B},{70,C}]` makes this return `None` for 55% — the UI shows `55%` with letter `—`. `tierFor` in `GradeProjector.tsx` has the same gap. Consider falling back to the lowest tier instead of `None`.

backend/db/migration_gradebook_curve.sql
```diff
@@ -0,0 +1,34 @@
+-- Gradebook bell-curve grading: per-course curve policy + per-assignment class stats.
```
AndresL230: 🔵 **Duplicate migration tree.** This file (and `migration_gradebook_drops.sql`, `migration_gradescope.sql`) is byte-identical to `db/migrations/0021_gradebook_curve.sql` (resp. 0019/0020). The runner only globs `migrations/*.sql`, so these loose copies never run and only exist to drift out of sync. Recommend deleting them and keeping the numbered sequence as the single source of truth.

backend/db/migrations/0001_baseline_schema.sql
```diff
@@ -53,6 +53,9 @@ CREATE TABLE IF NOT EXISTS user_courses (
     nickname        TEXT,
```
AndresL230, line 53: 🔵 **Editing an already-applied baseline migration is a no-op on existing DBs.** Migrations are tracked by filename, so these added curve columns never execute on staging/prod — they exist there only because idempotent `0021` also adds them. The edit also adds `curve_*` but not `drop_lowest` or `gradescope_assignment_id`, so anything deriving schema from the baseline alone (test fixtures, archived schema) gets tables missing those columns. Let `0019-0021` own the new columns; don't edit applied baselines.

frontend/src/components/Gradebook/curveUtils.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (classSd <= 0) return scorePct;
+  const z = (scorePct - classMean) / classSd;
+  return Math.max(0, Math.min(1, avgTarget + z * sdDelta));
+}
+
+/**
+ * Return a GradedAssignment with points_earned replaced by the curved value.
+ * Returns the original assignment unchanged if it has no curve data.
+ */
+export function applyCurveToAssignment(
```
AndresL230, line 28: 🔵 **Grade math is duplicated across four implementations** (this file, `GradeProjector.tsx`, `Course.tsx`, and the Python `gradebook_service.py`), plus `DEFAULT_SCALE`/`tierFor` and the `?? 0.83`/`?? 0.10` policy are copy-pasted in ~5 spots. This duplication is the root cause of the divergences in the curve/drop findings. Only the predictor genuinely needs client-side compute; the plain 'curved' display could consume the server's `percent`/`category_grade` instead of re-deriving them.

AndresL230:
## Review summary — Gradebook Feature

Reviewed at high recall (correctness + cleanup/altitude). **14 inline comments** posted on the diff; this comment covers the overview, one finding that doesn't map to a changed line, and what I checked that came back clean.

### Top asks before merge
1. **Confirm the `category_grade` change is intentional** (inline at `gradebook_service.py:149`) — it silently rewrites every existing grade in any category with unequal-point assignments (points-weighted → unweighted mean).
2. **Gate the Gradescope feature** so the critical issues can't bite: missing-dependency boot crash (`gradescope_service.py:27`), grade-wiped-to-NULL on re-sync (`gradescope.py:481`), and synced grades not counting because `category_id` is NULL (`gradescope.py:498`).
3. **Reconcile the curve frontend/backend math** — unset-policy divergence (`Course.tsx:356`), raw-vs-curved drop-lowest (`gradebook_service.py:102`), raw per-category grade (`gradebook.py:151`).

### Additional finding (no single changed line to attach to)
🔴 **The course-list endpoint ignores `curve_mode`.** `get_summary` (`backend/routes/gradebook.py:96`) never selects `curve_mode`/`curve_avg_target`/`curve_sd_delta` and calls `current_grade(...)` with no curve args (defaults to `'raw'`), whereas `get_course` (`:155`) computes curved. So the Landing/overview card shows e.g. 74%/C while opening the course shows 88%/B+ — the headline grade flips on click for any curved course.

### Lower-priority notes
- `sync_course` does one DB write per assignment (N+1) — batch into `insert([...])` / `upsert(..., on_conflict=...)`.
- `_load_creds` returns the raw decrypt exception in the 500 `detail` (`gradescope.py:130`) — minor info leak; return a generic message.
- Rate limiting is per-process in-memory, so `bu-sso`/`sync` caps aren't enforced across multiple replicas.

### Verified clean (so you don't re-check these)
- **eslint-suppressions / CI:** dropping the baseline + `--max-warnings` does **not** break the Frontend check — `eslint .` exits 0 on this branch (37 warnings, 0 errors).
- **auth.py googleapiclient → httpx switch:** *adds* error handling the old `.execute()` lacked; no token-refresh behavior lost. Not a CLAUDE.md violation (the `httpx` rule is scoped to Supabase; this is a Google call).
- **CLAUDE.md conventions:** Supabase access all goes through `table()`; gradescope router mounted at `/api/gradescope`; credentials/points/notes encrypted at write boundaries; tests in `backend/tests/`. No violations found.
- **`current_grade` weight renormalization** over graded categories is pre-existing (identical in `origin/main`), not introduced here.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

AndresL230:
Superseded by the DB modular redesign (#279): the gradebook is rebuilt enrollment-keyed (`gradebook_categories`/`assignments` on `enrollment_id`, curve + drop-lowest, per-semester + cumulative GPA) in `0021` + the gradebook slice. Closing as obsolete — this would collide with the redesigned schema. Reopen if any behavior here isn't covered by #279.

AndresL230:
Reopening. This was closed as superseded by the DB modular redesign (#279), but that was over-broad: the redesign only re-keyed the gradebook **schema**, it did not rebuild the **Gradescope import integration** (BU SSO sync, `gradescope_service.py`) in this PR. That feature work is net-new and shouldn't have been swept up. Needs a rebase onto the new enrollment-keyed gradebook schema, then it can resume review.

Issue #12, opened 2026-03-18
Schema Consolidation: Merge 1:1 Tables & Deduplicate Storage


AndresL230:
## Summary

Slim the database from 18 tables to ~14 by merging 1:1 relationships and removing duplicate data storage.

### Merge into `rooms`
- Fold `room_summaries` (`summary`, `member_hash`, `summary_updated_at`) directly into the `rooms` table. It's a 1:1 relationship — no reason for a separate table + join.

### Merge into `users`
- Fold `oauth_tokens` (`access_token`, `refresh_token`, `token_expires_at`) into `users`. Already storing `google_id` and `auth_provider` there.

### Merge into `graph_nodes`
- Fold `quiz_context` (`quiz_context_json`) into `graph_nodes`. It's just a JSONB blob keyed by `(user_id, concept_node_id)` — eliminate the extra table.

### Deduplicate flashcard storage
- Remove `flashcards` JSONB column from `documents` table. The standalone `flashcards` table already tracks spaced repetition state (`times_reviewed`, `last_rating`, `last_reviewed_at`), so it should be the single source of truth.

### Fix column types
- `assignments.due_date`: `text` → `timestamptz`
- `users.last_active_date`: `text` → `date`

## Impact
- **18 → ~14 tables**
- Fewer joins, simpler queries
- Eliminates data duplication for flashcards
- Enables proper date comparisons without casting

## Notes
- `graph_nodes.mastery_events` (JSONB array) is fine for now but may need its own table if we ever query/aggregate mastery history at scale.
- Consider consolidating `documents.summary` + `documents.key_takeaways` into a single `ai_output` JSONB column if neither is queried independently.

PR #253, opened 2026-06-22
fix(social): get_students 400s on nonexistent courses.user_id (#158)

### LINKED ISSUE #158: social.get_students selects courses.user_id (nonexistent column) → students/class-intel endpoint 400s
**Severity:** High
**Labels:** bug, backend, P1
**Location:** `backend/routes/social.py:426-431` (`get_students`); contrast schema `backend/db/supabase_schema.sql:31-45`

### Description
After the course_id migration (#12), `courses` is a shared catalog with **no `user_id`** column (columns: `id, course_code, course_name, ...`; enrollment lives in `user_courses`). `get_students` runs `table("courses").select("user_id,course_name")` and then indexes `courses_by_user[c["user_id"]]`. PostgREST rejects selecting a nonexistent column with HTTP 400 (`column courses.user_id does not exist`), so the whole students/class-intel endpoint errors. Even if it didn't 400, `courses_by_user` would be empty because enrollment is in `user_courses`. This endpoint is live (called by `frontend/src/lib/api.ts:380 getStudents`).

### Steps to reproduce
1. Open the Social / class-intel surface that calls `getStudents`.
2. Backend logs / network show a 400 from `/api/social/...students`.

### Expected vs actual
- **Expected:** returns the list of students enrolled in shared courses.
- **Actual:** 400 — the feature is broken.

### Suggested fix
Join through `user_courses` instead of selecting a column that no longer exists:
```python
rows = table("user_courses").select("user_id, courses(course_name)")
```

### Acceptance criteria
- `getStudents` returns 200 with correct per-course student grouping.
- A test exercises the endpoint against the post-#12 schema.

Jose-Gael-Cruz-Lopez:
After the #12 course_id migration, `courses` is a shared catalog with **no `user_id`** (enrollment lives in `user_courses`). `get_students` ran `table("courses").select("user_id,course_name")` and indexed `courses_by_user[c["user_id"]]`. PostgREST rejects selecting a nonexistent column with **HTTP 400** (`column courses.user_id does not exist`), so the entire students / class-intel endpoint errored — and it's live (`frontend/src/lib/api.ts` `getStudents`).

### Fix — `backend/routes/social.py`
Join through `user_courses` and embed the course name:
```python
enrollment_rows = table("user_courses").select("user_id,courses(course_name)")
for row in enrollment_rows:
    course = row.get("courses")
    name = course.get("course_name") if isinstance(course, dict) else None
    if name:
        courses_by_user[row["user_id"]].append(name)
```

### Tests — `backend/tests/test_social_get_students.py`
- 200 with correct per-course grouping + mastery stats against the post-#12 shape.
- Students sorted by name.
- Regression guard: the endpoint reads `user_courses` and never selects `user_id,course_name` off `courses`.

### Verification
- `ruff check .` clean; gated suite green (+3 new). The new tests fail against the pre-fix code (the catalog path yields no enrollments), confirming real coverage.

Closes #158.


AndresL230:
Superseded by the DB modular redesign (#279). `social.get_students` (#158) is rewired in the analytics slice to read `enrollments → course_offerings → courses`. Closing as obsolete. Reopen if not fully covered.

AndresL230:
Reopening. Closed as superseded by #279, but this is an app-logic bugfix (`social.get_students` 400 on nonexistent course, #158), not a DB-structure change. Keeping it open to confirm the fix is actually covered by the redesign's analytics slice before closing again — if the rewired `get_students` already handles the missing-course case, we can re-close with a pointer to that commit; otherwise this fix still stands.

AndresL230:
Closing this as **superseded** — the bug it targets is already fixed on `main`, and issue #158 is closed.

To be clear up front: **the diagnosis in this PR was correct when it was written.** `courses.user_id` genuinely didn't exist and the endpoint genuinely 400'd. What happened is that the schema moved underneath the branch while it was open.

### What superseded it

PR #270 (`db/analytics-code: social.py offering re-keying`, epic slice PR4 of the #279 DB modular redesign — merged 2026-06-24, commit `736abc1`) rewrote `get_students` on `main` to walk the enrollment chain:

```python
enrollment_rows = table("enrollments").select("user_id,course_offerings(courses(course_name))")
name_map = get_display_names([u["id"] for u in users])
```

That's the same fix this PR set out to make, arrived at independently, plus the `services/profiles.py` name resolution.

### Why it can't be merged forward

This branch carries migrations up to `0018`; `main` is at `0030`. The fix here targets the 0018-era schema, so rebasing wouldn't rescue it — it would swap one nonexistent reference for another:

- **`user_courses` no longer exists.** `0020_academics_split.sql:84` renamed it to `enrollments`. The new query would fail with a PostgREST "relation does not exist" (404) instead of the original 400 — still broken, different error.
- **The `courses(course_name)` embed can't resolve.** `0020:85` renamed `course_id` → `offering_id`, FK'd to `course_offerings(id)` (`0020:73-74`). There's no direct `enrollments → courses` relationship; it has to go through `course_offerings`.
- **`users.name` is gone.** `0024_identity_split.sql:25` dropped it; names moved to the encrypted `user_profiles.name` and resolve via `get_display_names()`. Reading it directly returns nothing.

GitHub already reports this as `CONFLICTING` on `backend/routes/social.py`, which is the two fixes colliding rather than ordinary drift.

### The tests would need rewriting, not rebasing

`tests/test_social_get_students.py` pins the old schema — `assert "user_courses" in recorder` would **fail** against `main`'s correct query. They also `patch("routes.social.table")` wholesale, so they'd pass against any table name and never reproduce the real PostgREST error.

### Worth salvaging separately

One idea here is better than what's on `main` today: neither this PR nor `main` routes through `services/academics.py`, which the conventions call for — both hand-roll the embed. `academics.py` exposes `user_offering_ids_for_course`/`offering_course_id` but has no bulk "names for all enrollments" helper. Worth a follow-up ticket.

Also surfaced while reviewing, unrelated to this PR and present on `main`: `get_students` returns every user in the DB with their courses, mastery stats, and top concepts. It's authenticated but not scoped — the session `user_id` it fetches at `social.py:426` is dead code. That looks intentional for a discovery surface, but it's a global cross-course disclosure worth an explicit decision.

Thanks @Jose-Gael-Cruz-Lopez — the analysis was right, the schema just moved. Closing.

Issue #175, opened 2026-06-12
Tracked vim swap files committed under backend/ + .gitignore gap


Jose-Gael-Cruz-Lopez:
**Severity:** Medium
**Labels:** tech-debt, infrastructure, P2
**Location:** `backend/.main.py.swn`, `backend/.main.py.swo`, `backend/.main.py.swp`; `.gitignore`

### Description
Three vim swap files for `backend/main.py` are tracked in git (`git ls-files` confirms all three). They are binary editor scratch state, leak the committer's editing session, and make vim warn "swap file already exists" for the next person who opens `main.py`. `.gitignore` has no `*.sw[a-p]`/`*~` pattern, so this keeps recurring.

### Suggested fix
```
git rm --cached backend/.main.py.sw{n,o,p}
```
Add to `.gitignore`: `*.sw[a-p]`, `*~`, `.*.swp`.

### Acceptance criteria
- No swap/temp files tracked (`git ls-files | grep -E '\.sw[a-p]$'` empty); `.gitignore` ignores them.

AndresL230:
Resolved on main: no tracked .swp/.swo files remain and the .gitignore gap was closed in c4a3627 (PR #254). Verified via git ls-files. Closing.

Issue #173, opened 2026-06-12
docker-compose.yml builds a frontend/Dockerfile that does not exist — docker compose up fails


Jose-Gael-Cruz-Lopez:
**Severity:** Medium
**Labels:** infrastructure, P2
**Location:** `docker-compose.yml:12-16`; missing `frontend/Dockerfile`

### Description
The `frontend` service declares `build: { context: ./frontend, args: { BACKEND_URL } }`, but there is no `frontend/Dockerfile` (only `backend/Dockerfile` exists). `docker compose build`/`up` aborts with "failed to read dockerfile". The frontend-audit doc (`docs/frontend-audit/00-overview.md:88`) even claims a `frontend/Dockerfile` exists, so the documented local-orchestration path is broken. The frontend service also has no `env_file`, so `NEXT_PUBLIC_*` vars aren't supplied.

### Steps to reproduce
1. From repo root, `docker compose up`.
2. Build fails on the missing `frontend/Dockerfile`.

### Expected vs actual
- **Expected:** `docker compose up` builds and runs both services.
- **Actual:** aborts immediately.

### Suggested fix
Add a multi-stage `frontend/Dockerfile` (`next build` → runtime) and an `env_file: ./frontend/.env`, or remove the `frontend` service from compose if Cloudflare Workers is the only intended deploy.

### Acceptance criteria
- `docker compose up` builds and starts frontend + backend, or the compose file no longer references a nonexistent Dockerfile.

AndresL230:
Resolved on main: frontend/Dockerfile now exists, added in c4a3627 (PR #254), so the docker-compose frontend build target is real. Closing.

Issue #97, opened 2026-05-09
[P3] Add Redis caching layer for Gemini LLM + OCR extraction


AndresL230:
## Problem

The two heaviest operations in the request path have no caching:

- **Gemini calls** in `backend/services/gemini_service.py` (`call_gemini`, `call_gemini_json`, `call_gemini_multiturn`) — every classify/summarize/extract on document upload, every study-guide generation, every room summary hits the API fresh. Latency is multi-second and cost scales linearly with traffic.
- **OCR/extraction** in `backend/services/extraction_service.py` (Docling / GOT-OCR / Tesseract). Re-uploads of the same file (common during testing, retries, and student re-submissions) re-run the full OCR pipeline.

There is currently no shared cache across workers — only the per-request `stat_cache` dict in `backend/routes/profile.py:491` and DB-backed table caches (`room_summaries`, `course_summary`, `study_guides`).

## Proposal

Introduce Redis as a shared cache, accessed through a thin wrapper module (e.g., `backend/services/cache.py`) so it can be mocked in tests the way `db/connection.py` is.

**Cache keys:**
- Gemini: `gemini:{sha256(model + prompt + params + response_schema)}`
- OCR: `ocr:{sha256(file_bytes)}:{backend_name}`

**TTLs:**
- Gemini classify/extract on document content: long (7–30 days) — content-addressed, safe to keep.
- Gemini summaries that depend on graph/mastery state: short or hash-keyed on the underlying state (mirror what `social_cache_service.py` already does with `member_hash`).
- OCR: long (content-addressed by file hash).

## Acceptance criteria

- [ ] `services/cache.py` wrapper with `get`, `set(ttl)`, and `get_or_compute` helpers
- [ ] `gemini_service.call_gemini*` route through the cache (opt-in via kwarg so existing callers can skip when needed)
- [ ] `extraction_service` caches OCR results by file-content hash + backend name
- [ ] `REDIS_URL` added to `.env.example` and `config.py`
- [ ] `docker-compose.yml` adds a `redis` service
- [ ] Tests use a fake/in-memory Redis (e.g., `fakeredis`) via `conftest.py`
- [ ] Cache hit/miss metrics logged so we can measure the win

## Tradeoffs / risks

- New infra dependency (Redis container + env config).
[...]

PR #254, opened 2026-06-22
chore(infra): pin deps + lockfile, untrack swap files, add frontend Dockerfile (#163, #175, #173)

### LINKED ISSUE #163: Backend dependencies are unpinned with no lockfile — non-reproducible builds, supply-chain exposure
**Severity:** High
**Labels:** dependencies, backend, infrastructure, P1
**Location:** `backend/requirements.txt`

### Description
~22 of ~30 lines have no version constraint — including behavior/security-critical `fastapi`, `uvicorn[standard]`, `httpx`, `google-genai`, `google-auth-oauthlib`, `google-api-python-client`, `pillow`, `pypdf`, `python-multipart`, `beautifulsoup4`. There is no `requirements.lock` / `constraints.txt` / `uv.lock` (confirmed via `git ls-files`). Every fresh `pip install` (CI, `backend/Dockerfile`, evals) resolves to whatever is newest on PyPI that day, so two builds days apart can ship different FastAPI/Pillow/transformers majors. Builds are non-reproducible, a specific dependency can't be audited/rolled back, and there's no floor against a yanked or compromised release.

### Steps to reproduce
1. `pip install -r backend/requirements.txt` today and again after an upstream release; compare `pip freeze`.

### Expected vs actual
- **Expected:** pinned, lockable, reproducible dependency set.
- **Actual:** floating versions; build output depends on the calendar date.

### Suggested fix
Pin runtime deps (`==`/`~=`) and generate a fully-resolved lockfile (`uv pip compile`/`pip-compile` → `requirements.lock`) that CI and Docker install from (`--require-hashes` where feasible).

### Acceptance criteria
- A committed lockfile exists and CI/Docker install from it.
- `requirements.txt` has explicit constraints on all runtime deps.

Jose-Gael-Cruz-Lopez:
Three infra-hygiene fixes from the backend audit.

### #163 — reproducible, pinned, lockable dependencies
- `backend/requirements.txt` — every runtime dep now has an explicit constraint (the ~22 previously-floating lines, incl. `fastapi`, `uvicorn`, `httpx`, `google-*`, `pillow`, `pypdf`, `beautifulsoup4`, are pinned to validated versions).
- `backend/requirements.lock` — fully-resolved, **hash-pinned, universal** lockfile (`uv pip compile --generate-hashes --universal --python-version 3.13`) for the fast (non-OCR) set CI installs. The OCR stack (torch/docling/transformers) stays out of the lock by design — multi-GB, installed from the CPU wheel index in Docker, excluded from the fast CI gate.
- `.github/workflows/ci.yml` — installs via `pip install --require-hashes -r requirements.lock` (was a date-dependent `pip install` of floating versions); cache key switched to the lock.
- `backend/tests/test_requirements_lock.py` — guards that nothing in `requirements.txt` is unpinned and the lock pins every non-OCR dependency.

### #175 — untrack committed vim swap files
- `git rm --cached backend/.main.py.sw{n,o,p}` (binary editor scratch state that leaked editing sessions and triggered "swap file already exists" warnings).
- `.gitignore` now ignores `*.sw[a-p]`, `.*.sw[a-p]`, `*~` so it can't recur.

### #173 — `docker compose up` no longer fails on a missing Dockerfile
- `frontend/Dockerfile` — new multi-stage build for the Next.js 16 app (uses `output: "standalone"`; node 22 + npm pinned `>=10.9 <11`; non-root runtime).
- `frontend/.dockerignore` — keeps the build context small and stops local `node_modules`/`.next` from clobbering the image.
- `docker-compose.yml` — adds the missing `env_file: ./frontend/.env` so `NEXT_PUBLIC_*` vars reach the build/runtime.

### Verification
- `ruff check .` clean; gated suite green (+3 new); `pip install --require-hashes -r requirements.lock` dry-run resolves; no swap files tracked (`git ls-files | grep '\.sw[a-p]$'` empty). No docker CLI locally, so the image build itself is left for a reviewer/CI with Docker.

Closes #163, #175, #173.

## Summary by CodeRabbit

* **New Features**
  * Added a production-ready Docker setup for the frontend, including a smaller runtime image and non-root execution.
  * Made the frontend service easier to configure locally with optional environment file support and a default backend URL.

* **Bug Fixes**
  * Tightened backend dependency handling for more reliable installs in CI.
  * Added checks to ensure dependencies are consistently pinned and lockfile-based.

* **Chores**
  * Expanded ignore rules for common editor scratch files and local build artifacts.


AndresL230:
@Jose-Gael-Cruz-Lopez pushed 5 commits (fast-forward, your history intact) and **converted this to a draft** while it gets sorted — it had a green check from 2026-06-24 sitting on a 419-commit-stale base, which is exactly the thing that gets merged by accident.

### The blocker: this PR would have turned `main` red

`main` gained `redis>=5.0` via `f12c99b` (#97) **after** this branch generated its lock. So on merge, **this PR's own guard test** — `test_lock_pins_every_non_ocr_requirement` — fails with `missing == ['redis']`. I confirmed by rebasing and running pytest before touching anything. Nice property, incidentally: the guard caught its own staleness.

Fixed: merged `main`, regenerated the lock with `uv` using the header's exact invocation (including the literal `/tmp/req.in`, since the committed `# via -r /tmp/req.in` annotations depend on that path). **Lock diff is exactly +4 lines** — `redis==8.0.1` + 2 hashes + via. Zero collateral version churn, which is the evidence the regeneration is faithful. Also pinned `redis>=5.0,<9` in `requirements.txt` — a floating constraint in the PR whose thesis is eliminating floating constraints.

Context you may want: **Redis is optional and off.** No `REDIS_URL` is set anywhere, `cache.py` no-ops without it, and the library is installed-but-never-imported. This is pure manifest bookkeeping — nothing is being deployed or enabled.

### The frontend Dockerfile never built. At all.

This is the one I did not expect. `docker build -f frontend/Dockerfile frontend/` fails at `npm run build` with `spawn .../workerd ENOENT`.

Root cause: `next.config.ts` calls `initOpenNextCloudflareForDev()` at module load, which spawns `@cloudflare/workerd-linux-64` — a **glibc** binary — on **musl** `node:22-alpine`. The 124MB file is right there; its ELF interpreter `/lib/ld-linux-x86-64.so.2` is not (alpine ships only `ld-musl-x86_64.so.1`). Cloudflare publishes no musl workerd.

Nothing caught this because **no workflow builds this image**, and CI runs `npm run build` on `ubuntu-latest` (glibc) — so `-slim` is what actually matches CI, contrary to the Dockerfile's own comment. Switched all 3 stages to `node:22-slim` and ported the runner's busybox `addgroup/adduser` to Debian `groupadd/useradd`. Image is larger than alpine; correctness over size, since alpine cannot work here at all.

Two side notes from this: `initOpenNextCloudflareForDev()`'s "Safe no-op in prod builds" comment is **false** — it spawns workerd during `next build`. And #173 was unverifiable as written; it's now build-verified.

### The client-bundle bug — and one thing my review missed

`ENV NEXT_PUBLIC_API_URL=${BACKEND_URL}` inlines the compose-internal hostname into the **client** bundle at build. The browser runs on the host, so `backend` is unresolvable → sign-in, onboarding, and newsletter all break under `docker compose up`, the exact scenario #173 exists to fix.

**What I missed:** `middleware.ts:16` also reads `NEXT_PUBLIC_API_URL`, and middleware runs **server-side**. Splitting the vars as I originally specified would have traded a broken browser for a broken auth check — `localhost:5000` is unreachable from inside the frontend container, so every protected route would bounce to sign-in. Middleware now prefers `BACKEND_URL` with a `NEXT_PUBLIC_API_URL` fallback; no-op in prod/staging where `wrangler.toml` sets both identically.

Proven against the real built image rather than asserted:
- `.next/static` → **0** hits for `backend:5000`, contains `localhost:5000`
- `routes-manifest.json` → `http://backend:5000/api/:path*`

That second one also empirically settles the compose comment: rewrites are serialized at **build** time, not evaluated at server runtime. Comment corrected.

### Judgement calls I did NOT make for you

- **`backend/Dockerfile` still installs from `requirements.txt`, not the lock.** The 3.13-vs-3.12 concern turned out to be a red herring (verified: `pip install --require-hashes --dry-run` on 3.12 → exit 0, 113 packages; the lock carries no `python_full_version` markers). The real blocker: the lock covers only the non-OCR subset while the image needs torch/docling/transformers. `--require-hashes` is all-or-nothing, and chaining `-r lock` then `-r requirements.txt` lets the OCR resolve silently upgrade locked pins — making the reproducibility claim *false* rather than true. Needs a second OCR-inclusive lock against the torch CPU index. Documented in-file; left for you.
- **Didn't add upper bounds to the 4 floating deps** (`pydantic-ai-slim`, `logfire`, `sse-starlette`, `pydantic-evals`). All are pre-1.0/fast-moving, where a `<1` ceiling is arbitrary (pre-1.0 breaks on *minor* bumps, so it gives false comfort) and silently blocks security updates. Inventing four ceilings is a policy call the team hasn't made; the lock already pins all four exactly. Instead **renamed** the weak test to `test_no_requirement_is_completely_unconstrained` with the gap documented — it only checked `[<>=!~]` presence, so `>=0.0.20` passed as "constrained."

### Still recommend splitting this

#175 (swap files) is **verified defect-free and merge-ready today** — I confirmed its `.gitignore` block survives the merge byte-identical — but it's bundled with two concerns that carried real bugs. The revertability argument is the sharp one: reverting the Dockerfile bug would also revert the dependency pinning and re-float the entire backend. Your call, but #175 could land in the next five minutes on its own.

### Verification

`ruff` clean · `test_requirements_lock.py` **3 passed** (was failing) · backend **933 passed** · frontend **77 passed** incl. `middleware.test.ts` · `docker build` **succeeds** · container `GET / → 200` as non-root.

Two caveats: the `1 failed + 1 error` in `test_ocr_pipeline.py` are **pre-existing** — I reproduced the identical result on a pristine `origin/main` worktree. And I did not run full `docker compose up` (needs a populated `backend/.env` + real credentials); the frontend image is build/run/bundle-verified, the compose wiring verified by inspecting the baked artifacts.

One contradiction worth a follow-up: `frontend/.env.example:8` says leave `NEXT_PUBLIC_API_URL` empty for same-origin, but `wrangler.toml` sets it to the API subdomain and both `SignInModal` and `middleware` require it non-empty. The docs and the code disagree.

AndresL230 (approved):
Approving. The three blockers are fixed and verified above: the lock now pins `redis==8.0.1` (its own guard test was going to turn `main` red on merge), the client bundle no longer carries the compose-internal hostname (**0** hits for `backend:5000` in the built `.next/static`), and the frontend image — which **never built at all** on alpine/musl — now builds and serves `200` as non-root.

`test_requirements_lock.py` 3 passed (was failing), backend 933 passed, frontend 77 passed, `docker build` succeeds. The `1 failed + 1 error` in `test_ocr_pipeline.py` are pre-existing and reproduce identically on a pristine `origin/main` worktree.

Merging, but the split recommendation stands for next time — bundling #163/#175/#173 meant a verified-clean concern (#175) sat blocked for three weeks behind two that carried real bugs, and the 92% lockfile noise is a large part of why the Dockerfile bug survived review. Two follow-ups left in the PR for you: the `backend/Dockerfile` lock install (needs a second OCR-inclusive lock — `--require-hashes` is all-or-nothing) and the upper bounds on the 4 pre-1.0 deps, which is a policy call rather than a fix.

Issue #190, opened 2026-06-12
COOKIE_DOMAIN is set unvalidated and CSRF defense relies solely on SameSite=Lax


Jose-Gael-Cruz-Lopez:
**Severity:** Low
**Labels:** security, frontend, P3
**Location:** `frontend/src/app/api/auth/session/route.ts:73-96`

### Description
Cookie flags are correct now (`httpOnly`, `secure`, `sameSite:'lax'`, fixed `maxAge`). Two residual hardening gaps: (1) `COOKIE_DOMAIN` is set verbatim from env with no validation — a misconfigured/overly-broad domain would scope the session cookie wider than intended across subdomains; (2) state-changing requests rely solely on `SameSite=Lax` for CSRF — there's no anti-CSRF token, so any future `SameSite=None` change (or a top-level GET side effect) would be unprotected.

### Suggested fix
Validate/whitelist `COOKIE_DOMAIN` at startup; adopt a `__Host-` cookie prefix (forbids `domain`, forces `secure`+path=/) or a double-submit CSRF token for the session endpoint.

### Acceptance criteria
- `COOKIE_DOMAIN` is validated; a documented CSRF strategy beyond SameSite exists for state-changing requests.

PR #255, opened 2026-06-22
fix(auth): verify + document session-token lifecycle; lock the cross-service contract (#168)

### LINKED ISSUE #168: Backend session token has a hard 5-minute lifetime with no refresh path (verify)
**Severity:** High *(please verify — see note)*
**Labels:** bug, backend, P1, question
**Location:** `backend/routes/auth.py:386` (only token mint, `exp = now + 300`); `backend/services/auth_guard.py:48-49` (hard-rejects expired); cookie `sapling_session` accepted at `auth_guard.py:19` but **never** `set_cookie`'d anywhere (grep).

### Description
The only place the backend mints a session token is the OAuth callback, with a 300-second expiry, and there is no `/refresh` route. The `sapling_session` cookie that `_decode_session` also accepts is never set by the backend. If that reading is correct, then after 5 minutes every `require_self`/`require_admin`/`get_session_user_id` request returns 401 "Session expired" and the only recovery is a full Google re-login.

**Note / verification:** this is a strong claim; before fixing, confirm whether a shorter-lived token is exchanged/refreshed by the frontend session BFF (`frontend/src/app/api/auth/session/route.ts`) or whether sessions genuinely die at 5 minutes. Filing as a flagged finding per audit policy.

### Steps to reproduce (to confirm)
1. Sign in; note the time.
2. Idle ~5+ minutes, then make any authenticated request.
3. Observe whether it 401s with "Session expired" and forces re-login.

### Expected vs actual
- **Expected:** a usable session duration with a refresh path.
- **Actual (if confirmed):** sessions expire after 5 minutes with no refresh.

### Suggested fix
Issue a longer-lived token and/or add a `/api/auth/refresh` endpoint that re-mints from a valid token or the stored OAuth refresh token; set the `sapling_session` cookie the decoder already supports.

### Acceptance criteria
- Confirmed behavior documented; if real, sessions persist for a configured TTL and refresh without a full re-login.

Jose-Gael-Cruz-Lopez:
#168 was filed as a **flagged "verify"** finding: does the backend session really die after 5 minutes with no refresh? I traced the full path across both services. **The worst case is not real** — the 300s token is only a one-shot redirect handoff.

### What actually happens
1. Backend OAuth callback mints a short-lived HMAC token (`exp = now + 300`) and redirects to the frontend with `?auth_token=…`.
2. The **frontend session BFF** (`frontend/src/app/api/auth/session/route.ts`) verifies it and re-mints a **30-day** token (`SESSION_MAX_AGE = 2592000`) in a **byte-identical** format, set as the `httpOnly`/`Secure` **`sapling_session` cookie** (scoped via `COOKIE_DOMAIN`).
3. `auth_guard._decode_session` reads that cookie, verifies the HMAC with the shared `SESSION_SECRET`, and accepts it for the full 30 days.

So sessions persist **30 days**, not 5 minutes. It holds as long as `SESSION_SECRET` matches across services and `COOKIE_DOMAIN` reaches the backend subdomain (both documented).

### Changes
- `backend/routes/auth.py` — named the magic `300` as `_REDIRECT_TOKEN_TTL_SECONDS` (env-overridable via `SAPLING_AUTH_REDIRECT_TOKEN_TTL`) and corrected the comment to say it's the redirect-handoff TTL, **not** the session TTL.
- `docs/decisions/0018-session-token-lifecycle.md` — documents the full lifecycle, the verification outcome, the operational preconditions, and follow-ups (sliding refresh is a frontend-BFF concern, out of scope).
- `backend/tests/test_auth_session_contract.py` — locks the cross-service contract: a frontend-style 30-day token is **accepted** by the backend decoder; expired, tampered, and wrong-secret tokens are rejected; the redirect TTL stays short.

### Verification
- `ruff check .` clean; gated suite green (+6 new). The acceptance criterion ("confirmed behavior documented; sessions persist for a configured TTL") is met.

Closes #168.

## Summary by CodeRabbit

- **New Features**
  - Added configurable expiration for one-time authentication handoff tokens, with safe limits and warnings for invalid settings.
  - Preserved support for long-lived sessions and legacy authentication flows.

- **Bug Fixes**
  - Improved handling of expired, tampered, or incorrectly signed session tokens.

- **Documentation**
  - Documented session-token lifecycle, authentication flow, configuration requirements, and operational considerations.

- **Tests**
  - Added coverage for session compatibility, token validation, expiration, and configuration behavior.


AndresL230:
@Jose-Gael-Cruz-Lopez I pushed 4 commits to this branch as part of a review sweep. Fast-forward only — nothing of yours was rewritten. Happy to back any of it out.

**First, the headline: your central claim is correct and I verified it independently.** Sessions last 30 days, not 5 minutes. The 300s token is a one-shot redirect handoff; the BFF (`route.ts:99-106`) re-mints a 30-day token into an httpOnly/Secure/SameSite=Lax cookie that `auth_guard._decode_session` accepts. Closing #168 as "not a bug" is right. **The code and tests are sound — I did not rewrite them.** I also confirmed the tests are real rather than testing conftest's stub, by mutation: disabling the expiry check at `auth_guard.py:47` fails `test_expired_token_is_rejected`; disabling signature verification at `:34` fails both tamper tests. `conftest.py:88` captures `_real_decode_session` *before* the monkeypatch, so it's the genuine decoder.

The changes are confined to the ADR prose, one test name, and one small guard.

### `09b95bd` — the ADR aside, this was the only real code bug

`auth.py:93-95` crashed the app at **import**:
```python
_REDIRECT_TOKEN_TTL_SECONDS = _clamp_redirect_ttl(int(os.getenv("SAPLING_AUTH_REDIRECT_TOKEN_TTL", "300")))
```
Reproduced: `SAPLING_AUTH_REDIRECT_TOKEN_TTL=abc` and `=` (empty) both raise `ValueError` at module import → the router fails to import → the app won't boot. Empty-string is realistic — declaring the var in Railway/Wrangler without a value does it. Fails closed, so not a security hole, but a nasty deploy-time footgun in the one code path this PR adds. `_clamp_redirect_ttl`'s docstring already promised to defend against "a misconfigured override" but only clamped range, not parseability — now it lives up to it via `_parse_redirect_ttl` → warn + fall back to 300.

### `187f419` + `350f1a7` — the ADR documented the wrong mechanism (the one that mattered)

`0018:47-51` said `COOKIE_DOMAIN` covers both subdomains so the browser sends `sapling_session` cross-origin with `credentials: 'include'`, and that a host-only cookie wouldn't reach the backend. That's not the topology: `api.ts:12` sets `API_URL = ''` and `next.config.ts:35` rewrites `/api/:path*` → `${BACKEND_URL}/api/:path*` **server-side**, so all 135 `fetchJSON` sites are same-origin and the Next rewrite forwards the `Cookie` header. A host-only cookie **would** work fine.

The genuinely load-bearing precondition — **`BACKEND_URL` must be set at *build* time** — wasn't mentioned at all. Concretely: it sits in wrangler's `[vars]` (worker *runtime* vars) while `next.config.ts` reads it at *build* time, so `[vars]` alone doesn't feed the rewrite → falls back to localhost → `/api/*` 500s. That's now the documented precondition, with `COOKIE_DOMAIN` demoted to what it actually governs (the `domain` attribute; still set at `wrangler.toml:19` via `sanitizeCookieDomain` #190, so nothing is broken).

This mattered because the ADR contradicted `frontend/.env.example:5-8`, which documents the opposite. An ADR is the authoritative reference — as written it would teach the next person to point an authed fetch at `NEXT_PUBLIC_API_URL` and reintroduce the 2026-06-30 onboarding-loop bug.

**`350f1a7` is me fixing my own mistake.** My first pass cited that bug as "live right now at `page.tsx:619`, tracked in #339." It is neither. That finding came from an agent reading **this branch, which is 419 commits behind `main`** — where the raw fetch genuinely still exists. On `main` it was fixed weeks ago: the file moved to `(public)/page.tsx` and routes through `submitOnboardingProfile()` → `fetchJSON`, with a comment recording the fix. I filed #339 off that and have since closed it as invalid. The ADR now describes the bug in past tense and points at that comment — a better cautionary example, since it actually happened.

### `1c1adf5` — renamed a test that blessed an unused credential channel

`_decode_session` (`auth_guard.py:19`) reads `request.query_params.get("auth_token")` **before** the cookie. I traced every occurrence: `auth.py:452` puts it in a redirect URL to the *frontend*; `callback/page.tsx:55` reads it and POSTs it to the BFF in a JSON body. **No client ever sends `?auth_token=` to the backend.**

`test_redirect_auth_token_query_param_is_accepted` + its comment ("arrives as `?auth_token` while fresh") asserted a constraint the code doesn't enforce — a 30-day session token in `?auth_token=` is accepted identically, and URL-borne tokens leak via access logs, `Referer`, and history. Renamed to `test_legacy_unused_auth_token_query_param_is_still_accepted` and redocumented as characterization, not endorsement. **Channel left in place** — removing it is a follow-up, and the renamed test makes that removal a visible edit rather than a silent contract break.

Also corrected the "byte-identical format" claim: proved by running both serializers that Python emits `{"user_id": "user_alice", …}` (spaces) and JS `JSON.stringify` emits none → different base64. Harmless (the verifier HMACs the received `payload_b64` opaquely and never re-serializes), but `_mint`'s docstring claiming it signs "exactly like the backend mint AND the frontend `signSession`" was false — it mirrors the **backend** exactly (`auth.py:440` uses default `json.dumps`), not `signSession`.

### One nuance that sharpens your case

"The browser never makes a cross-origin authed API call" is true, but for a sharper reason than I first wrote. `NEXT_PUBLIC_API_URL` *is* consumed in four places, and wrangler sets it to `https://api.saplinglearn.com`. Checking each: `SignInModal`'s uses are top-level OAuth *navigations* (not fetches); `middleware.ts` runs server-side with a hand-set `Cookie` header; and the remaining fetches omit `credentials` entirely. So no cross-origin authed browser call exists — and `COOKIE_DOMAIN` genuinely isn't load-bearing. **Had any of them passed `credentials: 'include'`, `COOKIE_DOMAIN` *would* have been the mechanism and this ADR rewrite would be wrong.**

### Follow-ups (not done here)

1. **Remove the `?auth_token=` channel** — no client uses it; it accepts a 30-day token with no ttl/purpose check.
2. **A shared JSON fixture as the real cross-service lock.** "Lock the cross-service contract" overstates what's here: `_mint` is a *third* Python re-implementation, so the test proves a Python-minted token is accepted, not a real frontend one. There's no `frontend/src/lib/sessionToken.test.ts`, and `route.ts:29-68` hand-rolls `verifyAuthToken` as a duplicate of `sessionToken.ts::verifySession` — so there are **two** frontend copies that can drift. I softened the framing; a checked-in fixture consumed by both suites would close it properly.
3. **Assert the build-time `BACKEND_URL` precondition in CI** — it's the one genuinely load-bearing config, it fails as a confusing 500 rather than a clear error, and nothing guards it.

Verification: `ruff check .` clean, 674 passed, auth contract 18 passed, and the import guard confirmed (`TTL='abc'`/`''` → warn + 300; `'86400'` → 600; `'60'` → 60).

AndresL230 (approved):
Approving. To restate what I said in the comment: **the code and tests here were already sound, and the central finding is correct** — sessions last 30 days, not 5 minutes, and #168 is properly closed as "not a bug." I verified the tests are real by mutation rather than reading (breaking the expiry check and the signature check each fail the right test). The changes above are the ADR mechanism fix, one import-time crash guard, and a test rename — nothing that touches the substance of your work. CI green on `350f1a7`.

Issue #110, opened 2026-06-08
[P1] Fix sub-44px touch targets + SSR desktop FOUC


AndresL230:
**Severity: P1** · Dimension: Responsive / Accessibility · From `docs/frontend-ui-audit.md`

Responsiveness is **100% JS-driven** at a single 768px breakpoint — the whole codebase has **0 width-based `@media` and 0 `@container` queries** (only 2 `prefers-reduced-motion` blocks). Two consequences: SSR always paints desktop, and touch targets are widely sub-44px.

### SSR / first-paint FOUC (CLS)
- `lib/useIsMobile.ts:8` — `useState(false)` means server + first client paint always assume desktop; mobile users get the desktop shell then a layout snap on hydration. With zero CSS width queries, CSS can't paper over it.
- **Fix:** gate the shell on a CSS `@media (max-width:768px)` for the initial paint, or render mobile-first and upgrade.

### Touch targets < 44×44 (WCAG 2.5.5 / mis-tap risk)
- `TopNav.tsx:141-143` hamburger (~28-40px), `:472-475` mobile menu links (~37px), `:217-218,233-234` settings/admin icons (32×32)
- `SideNav.tsx:307` nav rows (~31px), `:239-240` collapse button (24×24), `:269` expand (28px)
- **Fix:** enforce ≥44px hit areas on nav rows, toggles, and icon buttons (`Dialog`'s 44×44 close button is the model).

### Acceptance
- [ ] Initial paint matches viewport (no desktop→mobile snap on phones)
- [ ] Nav rows, hamburger, collapse/expand, and icon buttons are ≥44×44 touch targets

PR #311, opened 2026-07-03
fix(a11y): 44px touch targets for mobile nav (#110, part 1)

Jose-Gael-Cruz-Lopez:
Part of #110 (touch-target half). Brings the genuine **mobile touch surfaces** in `TopNav` to the 44×44 minimum, keeping the existing icon sizes centered in the enlarged hit area:

- Hamburger: ~28×26 → 44×44 (bars centered)
- Settings + Admin icon links: 32×32 → 44×44
- Mobile menu rows: ~37px → `minHeight: 44`

**Verified locally:** `npm run typecheck` clean; `eslint` on the file has 0 errors (1 pre-existing `` warning, unrelated).

**Deliberately not here** (flagged for follow-up):
- **SideNav** is the desktop sidebar (mouse; already meets the AA 24px floor). Forcing 44px there needs a visual pass on the compact layout.
- **SSR desktop-FOUC** (the other half of #110) is a separate change — needs an SSR/`useIsMobile` rework (`useState(false)` currently renders desktop on the server, then snaps on hydration). Tracked, not in this PR.

So #110 stays open; this is the low-risk touch-target slice.

### DIFF
--- a/frontend/src/components/TopNav.tsx
+++ b/frontend/src/components/TopNav.tsx
@@ -138,8 +138,8 @@ export function TopNav() {
             aria-expanded={mobileOpen}
             onClick={() => setMobileOpen((o) => !o)}
             style={{
-              padding: 6, display: "flex", flexDirection: "column", gap: 4,
-              width: 28, alignItems: "center",
+              padding: 6, display: "flex", flexDirection: "column", justifyContent: "center", gap: 4,
+              minWidth: 44, minHeight: 44, alignItems: "center",
             }}
           >
             {[0, 1, 2].map((i) => (
@@ -215,7 +215,7 @@ export function TopNav() {
               title="Settings"
               style={{
                 display: "inline-flex", alignItems: "center", justifyContent: "center",
-                width: 32, height: 32, borderRadius: "var(--r-sm)",
+                width: 44, height: 44, borderRadius: "var(--r-sm)",
                 color: pathname.startsWith("/settings") ? "var(--text)" : "var(--text-muted)",
                 transition: "background var(--dur-fast) var(--ease), color var(--dur-fast) var(--ease)",
               }}
@@ -231,7 +231,7 @@ export function TopNav() {
                 title="Admin"
                 style={{
                   display: "inline-flex", alignItems: "center", justifyContent: "center",
-                  width: 32, height: 32, borderRadius: "var(--r-sm)",
+                  width: 44, height: 44, borderRadius: "var(--r-sm)",
                   color: pathname.startsWith("/admin") ? "var(--text)" : "var(--text-muted)",
                   transition: "background var(--dur-fast) var(--ease), color var(--dur-fast) var(--ease)",
                 }}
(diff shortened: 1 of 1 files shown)


AndresL230 (commented):
Touch-target math all checks out (44px children fit the 56px bar, hamburger bars still center, menu rows compute correctly, nothing depends on the old widths). Two notes before merge:

**1. Settings/Admin enlargement isn't mobile-scoped.** The PR body says these are "genuine mobile touch surfaces", but the two icon links have no `isMobile` gate — their only gates are `isAuthenticated`/`isAdmin` — so the 32→44 change applies on desktop too: 44×44 hover-highlight boxes around 16px icons next to the 36px avatar chip, and the right cluster grows 12–24px wider on every authed desktop page. Either gate it the way the file already does elsewhere (e.g. lines 126–127, 188):

```tsx
width: isMobile ? 44 : 32, height: isMobile ? 44 : 32,
```

…or update the PR body to own the desktop change explicitly. Personally fine with either, just make scope and code agree.

**2. (Optional nit, non-blocking)** `44` is now hardcoded in four spots, and the deferred SideNav half of #110 will need it again. A fixed `--touch-target: 44px` token would tie them together — but note it should *not* live with the density-scaled `--pad-*`/`--row-h` tokens, since a WCAG minimum must not shrink under compact density. Fine to defer this to the token-unification work.

Also FYI `boxSizing: "border-box"` on the menu rows is redundant (globals.css already sets `* { box-sizing: border-box; }`) — harmless.

AndresL230 (approved):
Approving — the 44px claim is verified, not taken on faith.

Rather than reading the diff and assuming, I rebuilt the mobile `TopNav` as a standalone replica (same inline styles, same global reset, real Spectral/DM Sans webfonts) and measured computed geometry in Chrome:

| Control | Tappable element | Measured |
|---|---|---|
| Hamburger (`TopNav.tsx:136-159`) | the `<button>` carrying `onClick` | **44.00 × 44.00** |
| Settings (`:212-226`) | `<Link>` → `<a>` | **44.00 × 44.00** |
| Admin (`:228-242`) | `<Link>` → `<a>` | **44.00 × 44.00** |
| Mobile menu rows (`:469-485`) | `<Link>` → `<a>` | **238.67 × 44.00** |

The usual failure mode — padding a wrapper while the actual `<a>`/`<button>` stays small — is not present here; the constraints sit directly on the tappable elements. `globals.css:157` (`* { box-sizing: border-box }`) is load-bearing: it makes `minWidth/minHeight: 44` true outer dimensions, so the hamburger's `padding: 6` is absorbed inside the box rather than added to it. The added `justifyContent: "center"` correctly re-centers the bars.

**320px (iPhone SE) fits.** Admin worst case: `12px padding ×2 + 8px gap ×2 + 44 burger + 85.13 logo + 140 right cluster = 309px` → 11px slack, one line, no wrap or clipping.

I also tested a hypothesis that would have sunk this — the icons use `width: 44` without `flexShrink: 0`, which normally means silent shrinkage under compression. They hold at 44.0 at 320/300/280px and with a hypothetical extra icon, because flex items default to `min-width: auto`. Non-issue, disproved.

Scoping is correct: `TopNav` renders only via `ShellFrame` (warm app-shell); the cool `.landing-page` surface goes through `(public)/layout.tsx` and never renders it. And on mobile `useSidebar` is always false, so this is genuinely the mobile nav.

Deliberate deferrals are accurate and #110 correctly stays open: `SideNav` targets (desktop, mouse-driven) and the SSR desktop FOUC (`useIsMobile.ts:8` is `useState(false)`, so phones get the desktop shell then snap on hydration) both remain.

One nit, not worth holding: `:474` `boxSizing: "border-box"` is redundant given the global reset — though it reads as useful local documentation of intent. No regression test, but jsdom has no layout engine, so a meaningful assertion isn't achievable there anyway; only an inline-style string match, which would just restate the diff.

Issue #111, opened 2026-06-08
[P2] Optimize runtime animations + add lazy/next-image loading


AndresL230:
**Severity: P2** · Dimension: Performance · From `docs/frontend-ui-audit.md`

Bundle/lazy-load discipline is strong (three.js, mermaid, d3, MarkdownChat all lazy-loaded; expensive components memoized). The gaps are hand-rolled runtime animation and images bypassing Next's pipeline.

### Animation hot spots
- **Landing hero canvas** — `page.tsx:404-446` — 60fps RAF runs a nested O(N²) link pass over all nodes **and** sets `ctx.shadowBlur` per node (very expensive). **No `prefers-reduced-motion` guard anywhere in `page.tsx`** — reduced-motion users still burn full CPU/GPU. → gate RAF behind reduced-motion; drop `shadowBlur` for a pre-rendered glow sprite; cap the link pass.
- **Floating-cards RAF** — `page.tsx:507-516` — calls `querySelectorAll('.floating-card')` + re-parses `dataset` floats every frame. → resolve the NodeList + parse once.
- **AtmosphericBackdrop** — `AtmosphericBackdrop.tsx:93-139` (mounted app-wide via `ShellFrame.tsx:30,66`) — repaints 14 full-viewport radial gradients at 60fps on a CPU canvas, forever. Reduced-motion is handled; everyone else never throttles. → pre-bake each orb to an offscreen canvas + `drawImage`; consider ~30fps cadence.
- **KnowledgeGraph2D** — `KnowledgeGraph2D.tsx:172-174,306` — d3 `.on("tick", forceRerender)` reconciles the whole SVG each tick; `onPointerMove` calls `setTooltipPos` on every move; edges do O(E·N) `find()` lookups (`:397-398`). → drive positions via refs/direct DOM writes; only set tooltip when hovered; precompute an id→node Map.
- **Spotlight card** — `page.tsx:543-547` — `getBoundingClientRect()` on every mousemove (P3). → cache on `mouseenter`/resize.

### Images
[...]

PR #312, opened 2026-07-03
perf(images): lazy-load below-the-fold remote images (#111, part 1)

Jose-Gael-Cruz-Lopez:
Part of #111 (image-loading slice). Adds `loading="lazy"` to the 6 remote/user-content `` elements that render below the fold: Social chat attachments, MarkdownChat markdown images, Settings cosmetic overlay + banner, Admin cosmetic preview + catalog thumbnails. The fixed 28×28 catalog thumb also gets `width`/`height` attrs to reserve space.

Responsive images get lazy only — forced fixed dims would distort them; their space is already reserved by the parent.

**Verified locally:** `npm run typecheck` clean; `eslint` 0 errors on the changed files (pre-existing `no-img-element` warnings unchanged).

**Deliberately not here** (the rest of #111):
- Runtime-animation perf — landing-canvas reduced-motion guard, per-frame `shadowBlur`/`querySelectorAll`, atmospheric-orb throttle, KnowledgeGraph2D per-tick re-render. Real work, separate PR.
- Full `next/image` — non-trivial on OpenNext/Cloudflare (no built-in optimizer; needs a custom loader).

So #111 stays open; this is the low-risk lazy-loading slice.

### DIFF
--- a/frontend/src/components/MarkdownChat.tsx
+++ b/frontend/src/components/MarkdownChat.tsx
@@ -199,6 +199,7 @@ const COMPONENTS: Components = {
     <img
       src={typeof src === "string" ? src : undefined}
       alt={alt || ""}
+      loading="lazy"
       style={{
         maxWidth: "100%",
         height: "auto",
--- a/frontend/src/components/screens/Admin.tsx
+++ b/frontend/src/components/screens/Admin.tsx
@@ -938,7 +938,7 @@ function CosmeticsTab() {
               />
             </div>
             {form.asset_url && (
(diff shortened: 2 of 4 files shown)


AndresL230 (approved):
Reviewed — approving. All 6 lazy sites verified: the Settings/Admin previews sit in explicitly sized containers, the catalog thumb's new width/height attrs match its styles, coverage sweep found no missed remote images, and the next/image deferral is accurate (no loader configured for OpenNext/Cloudflare).

One known trade-off, accepted rather than blocking: room-chat attachments have no reserved height, so scrolling up through history now pops images in as they load — invisible-ish on Chrome/Firefox (scroll anchoring), visible jumps on Safari/iOS (no anchoring). Can't be fixed here since `room_messages` stores no image dimensions; filed #315 to capture width/height at upload and reserve an aspect-ratio box.

Issue #109, opened 2026-06-08
[P1] Make Admin / Gradebook / Settings responsive + adopt shared Dialog


AndresL230:
**Severity: P1** · Dimension: Responsive · From `docs/frontend-ui-audit.md`

Consumer flows (Dashboard, Learn, Tree, Study, Library, Calendar, Social) reflow well via the `useIsMobile` hook — but the data-dense **Admin, Gradebook, and Settings** surfaces were built desktop-only and never got the hook, and most modals bypass the shared `Dialog`.

### Non-responsive pages
- **Admin** — `screens/Admin.tsx:398,629,904` — hard-coded `gridTemplateColumns:"minmax(280px,360px) 1fr"`, no `isMobile` branch; overflows ≤375px. → collapse to `1fr` below 768px, or `repeat(auto-fit, minmax(min(280px,100%),1fr))`.
- **Gradebook** — entirely non-responsive (0 `useIsMobile` hits). Modals at fixed `minWidth:420/460/360`: `Gradebook/AssignmentModal.tsx:76`, `EditWeightsModal.tsx:62`, `SyllabusUploadFlow.tsx:102`, `LetterScaleEditor.tsx:50` overflow phones. → `width:min(420px,100%-32px)`; route through `Dialog`.
- **Settings** — `screens/Settings.tsx:259,311` — fixed `gridTemplateColumns:"180px 1fr"`, no mobile branch (P3). → stack below ~600px.

### Modal layer
- `Dialog.tsx` has no `useBodyScrollLock` → scroll-bleed behind overlays on mobile. The hook exists but only 5 components use it. → call `useBodyScrollLock(open)` inside `Dialog` so every consumer inherits it.
- ~6 hand-rolled modals (all of Gradebook, DocumentUpload, ManageCourses) bypass `Dialog`'s responsive width + focus trap → migrate them.

### Acceptance
- [ ] Admin / Gradebook / Settings reflow to one column on mobile
- [ ] Grade modals fit a 360px viewport
- [ ] `Dialog` locks body scroll; hand-rolled modals migrated to it

PR #313, opened 2026-07-03
fix(ui): lock body scroll in shared Dialog (#109, part 1)

Jose-Gael-Cruz-Lopez:
Part of #109 (AC3 slice). `Dialog.tsx` never called the existing `useBodyScrollLock` hook, so any modal built on the shared Dialog let the mobile background scroll-bleed behind the overlay. This wires `useBodyScrollLock(open)` into `Dialog` — the exact pattern the 5 other portal modals already use.

**Verified locally:** `npm run typecheck` clean; `eslint` clean on `Dialog.tsx`.

**Not here** (the substance of #109, separate work): making Admin/Gradebook/Settings responsive (they use inline `gridTemplateColumns` with no mobile branch) and migrating the ~8 hand-rolled Gradebook modals onto the shared `Dialog`. So #109 stays open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved modal behavior by preventing the page behind an open dialog from scrolling.
  * Reduced background scroll bleed while the dialog is visible.


AndresL230 (changes requested):
The wiring itself is right (hook signature, rules of hooks, SSR safety, cleanup on unmount all check out), but this introduces the first reachable **overlapping-lock** path, and `useBodyScrollLock` can't handle it.

`useBodyScrollLock.ts:6-8` is a naive save/restore with no refcount:

```ts
const prev = document.body.style.overflow;
document.body.style.overflow = "hidden";
return () => { document.body.style.overflow = prev; };
```

Concrete interleave (verified end-to-end in code):

1. On `/study`, open the flashcard import modal (the shared Dialog's consumer) → lock saves `prev=''`, sets `hidden`.
2. `FeedbackFlow` is mounted globally in `ShellFrame` and **auto-opens on its passive 45s timer** (authenticated + 3-day cooldown, zIndex 200) → its lock saves `prev='hidden'`.
3. `FlashcardImportModal.commit()` calls `onClose()` after its `await`, so the Dialog closes with no user click → Dialog's cleanup restores `overflow=''` while the feedback modal is still open → scroll-bleed behind it (the exact bug this PR fixes).
4. User dismisses feedback → its cleanup restores its saved `'hidden'` → **body stuck at `overflow:hidden` with no modal open**; page unscrollable until navigation.

The 45s timer firing while someone is pasting/generating flashcards is entirely plausible, so this isn't theoretical.

Suggested fix, at the hook level so all 6+ consumers benefit: make `useBodyScrollLock` refcounted — module-level counter, save the original overflow on 0→1, restore it only on 1→0. Then this PR's Dialog wiring is safe as-is, no change needed in Dialog.tsx.

AndresL230:
Re-reviewed as part of a sweep across the open `fix/*` PRs. Keeping this open for rework — but there's a second problem that changes what the fix should be, so please read this before addressing the existing review.

### 1. The existing review is still outstanding

The refcount request on `useBodyScrollLock` hasn't been addressed — noting it for completeness rather than as a nag. The branch's only commit (`0576956`, 02:57Z) predates the review (04:48Z), and the hook is byte-for-byte unchanged. That interleave is real; I re-derived it independently:

`FeedbackFlow` is mounted globally by `ShellFrame` in **both** layout branches and auto-opens on a 45s timer at `z-index: 200` (above Dialog's `100`). `FlashcardImportModal` calls `onClose()` after `await importCommit(...)` — so the Dialog can close on its own, with no user click, at an arbitrary moment. Sequence:

1. Dialog opens → saves `prev = ''`, sets `hidden`
2. FeedbackFlow auto-opens → saves `prev = 'hidden'`, sets `hidden`
3. Dialog's import completes → cleanup restores `''` — **unlocks while feedback is still open**
4. Feedback closes → cleanup restores `'hidden'` — **with no modal open at all**

Body is now stuck at `overflow: hidden` until navigation. The `if (!active) return` early-return means the false→true transition registers no cleanup, so the stale `'hidden'` really does get captured.

### 2. The new problem: the lock can't fix the bleed it targets

`ShellFrame.tsx:25-26` and `:61-62` root both layouts at:

```js
height: "100vh",
overflow: "hidden",
```

with `<main>` at `:37-38` / `:73-74` as `flex: 1; overflowY: "auto"`.

So **`<body>` is not the scroll container in the app shell** — its only child is exactly viewport-height, and `globals.css` sets no body overflow. `document.body.style.overflow = 'hidden'` acts on an element that never scrolls.

Dialog's only consumer is `FlashcardImportModal` → `Study.tsx`, which sets `minHeight: 100vh` inside a shorter `<main>`. **The thing that bleeds is `<main>`, not `<body>`.** So on the one screen this ships to, the lock is a no-op.

Two reviewers reached this independently from different angles — one via ShellFrame's scroll container, one via the modal portal analysis — so I'm fairly confident in it.

To be fair to the change: on mobile, `height: 100vh` can exceed the visual viewport under a dynamic Safari toolbar, so body may scroll by that sliver, which this would suppress. But that's not "background content scrolls behind the modal." And separately, `body { overflow: hidden }` doesn't prevent touch scrolling on iOS anyway — the platform the description names.

### 3. #335 interacts with this

#335 (`fix(ui): pin full-height shell screens`) converts the full-height screens from `100vh` to `100%` so they stop overflowing `<main>` by the 56px `TopNav` height. No textual conflict — it doesn't touch `Dialog.tsx` or any `overflow` property. But it further cements `<main>` as the sole scroll container, and it removes most of the bleed on shell screens as a side effect.

### Suggested shape

1. **Refcount the hook** as the original review asked — module-level counter, save original overflow on 0→1, restore only on 1→0. That fixes the freeze for all consumers.
2. **Retarget the lock.** Lock the `<main id="main-content">` scroller (or expose it via a shell-level context) instead of `document.body`. Worth checking against #335 first — it may make the Study-screen case moot, in which case this becomes purely about the pre-auth/landing pages where `body` genuinely does scroll.
3. While you're in here: **nine other components hand-roll this lock** (`SignInModal`, `DocumentUploadModal`, `ManageCoursesModal`, `Calendar`, `Settings`, `Dashboard`, `Tree`, and `(public)/page.tsx` twice), and they're inconsistent — `Calendar`/`Settings` save-and-restore the prior value; `Dashboard`/`Tree`/`DocumentUploadModal` reset to `""` unconditionally, which breaks stacking the same way. A refcounted hook is the right place to converge them, which makes #109 a bigger win than this diff.

The instinct is right and #109 is real — the target just needs to move.

AndresL230:
@Jose-Gael-Cruz-Lopez pushed 5 commits here as part of a review sweep. **Your commit `0576956` is preserved** — I merged `main` in rather than rebasing, specifically so nothing of yours could be stranded. Fast-forward only.

Fair warning: this outgrew the original 5 lines to 18 files, +369/−48. I think that's the right size for what #109 actually needs, but say the word and I'll carve it back.

### Both problems are real; I re-verified each against `origin/main`

Given a stale-branch mixup burned me earlier today, I checked every claim against `main` rather than this branch (which was **73** commits behind). All of it is live:

**1. The clobbering bug your review caught.** `useBodyScrollLock` on main is byte-for-byte the naive impl. Now refcounted — but **per element**, via `Map<HTMLElement, {count, prevOverflowX, prevOverflowY}>`. A shared counter would let a `<body>` lock hold `<main>` hostage. Only 0→1 captures, only 1→0 restores. This also fixes **5 existing consumers** that silently inherited the freeze (`DisclaimerModal`, `FeedbackFlow`, `ReportIssueFlow`, `SessionFeedbackFlow`, `Library`), plus the 9 hand-rolled sites — all migrated, none deferred.

**2. The lock was aimed at the wrong element.** `ShellFrame` roots both layouts at `overflow:hidden` with `<main>` as `overflowY:auto`, so `<body>` never scrolls in the shell and `document.body.style.overflow = "hidden"` was a no-op on the one screen Dialog ships to (`FlashcardImportModal` → `Study.tsx:655`). Now resolves `document.querySelector("[data-scroll-container]") ?? document.body` at lock time.

Chose a DOM query over a shell context because `SignInModal` lives *only* pre-auth while others live only in-shell — a query gets both right with no provider wiring. Used a new `data-scroll-container` attribute rather than reusing `#main-content` because that id is an **accessibility** contract (skip-link + focus target), and piggybacking scroll semantics on it would couple two unrelated concerns. The `(public)` layout is a bare `<div>`, so the `<body>` fallback is automatically correct pre-auth — those two locks are unchanged in behavior.

### The part worth reading — a bug that would have been worse than the original

The hook touches only the `overflow-x`/`overflow-y` **longhands**, never the `overflow` shorthand. That is deliberate and load-bearing.

`ShellFrame` sets `overflow-y: auto` **inline via React**. In a real browser, setting `style.overflow = "hidden"` and later restoring the shorthand to its previous value (`""`) **deletes that inline `overflow-y: auto` outright** — leaving `<main>` permanently unscrollable after any modal closes, with no re-render to put it back. The obvious implementation trades a scroll-bleed for a hard freeze.

**jsdom does not model this** — it reports `overflow-y: auto` surviving `overflow: hidden`. So no test in this repo could have caught it. It was found by probing the CSSOM directly.

### Tests: 11 new, and they bite

`frontend/src/lib/useScrollLock.test.ts` — overlapping locks in both release orders, 3-deep nesting, `false→true→false` while another lock is held, unmount-while-active, per-element `<main>`/`<body>` isolation, shell-vs-pre-auth targeting, and exact round-trip of the inline `overflow-y: auto`.

I mutation-tested them rather than trusting that they pass: stripping the refcount fails **4 of 11**; reverting to the naive hook wholesale fails **8 of 11**. They pin the reported bug, not the implementation's shape.

**But no test proves the bleed is fixed.** jsdom has no layout engine — nothing scrolls or paints, so only style bookkeeping is observable. Someone should drive `/study?mode=cards` → open the import modal → confirm `<main>` doesn't scroll → then let `FeedbackFlow` fire on its 45s timer over it and confirm no freeze. That is the actual acceptance test and it needs a browser.

### Note on ordering

#335 merged while this was in flight and touched the same files (`ShellFrame`, `Settings`). I cherry-picked onto current `main` and verified both survived: ShellFrame has `100dvh` (#335) **and** `data-scroll-container` (#313); Settings has `FullHeightScreen` (#335) **and** `useScrollLock` (#313). Also fixed a comment #335 left saying `100vh`.

Worth knowing: **#335 reduces the need for this** — with screens now filling `<main>` exactly, there's less left to bleed. This still matters for the freeze bug and the pre-auth pages.

### Nit I left in

`__resetScrollLocksForTests` is a test-only export in production code. Mildly smelly; the alternative was relying on every test unmounting cleanly, which is more fragile. Open to opinions.

AndresL230 (approved):
Approving. The original instinct and #109 were right — the target and the refcount just needed fixing, and both are done above. CI green on `5e2cee2`, 88 tests passing (11 new), and I mutation-tested the new tests rather than trusting them: stripping the refcount fails 4/11, reverting to the naive hook fails 8/11. They pin the bug, not the implementation.

Restating the one thing the green check does not cover: **no test here proves the scroll-bleed is fixed.** jsdom has no layout engine. Worth driving `/study?mode=cards` → open the import modal → confirm `<main>` does not scroll → let `FeedbackFlow` fire on its 45s timer over it → confirm no freeze. That is the real acceptance test.

PR #335, opened 2026-07-13
fix(ui): pin full-height shell screens so only the body scrolls (#331)

### LINKED ISSUE #331: Chat/Tutor: top nav bar scrolls out of view in horizontal-nav mode
## Summary

In the **chat/Tutor feature** (`/learn`), when the layout preference is the **horizontal top nav bar** (not the sidebar), the page becomes scrollable and the sticky top nav scrolls out of view. The nav should stay fixed and only the chat body should scroll.

## Repro
1. Set layout preference to the horizontal **top nav** shell (not sidebar).
2. Open `/learn` and start / resume a tutor session.
3. Scroll — the whole screen scrolls and the 56px top nav bar disappears off the top.

Only happens in **horizontal-nav mode**; the sidebar layout is unaffected.

## Root cause (grounded in code)

`ShellFrame` (`frontend/src/components/ShellFrame.tsx`) builds the horizontal layout as a `100vh` flex **column**: `TopNav` (fixed 56px, `position: sticky`) + `<main flex:1 overflowY:auto>`. So `main`'s usable height is `100vh − 56px`.

But the chat screen hardcodes `height: 100vh` on its root containers:
- `frontend/src/components/screens/Learn.tsx:434` (entry screen)
- `frontend/src/components/screens/Learn.tsx:503` (active session)

A `100vh` child inside a `100vh − 56px` container overflows `main` by exactly the nav height, so `main` becomes scrollable and the content (and with it the visual anchor of the nav) shifts. In **sidebar** mode the `SideNav` is a full-height *row* sibling, so `100vh` matches the viewport and the bug doesn't manifest — which is why it's horizontal-mode-only.

`TOP_NAV_HEIGHT` (56) is exported from `TopNav.tsx:75` but is **never** used in any content-height calculation.

## Fix direction
- The chat screen should **fill its container** (`height: 100%` / rely on `main`'s `flex:1`) instead of `height: 100vh`, or subtract `TOP_NAV_HEIGHT` in horizontal mode.
- **Same `height: 100vh` pattern affects other screens** and likely has the same bug: `Quiz.tsx`, `Library.tsx`, `notetaker/page.tsx`, `course-planner/page.tsx`, `Social.tsx`. Worth fixing generally (a shared full-height wrapper) rather than one-off per screen.

## Acceptance
- In horizontal-nav mode on `/learn`, the top nav stays fixed and only the chat message area scrolls; no whole-page scroll pushes the nav off-screen.

Darkest-Teddy:
## Fix: pin full-height shell screens so only the body scrolls

**Closes #331**

### The bug, in code

`ShellFrame` (top-nav layout) — `<main>` is `100vh − navHeight`:

```tsx
<div style={{ display: "flex", flexDirection: "column", height: "100vh", overflow: "hidden" }}>
  <TopNav />                                  {/* 56px */}
  <main style={{ flex: 1, overflowY: "auto" }}>{children}</main>  {/* 100vh − 56px */}
</div>
```

But the screens forced the **full viewport** inside that shorter `<main>`:

```tsx
// Learn.tsx, Quiz.tsx, Library.tsx, Social.tsx, notetaker, course-planner
<div style={{ display: "flex", height: "100vh", ... }}>   // ← taller than <main> by 56px
```

`100vh > (100vh − 56px)` → `<main>` overflows → the whole page scrolls, dragging the chat header/input out of view. The sidebar layout was fine because `SideNav` sits *beside* `<main>` (row), so `<main>` gets the full `100vh` and `100vh` matched exactly.

### The fix

Fill the parent, not the viewport — `100vh` → `100%`, wrapped in one shared component:

```tsx
// components/FullHeightScreen.tsx  (new)
export function FullHeightScreen({ children, direction = "column", className, style }: {
  children: React.ReactNode;
  direction?: "row" | "column";
  className?: string;
  style?: React.CSSProperties;
}) {
  return (
    <div
      className={className}
      style={{ display: "flex", flexDirection: direction, height: "100%", minHeight: 0, ...style }}
    >
      {children}
    </div>
  );
}
```

`height: 100%` resolves against `<main>` (which has a definite height in **both** layouts), so top-nav is fixed and sidebar is unchanged.

### The diffs

```diff
// Learn.tsx (entry + active session), Quiz.tsx, course-planner/page.tsx
-    <div style={{ display: "flex", height: "100vh", flexDirection: "column" }}>
+    <FullHeightScreen>

// Library.tsx, Social.tsx (row layouts)
-    <div style={{ display: "flex", height: "100vh" }}>
+    <FullHeightScreen direction="row">

// notetaker/page.tsx  (root + loading/empty states)
-          height: "100vh",
+          height: "100%",

// Study.tsx  (stays scrollable, now bounded to <main>)
-    <div style={{ display: "flex", flexDirection: "column", minHeight: "100vh" }}>
+    <div style={{ display: "flex", flexDirection: "column", minHeight: "100%" }}>
```

Standalone pages (`about`, `terms`, `privacy`, `careers`, `auth`, `pending`, `Onboarding`) keep `100vh` — they own the full viewport, correctly.

### Files changed

```
 frontend/src/components/FullHeightScreen.tsx     | 47 ++++++++++++++++++  (new)
 frontend/src/components/screens/Learn.tsx        |  9 +++--
 frontend/src/components/screens/Quiz.tsx         |  5 ++-
 frontend/src/components/screens/Library.tsx      |  5 ++-
 frontend/src/components/screens/Social.tsx       |  5 ++-
 frontend/src/components/screens/Study.tsx        |  2 +-
 frontend/src/app/(shell)/course-planner/page.tsx |  5 ++-
 frontend/src/app/(shell)/notetaker/page.tsx      |  6 +--
 8 files changed, 68 insertions(+), 16 deletions(-)
```

### #331 checklist
- ✅ Top nav fixed; only the body scrolls
- ✅ `height: 100vh` → `100%`
- ✅ Shared full-height wrapper (`<FullHeightScreen>`) applied across affected screens
- ✅ Sidebar layout unaffected (`100%` resolves to the same `100vh` there)

### Testing
```
npx tsc --noEmit   # passes for changed files (pre-existing react-force-graph-3d error in KnowledgeGraph3D.tsx is unrelated)
npx eslint <changed files>   # 0 errors (2 pre-existing warnings)
```
Manual check still recommended: `/learn`, `/quiz`, `/library`, `/social`, `/notetaker` in top-nav layout (header/input pinned, body scrolls) + sidebar layout unchanged.

### DIFF
--- a/frontend/src/app/(shell)/course-planner/page.tsx
+++ b/frontend/src/app/(shell)/course-planner/page.tsx
@@ -1,10 +1,11 @@
 "use client";
 import { TopBar } from "@/components/TopBar";
 import { Icon } from "@/components/Icon";
+import { FullHeightScreen } from "@/components/FullHeightScreen";
 
 export default function CoursePlannerPage() {
   return (
-    <div style={{ display: "flex", height: "100vh", flexDirection: "column" }}>
+    <FullHeightScreen>
       <TopBar title="Course Planner" />
       <main
         style={{
@@ -44,6 +45,6 @@ export default function CoursePlannerPage() {
           building it — check back soon.
         </div>
       </main>
-    </div>
+    </FullHeightScreen>
   );
 }
--- a/frontend/src/app/(shell)/notetaker/page.tsx
+++ b/frontend/src/app/(shell)/notetaker/page.tsx
@@ -480,7 +480,7 @@ export default function NotetakerPage() {
           display: "flex",
           alignItems: "center",
           justifyContent: "center",
-          height: "100vh",
+          height: "100%",
           color: "var(--text-muted)",
           fontSize: 13,
         }}
@@ -499,7 +499,7 @@ export default function NotetakerPage() {
             flexDirection: "column",
             alignItems: "center",
             justifyContent: "center",
-            height: "100vh",
+            height: "100%",
             padding: 32,
             textAlign: "center",
             gap: 12,
@@ -556,7 +556,7 @@ export default function NotetakerPage() {
   }
 
   return (
-    <div style={{ display: "flex", flexDirection: "column", height: "100vh" }}>
+    <div style={{ display: "flex", flexDirection: "column", height: "100%" }}>
       <div
         style={{
           flex: 1,
--- a/frontend/src/components/FullHeightScreen.tsx
+++ b/frontend/src/components/FullHeightScreen.tsx
@@ -37,7 +37,6 @@ export function FullHeightScreen({
         display: "flex",
         flexDirection: direction,
         height: "100%",
-        minHeight: 0,
         ...style,
       }}
     >
--- a/frontend/src/components/FullHeightScreen.tsx
+++ b/frontend/src/components/FullHeightScreen.tsx
@@ -6,7 +6,7 @@ import React from "react";
  * Full-height root for screens rendered inside `ShellFrame`'s `<main>`.
  *
  * ShellFrame's horizontal-nav layout stacks the 56px `TopNav` above `<main>`
- * in a `100vh` flex column, so `<main>` is only `100vh - 56px` tall. Screens
+ * in a `100dvh` flex column, so `<main>` is only `100dvh - 56px` tall. Screens
  * that hardcoded `height: 100vh` were therefore taller than their container,
  * which overflowed `<main>` and made the whole page scroll — dragging chat
  * headers/inputs out of view instead of scrolling just the body (issue #331).
--- a/frontend/src/components/FullHeightScreen.tsx
+++ b/frontend/src/components/FullHeightScreen.tsx
@@ -1,0 +1,47 @@
+"use client";
+
+import React from "react";
+
+/**
+ * Full-height root for screens rendered inside `ShellFrame`'s `<main>`.
+ *
+ * ShellFrame's horizontal-nav layout stacks the 56px `TopNav` above `<main>`
+ * in a `100vh` flex column, so `<main>` is only `100vh - 56px` tall. Screens
+ * that hardcoded `height: 100vh` were therefore taller than their container,
+ * which overflowed `<main>` and made the whole page scroll — dragging chat
+ * headers/inputs out of view instead of scrolling just the body (issue #331).
+ *
+ * Filling the parent with `height: 100%` pins the screen to `<main>` exactly,
+ * so inner regions scroll internally. `<main>` has a definite height in both
+ * the sidebar and top-nav layouts, so the percentage resolves correctly in
+ * each — keeping the sidebar layout unaffected.
+ *
+ * Use this instead of `height: 100vh` for any full-height shell screen.
+ */
+export function FullHeightScreen({
+  children,
+  direction = "column",
+  className,
+  style,
+}: {
+  children: React.ReactNode;
+  /** Flex direction of the root. Defaults to "column". */
+  direction?: "row" | "column";
+  className?: string;
+  style?: React.CSSProperties;
+}) {
+  return (
+    <div
+      className={className}
+      style={{
+        display: "flex",
+        flexDirection: direction,
+        height: "100%",
+        minHeight: 0,
+        ...style,
+      }}
+    >
+      {children}
+    </div>
+  );
+}
--- a/frontend/src/components/ShellFrame.tsx
+++ b/frontend/src/components/ShellFrame.tsx
@@ -11,6 +11,16 @@ import { AchievementUnlockWatcher } from "./AchievementUnlockWatcher";
 import { useLayoutPref } from "@/lib/useLayoutPref";
 import { useIsMobile } from "@/lib/useIsMobile";
 
+/**
+ * `100dvh`, not `100vh`: on iOS Safari `100vh` is the *large* viewport (the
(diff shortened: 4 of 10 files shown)


AndresL230:
@Darkest-Teddy heads up — I pushed 3 commits to this branch as part of a review sweep, and retargeted the base. Nothing of yours was rewritten (fast-forward only). Happy to back any of it out if you disagree.

### Base retargeted: `Gradebook` → `main`

This one was load-bearing. `Gradebook` was **already merged into `main`** via #241 on 2026-06-28 — the branch is a stale post-merge leftover sitting at `067b46e`. As targeted, **this PR would have merged into `Gradebook` and shipped nothing to users**, while closing successfully and looking fine.

It is not stacked and there is no merge-order constraint: `git diff origin/Gradebook...origin/scrolling` and `git diff origin/main...origin/scrolling` are byte-identical, so retargeting is a no-op on the diff.

### `3fa6cf5` — `Settings.tsx` had the same bug (the main one)

`Settings.tsx:232` had `height: calc(100vh - 112px)`, where 112px is the TopBar height. Root had no height, so `TopBar (112) + calc(100vh - 112)` = exactly `100vh` — correct in the sidebar layout, but 56px too tall inside the top-nav layout's `100vh - 56px` `<main>`. Result: two nested scrollbars on `/settings` and the bottom 56px below the fold, **including the delete-account controls on the `data` tab**. Mobile always takes the top-nav path.

Fixed it the way this PR fixes the others — `FullHeightScreen` + `flex: 1, minHeight: 0` — which also deletes the 112px magic number so the row can't silently break if TopBar is ever resized. (`minHeight: 0` is genuinely needed *here*, unlike in `FullHeightScreen` — see below — because the row is now a flex item with `overflow: visible`, so its automatic minimum size would otherwise refuse to shrink.)

### `1113035` — `100vh` → `100dvh` in `ShellFrame`

There was zero `dvh` usage in `frontend/src`. On iOS Safari `100vh` is the *large* viewport (toolbar collapsed), so with the toolbar expanded the shell runs past the visual viewport and hides its own bottom edge — the `/learn` composer most visibly, which is the exact element #331 is about.

This became load-bearing *because of this PR*: previously the 56px overflow gave incidental scroll slack to drag that edge into view; now screens fill `<main>` exactly with zero slack.

No `100vh` fallback, deliberately: a fallback needs two declarations of the same property, which a React inline style object can't express (it would mean moving the shell root into a CSS class), and it would buy nothing — `globals.css` already leans unguarded on `color-mix()` and `:has()` (14 hits) plus `overflow-x: clip`, all equal or *narrower* support than `dvh` (Baseline since 2022). Any browser needing the fallback is already broken by the stylesheet.

Worth noting `dvh`'s usual objection (value changes mid-scroll → resize thrash) doesn't apply here: the shell root is `overflow: hidden` and scrolling lives in inner panes, so the document never scrolls and the toolbar never collapses.

### `40fdc91` — dropped inert `minHeight: 0` from `FullHeightScreen`

Its documented parent is `<main>`, which has no `display` property → block container. So the root is a block-level box where `min-height: auto` already computes to 0. Verified all six call sites are direct children of `<main>` (App Router adds no DOM wrappers). Removed rather than commented, since inert CSS on a shared primitive reads as load-bearing and gets cargo-culted; callers can still pass it via `style`.

### One correction to my own review

I claimed `globals.css`'s `overscroll-behavior-y: none` / `min-height: 100vh` were on `body` and would blunt the `dvh` fix. **Wrong** — they're scoped to `.landing-page` (opens at `:740`), the pre-auth marketing page, and never touch the shell. `body` only sets margin/padding/font (`:160-161`). So the fix is fully effective, not half.

### On testing — read this before trusting the green check

**No test in this repo can verify any of this.** jsdom has no layout engine; `getBoundingClientRect` returns 0. There are zero `toHaveStyle`/`getComputedStyle` assertions in `frontend/src`. tsc + eslint + 68 vitest tests pass, but that only proves I didn't break component contracts — it is **not** evidence the pinning works. The correctness argument rests on `<main>` having a definite height in both layouts (sidebar: stretched in a row container; top-nav: `flex: 1` in a column container), which is what makes `height: 100%` resolve — the same premise the five already-merged screens rely on. Worth a real device check on `/learn` and `/settings`.

### Left alone

`Tree.tsx:311` and the Gradebook `--row-h` screens (`Landing.tsx:218`, `Course.tsx:555`) — pre-existing, tracked in #341, assigned to you. Notably `--row-h` is a *density* token (40/34/48px), not a nav height, so those offsets are semantically wrong *and* density-dependent.

AndresL230 (approved):
Approving. The original diff was sound — the 56px `TopNav` arithmetic checks out and all five converted screens' scroll chains terminate correctly in a `flex: 1` + `overflowY: auto` box, so nothing escapes into `<main>` and no double-scroll is introduced. The three blockers (dead base branch, missed `Settings.tsx`, `100vh` on iOS) are addressed in the commits above and CI is green on `40fdc91`.

Reiterating the caveat from my comment, because the green check overstates what was verified: **no test in this repo can observe a layout regression** — jsdom has no layout engine. The suite passing proves component contracts are intact, not that the pinning works. A real-device pass on `/learn` and `/settings` is worth doing.

PR #300, opened 2026-07-01
feat(flashcards): flashcard generation/import → Pydantic AI agent (#146)

Jose-Gael-Cruz-Lopez:
Closes #146 (Agent-migration epic #152, milestone #2).

## What
Routes all five flashcard LLM seams through one `flashcard_agent` — no `gemini_service` import remains in the flashcard path.

| Seam | Before | After |
|---|---|---|
| `extract_cards_from_image` (OCR-split) | `call_gemini` | `flashcard_agent` |
| `gemini_generate_cards` | `call_gemini` | `flashcard_agent` |
| `gemini_cleanup_cards` | `call_gemini` | `flashcard_agent` |
| `gemini_cloze` | `call_gemini` | `flashcard_agent` |
| `generate_flashcards` (main AI gen) | `gemini_service.generate_flashcards` | **moved** to `flashcard_import_service`, runs `flashcard_agent` |

## Notes
- New `agents/flashcard.py` — `Flashcards { cards: list[FlashCard{front, back}] }`; one agent, task instructions stay in the existing prompt templates (passed as the user message). Registered `flashcard` (`gemini-2.5-flash`) in `_providers`.
- `_run_flashcard_agent(prompt)` filters cards missing a side (preserving `_parse_card_json`'s behavior) and **degrades to `[]` on agent failure** — matching the old "bad LLM output → []" resilience (never raises).
- Behavior preserved: `gemini_cleanup_cards` still falls back to its input on empty cleanup; `extract_cards_from_image` still short-circuits on empty OCR without calling the agent.
- `generate_flashcards` moved off `gemini_service` (prompt-building verbatim); `routes/flashcards.py` imports it from `flashcard_import_service`.

## Testing
- Rewrote the 7 `call_gemini` patches in `test_flashcard_import_service.py` to mock `flashcard_agent`; added empty-filter, agent-failure→`[]`, cleanup-fallback, and `generate_flashcards` tests. Route tests are unchanged (they patch the whole functions).
- Full backend suite: **832 passed** (the 2 `test_storage_service` failures pre-exist on `main` — missing SUPABASE env). `ruff` clean.

Out of scope: the non-LLM parsers (`parse_xlsx`/`parse_anki_apkg`/`scrape_quizlet_url`/`dedup`), and deleting the rest of `services/gemini_service.py` (#151).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
[...]

PR #336, opened 2026-07-15
docs(flashcards): correct error-contract comment (#300 review follow-up)

Jose-Gael-Cruz-Lopez:
Follow-up to the code review of #300.

`_run_flashcard_agent`'s docstring claimed the error contract "mirrors the legacy `call_gemini` flashcard path." That's only true for the 4 import seams (which already returned `[]` on bad JSON). For `/generate`, the legacy path **raised** on malformed output (surfacing a 502); the unified agent path degrades it to `[]`. Reworded to state this is a deliberate contract change for `/generate`, not a faithful mirror.

Comment-only; no behavior change. `test_flashcard_import_service.py` 35/35 pass.

### DIFF
--- a/backend/services/flashcard_import_service.py
+++ b/backend/services/flashcard_import_service.py
@@ -228,12 +228,14 @@ _AGENT_RETRIES = 1
 def _run_flashcard_agent(prompt: str) -> list[Card]:
     """Run the flashcard agent on a rendered prompt and return front/back dicts.
 
-    Filters out cards missing either side. Error contract mirrors the legacy
-    ``call_gemini`` flashcard path:
+    Filters out cards missing either side. Error contract (unified across the
+    generate + import seams):
 
     * A model whose output can't satisfy the ``Flashcards`` schema is treated as
-      "bad LLM output" and degrades to ``[]`` — logged, never silent (this is
-      the old JSON-parse ``[]`` resilience).
+      "bad LLM output" and degrades to ``[]`` — logged, never silent. This
+      matches the import seams' long-standing bad-JSON ``[]`` resilience; note
+      it is a deliberate contract change for ``/generate``, whose legacy path
+      *raised* (surfacing a 502) on malformed output.
     * Transient provider errors (HTTP 429/5xx) are retried once with a ~2s
       backoff, then re-raised.
     * Every other failure — transport, missing GEMINI_API_KEY,


AndresL230 (approved):
Approving. Both headline claims verified against the pre-#300 code at `756d276`, and the new wording is strictly more accurate than what it replaces.

The old comment said the `/generate` path "mirrors the legacy `call_gemini` flashcard path" — that was flatly wrong. Legacy `/generate` (`gemini_service.generate_flashcards`) parsed with an unguarded `json.loads(cleaned)` and raised `ValueError`, which `routes/flashcards.py:203-204` caught as bare `Exception` → **502**. So it raised; it did not degrade to `[]`. The legacy *import* seams' `_parse_card_json` genuinely did return `[]` on `JSONDecodeError`. The correction is real.

**I verified the comment empirically rather than by inspection** — patched `flashcard_agent.run` and observed actual behavior:

| Injected failure | Actual result | Matches comment? |
|---|---|---|
| Schema retry exhaustion (`UnexpectedModelBehavior`) | `[]` → HTTP 200 | ✅ bullet 1 |
| `ModelHTTPError(503)` | retried once, then raised → 502 | ✅ bullet 2 |
| `ModelHTTPError(401)` | raised → 502 | ✅ bullet 3 |
| `RuntimeError` | raised → 502 | ✅ bullet 3 |
| **`ContentFilterError`** | **`[]` → HTTP 200** | ❌ not described |
| **Token-limit overrun** | **`[]` → HTTP 200** | ❌ not described |

Also confirmed `ModelHTTPError` is not a subclass of `UnexpectedModelBehavior` (siblings under `AgentRunError`), so the `except` ordering at `:251`/`:254` is safe and `result` can't be `None` at `:263`. 35/35 tests pass.

**Merging as-is, but bullet 1 is under-inclusive — and the real bug is the code, not the comment.** `ContentFilterError` **is** a subclass of `UnexpectedModelBehavior` (MRO: `ContentFilterError → UnexpectedModelBehavior → AgentRunError → RuntimeError`), so the `except` at `:251` swallows content-filter blocks too. That is not theoretical here: `pydantic_ai/models/google.py:152-158` maps **`RECITATION`** → `content_filter`, and Gemini returns `RECITATION` precisely when output closely reproduces source material — while `agents/flashcard.py`'s system prompt instructs the model to "Base card content on the material provided; do not invent facts it does not support."

Concrete: student generates flashcards from a textbook chapter → Gemini finishes with `RECITATION` → `POST /api/flashcards/generate` returns **`200 {"cards": []}`**. Zero cards, nothing to retry, indistinguishable from "no cards found." That is verbatim the P1 #300's review raised. Same for a token-limit overrun against the pinned `max_tokens=8192`, realistic on large OCR markdown.

The fix is to narrow the catch so `ContentFilterError` propagates → 502, not to broaden the sentence. Filed separately — no reason to hold a docs correction on it.

Two smaller nits, non-blocking: "a deliberate contract change for `/generate`" is incomplete (legacy `call_gemini` raised on empty/filtered responses *before* `_parse_card_json` ran, so the import seams changed too — they only returned `[]` on genuinely bad JSON); and bullet 3's "propagates → 502" holds for four of five seams — `/import/parse` catches `ValueError` → 422 first (`routes/flashcards.py:391-392`), currently harmless since everything re-raised derives from `RuntimeError`.

Issue #354, opened 2026-07-18
Shared google.genai client is unsafe across event loops — fresh-client sweep for remaining run_agent_sync routes


AndresL230:
## Summary

PR #349 (streaming tutor) uncovered that the **module-level shared `google.genai` client** (`backend/agents/_providers.py::_provider`) is **unsafe to use across event loops**. Its httpx connection pool caches connections bound to whatever event loop first opened them; when a later call reuses/closes such a connection on a different (or already-closed) loop, `asyncio.call_soon` fires on a dead loop and raises:

```
RuntimeError: Event loop is closed
```

`run_agent_sync()` (`backend/agents/_run.py`) drives agents on **throwaway per-request event loops** via `asyncio.run()` inside FastAPI's reused sync-handler threadpool threads. The first call in a thread opens a pooled connection on loop L1; L1 closes; the next call in that same thread trips over the stale connection → 500. (The same root cause caused the streaming-tutor crash fixed in #349.)

### Reproduction (pre-existing, on `main`)

```
POST /api/graph/{user_id}/concept-description   (repeat 3x)
→ 500, 200, 500   (flaky)
```
Confirmed on unmodified `main`. The 500s also surfaced as React errors in the Learn knowledge-map (concept descriptions failed to load).

## Already fixed in #349's branch (the pattern to follow)

Helpers added in `agents/_providers.py`:
- `fresh_google_model(name)` — a `GoogleModel` on a brand-new `GoogleProvider` (fresh client + connection pool).
- `fresh_model_for(task)` and `model_name_for(task)`.

Applied at two sites (fixes them — repeat calls now 5/5 → 200):
- `routes/graph.py` (`concept_describe`) → `model=fresh_model_for("concept_describe")`
- `services/course_context_service.py` (`course_summary`) → `model=fresh_model_for("course_summary")`

`run_agent_sync` was also hardened to fail cleanly (close the coroutine — no "never awaited" warning) when misused from within a running loop.

## Remaining call sites needing the same one-line fix

Each `run_agent_sync(<agent>.run(...))` below must pass `model=fresh_model_for("<task>")` (or `fresh_google_model("<name>")`):

| File:line | Agent | Task arg |
|---|---|---|
| `routes/study_guide.py:104` | `study_guide_agent` | `"study_guide"` |
| `routes/social.py:142` | `social_summary_agent` | `"social_summary"` |
[...]

PR #357, opened 2026-07-18
fix(agents): fresh google client for run_agent_sync agent routes (part of #354)

AndresL230:
Part of #354. Stacked on #349 (`feat/streaming-tutor`) — merge that first.

## Root cause

Same one as #349's streaming crash: the module-level shared `google.genai` client (`agents/_providers.py::_provider`) is **unsafe across event loops**. Its httpx pool caches connections bound to the loop that opened them; reusing/closing one on a different (or already-closed) loop trips `RuntimeError: Event loop is closed`.

`run_agent_sync()` drives agents on throwaway per-request loops (`asyncio.run`) inside FastAPI's reused sync-handler threadpool threads — so the 2nd call in a thread trips over the 1st call's stale connection.

**Reproduced on `main`:** `POST /api/graph/{user_id}/concept-description` returns a flaky `500 / 200 / 500` on repeat calls (surfaced as React errors in the Learn knowledge-map).

## What this does

- `concept_describe` (`routes/graph.py`) and `course_summary` (`services/course_context_service.py`) now run on a fresh client via `model=fresh_model_for(...)`.
- `run_agent_sync` hardened: when misused from a running event loop it now closes the coroutine (kills the "coroutine was never awaited" warning) and raises a clear error the caller degrades on — instead of blocking the loop or leaking.
- Adds `tests/test_run_agent_sync_loop.py`.

The `fresh_google_model` / `fresh_model_for` / `model_name_for` helpers land in #349.

## Verification

- `POST /api/graph/{u}/concept-description` × 5 in a row → **200 every time** (was 500/200/500 on `main`); 0 `Event loop is closed` in logs.
- Streaming (#349) still reaches `done` with graph deltas; non-streaming `/chat` unaffected.
- `161 passed` across the affected suites (learn, graph, quiz, social, study_guide, course_context, streaming, run_agent_sync).

## Not in this PR (remaining #354 scope)

The same one-line fresh-client fix still needs applying to the other sync-handler agent routes: `study_guide.py:104`, `social.py:142`, `documents.py:245`, `quiz.py:459`, `flashcard_import_service.py:249`, `main.py:222`. Tracked in #354.

### DIFF
--- a/backend/agents/_run.py
+++ b/backend/agents/_run.py
@@ -21,7 +21,29 @@ T = TypeVar("T")
 def run_agent_sync(coro: Coroutine[Any, Any, T]) -> T:
     """Run an agent coroutine to completion from synchronous code.
 
-    Must be called from a thread with no running event loop (the case for
-    FastAPI sync handlers and direct unit-test calls under TestClient).
+    Intended for FastAPI sync ``def`` handlers, which FastAPI runs in a worker
+    thread with no event loop — there ``asyncio.run`` drives the agent.
+
+    Some sync helpers are ALSO reached from *async* handlers via a sync call
+    chain (e.g. ``build_system_prompt`` -> ``get_course_context`` -> here, from
+    the async legacy chat / start-session paths). A loop is already running on
+    that thread, so ``asyncio.run`` can't be used and — critically — we must not
+    block the event loop on an LLM round-trip. The old code let ``asyncio.run``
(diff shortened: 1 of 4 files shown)


AndresL230:
Folding into #349 — same root cause (shared google client unsafe across event loops) and same helper. The commit now lives on `feat/streaming-tutor`, so this reviews and merges as one PR.

Issue #355, opened 2026-07-18
Duplicate subject_root node in /api/graph → React duplicate-key warnings in KnowledgeGraph2D


AndresL230:
## Summary

`GET /api/graph/{user_id}` returns the **subject-root node duplicated**. For `seed-user-demo`, `subject_root__seed-course-cs101` appears **twice** in `nodes` (observed in the raw payload). The Learn knowledge map renders `nodes.map(n => <li key={n.id}>...)`, so the duplicate id produces a React warning:

```
Encountered two children with the same key, `subject_root__seed-course-cs101`.
Keys should be unique ...
```

Six such warnings accumulated in the Next dev overlay during the #349 smoke (one per duplicated subject root across the seeded courses).

## Where

- **Rendering site:** `frontend/src/components/KnowledgeGraph2D.tsx:685` (`<li key={n.id}>`).
- **Root:** the duplicate originates in the graph read — subject-root synthesis or a join fan-out in `backend/routes/graph.py` / `backend/services/graph_service.py`. Confirm by counting `nodes[].id` in the raw `GET /api/graph/{user_id}` payload.

## Fix direction

Dedup by node `id` at the **source** (graph read) so the API never returns two nodes with the same id. A frontend-only key workaround is a band-aid — fix the payload.

## Acceptance criteria

- [ ] `/api/graph/{user_id}` returns unique `nodes[].id` (assert in a backend test against seed data).
- [ ] No React duplicate-key warning on `/learn` for a course that has a subject root.

## Notes
Pre-existing (visible before any streaming). Not caused by #349, but #349's map rendering surfaces it. Reproduce with the seeded `seed-user-demo` (`python -m db.seed_staging`).

_Filed by Claude Code during the PR #349 e2e-smoke investigation._

AndresL230:
Additional evidence from the #395 graph-integrity journey (PR #433), which reproduces this on a migrations-only local schema + rich seed:

`GET /api/graph/rich-user-active` → `nodes_total=18 nodes_unique=17 dup_node_ids=["subject_root__rich-course-cs101"] edges_total=25 edges_unique=20 dup_edges=5`

[...]

PR #358, opened 2026-07-18
fix(agents): guard run_agent_sync against running event loops (#354 follow-up)

AndresL230:
## Summary

Reworked 2026-07-29: the original fresh-client sweep + subject_root dedup were superseded on main by the #447–#454 batch (`_LoopSafeGoogleModel` in #453 solved cross-loop client safety properly; the #355 dedup landed with its own regression test). This PR now carries only the piece main still lacks:

- **`run_agent_sync` running-loop guard** — called from an async handler's sync call chain (e.g. `build_system_prompt` → `get_course_context`), it used to let `asyncio.run` raise opaquely and leak a "coroutine was never awaited" warning. Now it closes the coroutine and raises a clear `RuntimeError` the try/except-guarded callers degrade on.
- **`HEALTH_PROBE_MODEL`** constant in `agents/_providers.py`, used by `agents/health.py`, so probe sites can't drift.
- Regression tests for both loop paths (`tests/test_run_agent_sync_loop.py`).

The streaming-tutor interrupt/retry ADR that was on this branch moves to #349 (main's 0019 slot is now taken by the SAPLING_MODEL_MODE seam ADR).

## Testing

- `pytest tests/ -q`: 1181 passed, 27 skipped.
- `ruff check .`: clean (one pre-existing hit in an untracked local ops script only).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved agent execution errors when synchronous calls are made from an asynchronous context, with clearer guidance to use asynchronous execution.
  - Prevented coroutine cleanup warnings in unsupported execution scenarios.
  - Kept the connectivity health check aligned with the configured model.

- **Tests**
  - Added regression coverage for synchronous execution and active event-loop handling.

### DIFF
--- a/backend/agents/_providers.py
+++ b/backend/agents/_providers.py
@@ -95,6 +95,12 @@ _DEFAULTS: dict[AgentTask, str] = {
 }
 
 
+# The admin `/api/gemini-test` connectivity probe has no `_DEFAULTS` task of
+# its own; it just needs the cheapest round-trip. Pinned here so probe sites
+# (agents/health.py) can't drift from each other.
+HEALTH_PROBE_MODEL = "gemini-2.5-flash-lite"
+
+
 def _new_provider() -> GoogleProvider:
     """A fresh `GoogleProvider` (fresh `google.genai` client + httpx connection
     pool). CI and import-time tools don't have GEMINI_API_KEY set; the dummy
--- a/backend/agents/_run.py
+++ b/backend/agents/_run.py
@@ -25,8 +25,9 @@ def run_agent_sync(coro: Coroutine[Any, Any, T]) -> T:
     thread with no event loop — there ``asyncio.run`` drives the agent.
 
     Some sync helpers are ALSO reached from *async* handlers via a sync call
-    chain (e.g. ``build_system_prompt`` -> ``get_course_context`` -> here, from
-    the async legacy chat / start-session paths). A loop is already running on
+    chain (e.g. ``_legacy_chat`` -> ``apply_graph_update`` ->
+    ``update_course_context`` -> ``_generate_summary_with_gemini`` -> here,
+    and the same shape from the async upload pipeline). A loop is already running on
     that thread, so ``asyncio.run`` can't be used and — critically — we must not
     block the event loop on an LLM round-trip. The old code let ``asyncio.run``
     raise and leaked a "coroutine 'run' was never awaited" warning. Instead,
--- a/backend/agents/_run.py
+++ b/backend/agents/_run.py
@@ -21,7 +21,29 @@ T = TypeVar("T")
 def run_agent_sync(coro: Coroutine[Any, Any, T]) -> T:
     """Run an agent coroutine to completion from synchronous code.
 
-    Must be called from a thread with no running event loop (the case for
-    FastAPI sync handlers and direct unit-test calls under TestClient).
+    Intended for FastAPI sync ``def`` handlers, which FastAPI runs in a worker
+    thread with no event loop — there ``asyncio.run`` drives the agent.
+
+    Some sync helpers are ALSO reached from *async* handlers via a sync call
+    chain (e.g. ``build_system_prompt`` -> ``get_course_context`` -> here, from
+    the async legacy chat / start-session paths). A loop is already running on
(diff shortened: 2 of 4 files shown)


Darkest-Teddy (commented):
Reviewed the full diff, all PR comments (CodeRabbit ran Free-plan summary-only; no line findings), and ran the 3 affected test files locally in an isolated worktree — 48 passed, and CI is green on this head. The #354 fresh-client sweep, the run_agent_sync loop-guard, and the #355 subject-root dedup are all correct and well-covered (the AST guard in test_fresh_client_sweep.py is a nice revert-proof check). No bugs found; the two notes below are minor and non-blocking.

backend/agents/_providers.py
```diff
@@ -101,3 +101,38 @@ def model_for(task: AgentTask) -> GoogleModel:
...
+
+    A streaming tutor turn makes TWO sequential streamed model requests in one
+    run (reply → tool calls → continuation). Reusing the shared client's pooled
+    HTTP connection across them surfaced `RuntimeError: Event loop is closed`
+    in google-genai's httpx teardown (the pooled connection carried a stale
+    event-loop reference). A per-request client, created and used only on the
+    live request loop, sidesteps the cross-loop connection reuse.
+    """
+    provider = GoogleProvider(api_key=GEMINI_API_KEY or "dummy-key-for-import")
+    return GoogleModel(name, provider=provider)
```
Darkest-Teddy: Nit (resource hygiene, non-blocking): each call builds a brand-new GoogleProvider -> google.genai/httpx client + connection pool that is never explicitly closed. For the 7 run_agent_sync sites the asyncio.run loop teardown reclaims the connections, but the one awaited path (routes/documents.py::_extend_via_agent, concept_scan) leaves an unclosed AsyncClient on the live request loop for GC to reap. These are low-frequency ops so real-world impact is small and the correctness win clearly justifies losing connection reuse — just flagging that under load this can accumulate unclosed clients / socket-cleanup churn. No change required to merge.

backend/main.py
```diff
@@ -217,10 +217,17 @@ def gemini_test(request: Request):
...
     from agents.health import health_probe_agent
     try:
+        # Fresh per-request client (#354): run_agent_sync's asyncio.run loop
+        # can't reuse the shared provider's pooled connection safely. The health
+        # agent has no _DEFAULTS task, so pin its model name directly.
         result = run_agent_sync(
-            health_probe_agent.run('Reply with exactly the text: Gemini OK')
+            health_probe_agent.run(
+                'Reply with exactly the text: Gemini OK',
+                model=fresh_google_model("gemini-2.5-flash-lite"),
```
Darkest-Teddy: Nit (DRY, non-blocking): the model name "gemini-2.5-flash-lite" is hardcoded here and again in agents/health.py where health_probe_agent is constructed. They must stay in lockstep by hand; a shared constant (or a fresh_model_for-style helper for the probe) would prevent drift if the probe model ever changes.

Darkest-Teddy:
Follow-up on my two review nits:

**Nit #2 (DRY) — fixed in 08a2430.** Hoisted the probe model name into a single `HEALTH_PROBE_MODEL` constant in `agents/_providers.py`, now referenced by both `agents/health.py` (agent default) and `main.py` (fresh-client override). No behavior change — resolved model name is identical, and the sweep/auth/concept-scan tests stay green.

**Nit #1 (client hygiene) — retracted, leaving as-is.** On closer inspection I overstated this. I claimed the concept_scan path leaves an unclosed client on the *live request loop*, but `documents._extend_via_agent` is itself driven through `run_agent_sync` (`_extend_concepts` → `run_agent_sync(_extend_via_agent(...))`), so **all 8 sweep sites run on throwaway `asyncio.run` loops** — the loop teardown reclaims the sockets, and there's no cross-loop or fd leak. An explicit-close fix would mean reaching into google-genai internals (the client exposes only a sync `close()`, no async `aclose()`) plus `try/finally` boilerplate at every site — fragile coupling for ~zero real benefit. Not worth it; the per-request fresh client is the correct, minimal fix for #354.

Both items resolved. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #345, opened 2026-07-15
fix(agents): bound note_chat orchestrator + remaining worker agents with usage limits

Darkest-Teddy:
## Description
Closes out the residual from the #327 audit that PR #243 left open: the `note_chat` orchestrator and three single-shot workers ran without `usage_limits`, defaulting to Pydantic AI library maximums (request_limit=50, no token cap).

## Changes Made
- `routes/notes.py::note_chat` — runs under `ORCHESTRATOR_LIMITS`. Guardrail trips (`UsageLimitExceeded` / `UnexpectedModelBehavior`) degrade to an in-band `200 {reply, degraded: true}` instead of an uncaught 500. This path has no legacy fallback by design (ADR 0017), so in-band degrade mirrors the established `calendar_service._degraded_result` pattern rather than inventing a fallback.
- `routes/notes.py::summarize` / `extract-concepts` — run under `WORKER_LIMITS` via a shared `_run_note_worker` helper that converts guardrail trips to a 503 with a user-facing detail (adding limits without handling would have created the exact uncaught-500 path the issue warns about).
- `services/calendar_service.py::_extract_via_agent` — `syllabus_extraction_agent.run` now passes `WORKER_LIMITS`, matching the same agent''s other run-site in `agents/document.py`. Its caller already degrades gracefully on guardrail trips, so no handling change was needed.
- Tests pin the `usage_limits` kwarg identity at all four run-sites plus the new degrade/503 behavior (8 new tests, written TDD red→green).

## Related Issues
Closes #329

## Testing
- [x] Tested locally
- [x] Added/updated tests

Full backend suite: **959 passed**. The only non-passes are the documented pre-existing baseline failure (`test_save_to_db`, live-Supabase) and a pre-existing Windows-only flake in `test_flashcard_import_service.py::TestRateLimit::test_sixth_call_returns_retry_after` (untouched by this PR; passes on Linux).

## Notes for Reviewers
- `WORKER_LIMITS` / `ORCHESTRATOR_LIMITS` values are unchanged from #243 (`agents/__init__.py`).
- The `degraded: true` key on the chat response is additive; the frontend only reads `reply`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015ori1maMbbFjkpCS7jPgjj

## Summary by CodeRabbit

* **Reliability**
[...]

PR #359, opened 2026-07-20
fix(frontend): make DEPLOY_ENV the single source of truth for env config

Darkest-Teddy:
## Why

Login on **staging.saplinglearn.com** bounced to `/?error=session_expired` with nothing in the browser console.

Root cause (verified by probing staging): the worker serving `staging.saplinglearn.com` was running with **production** config — `GET https://staging.saplinglearn.com/dashboard` (no cookie) returned `307 → https://api.saplinglearn.com/api/auth/google` (the *prod* backend, not `api.staging.saplinglearn.com`). So the `[env.staging]` overrides weren't in effect.

Consequence: staging's login round-tripped through the prod backend, which signs the handoff token with prod's `SESSION_SECRET` and redirects to prod's `FRONTEND_URL`. The `sapling_session` cookie that came back was prod-signed and scoped to `.saplinglearn.com` (a parent-domain cookie, so it also reaches `staging.saplinglearn.com`). Staging's middleware then verified it with staging's `SESSION_SECRET` → HMAC mismatch → `middleware.ts:56` → `session_expired`. The staging backend was healthy throughout.

This is the mirror of the footgun `deployGuard.ts` was built for (#190). The guard's check that catches a consistent-but-wrong-target build only arms when `DEPLOY_ENV` is set — and it wasn't set on either Workers Build, so it slept.

## What changed

`DEPLOY_ENV` becomes the single knob driving every environment-specific value, with the explicit vars kept as backward-compatible fallbacks.

- **`deployGuard.ts`** — `resolveFrontendEnv` (derive `apiUrl`/`cookieDomain` from `FRONTEND_ENVS` when `DEPLOY_ENV` is set; else fall back to explicit vars), plus `expectedEnvForHost` / `detectHostConfigMismatch`. Pure + unit-tested.
- **`middleware.ts`** — derive `API_URL` via the resolver; on a protected route, flag a host/backend mismatch with a loud server log + a distinct **`env_misconfig`** code instead of the misleading `session_expired`.
- **`app/api/auth/session/route.ts`** — cookie `Domain` derived from the resolver (a staging build can't mint a `.saplinglearn.com` cookie).
- **`next.config.ts`** — build-time `BACKEND_URL`/`NEXT_PUBLIC_API_URL`/`COOKIE_DOMAIN` derived from `DEPLOY_ENV`.
- **`wrangler.toml`** — `DEPLOY_ENV` set for `[vars]` (production) and `[env.staging.vars]` (staging).
- **`SignInModal.tsx`** — user copy for `env_misconfig`.
- **ADR `0020`** — root cause + required deploy follow-up.

## Verification

- `tsc --noEmit` → exit 0.
- `eslint` on changed files → 0 errors (one pre-existing `` warning).
- New pure-function logic verified via `tsx` (all assertions pass). `vitest` currently crashes at startup on local Node 20.12.1 (`rolldown` → `util.styleText` array form needs Node ≥ 20.16/22) — pre-existing and unrelated; CI on a supported Node will run the added specs.

## ⚠️ Required deploy follow-up (this PR does NOT fix the running deployment)

Only a staging redeploy clears the live error:

1. Set `DEPLOY_ENV` as a **build** variable on each Workers Build — `staging` / `production`.
2. Staging deploy command must be `npx wrangler deploy --env staging`.
3. Confirm `staging.saplinglearn.com` routes to the **`frontend-staging`** worker, not `frontend`.
4. Verify: `curl -sSI https://staging.saplinglearn.com/dashboard` should show `Location: https://api.staging.saplinglearn.com/api/auth/google`.

## Summary by CodeRabbit

* **New Features**
  * Course content indexing now uses content-addressed chunk IDs and avoids duplicate chunks within indexing runs.
  * Deployment environment configuration now drives API routing and session cookie domain behavior more consistently.

* **Bug Fixes**
  * Added runtime detection for frontend/back-end configuration mismatches to prevent staging/production session issues.
  * When misconfigured, users are redirected to sign in with clearer “environment misconfiguration” messaging.

* **Chores**
  * Added a one-time migration tool to safely deduplicate existing course chunk records, with a preview mode before applying changes.

* **Tests**
  * Expanded coverage for deduplication planning and deploy environment derivation/mismatch detection.

### DIFF
--- a/backend/scripts/dedupe_course_chunks.py
+++ b/backend/scripts/dedupe_course_chunks.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+"""
+One-time dedupe: migrate course_chunks document rows to content-hash ids.
+
+Chunk ids used to be sha256(doc_id::index::text), so the same lecture
+slides uploaded by N students stored N copies of every chunk.
+services/rag_service.py::chunk_id now scopes ids to
+sha256(course_id::text); this script rewrites existing rows to that
+scheme and collapses duplicates. Catalog rows (category=catalog) use
+their own id scheme and are untouched.
+
+For each group of rows sharing chunk_id(course_id, chunk_text), the
+winner is the first row that already has an embedding (avoids
+re-embedding); its metadata (doc_id/uploader_id/chunk_index) carries
+over, matching the last-writer-wins contract in rag_service.
+
+Dry-run by default. Run from backend/:
+    python scripts/dedupe_course_chunks.py            # preview
+    python scripts/dedupe_course_chunks.py --apply    # rewrite rows
+
+Reads .env; for staging run under `dotenv -f .env.staging run -- ...`.
+"""
+import argparse
+import sys
+from pathlib import Path
+
+from dotenv import load_dotenv
+
+BASE = Path(__file__).parent.parent
+load_dotenv(BASE / ".env")
+sys.path.insert(0, str(BASE))
+
+from db.connection import table  # noqa: E402
+from services.rag_service import chunk_id  # noqa: E402
+
+PAGE_SIZE = 1000
+UPSERT_BATCH = 200
+DELETE_BATCH = 50
+
+COLUMNS = (
+    "id,course_id,doc_id,uploader_id,chunk_index,chunk_text,chunk_hash,"
+    "embedding,category,semester,section_id,school"
+)
+
+
+def fetch_document_rows() -> list[dict]:
+    rows: list[dict] = []
+    offset = 0
+    while True:
+        page, total = table("course_chunks").select_with_count(
+            COLUMNS,
+            filters={"category": "eq.document"},
+            order="id.asc",
+            limit=PAGE_SIZE,
+            offset=offset,
+        )
+        rows.extend(page)
+        offset += len(page)
+        if not page or offset >= total:
+            return rows
+
+
+def plan_migration(rows: list[dict]) -> tuple[list[dict], list[str]]:
+    """Group rows by content-hash id; return (records to upsert, ids to delete)."""
+    groups: dict[str, list[dict]] = {}
+    for row in rows:
+        groups.setdefault(chunk_id(row["course_id"], row["chunk_text"]), []).append(row)
+
+    upserts: list[dict] = []
+    deletes: list[str] = []
+    for new_id, members in groups.items():
+        winner = next((m for m in members if m.get("embedding")), members[0])
+        if winner["id"] != new_id:
+            upserts.append({**winner, "id": new_id, "chunk_hash": new_id})
+        deletes.extend(m["id"] for m in members if m["id"] != new_id)
+    return upserts, deletes
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Collapse duplicate course_chunks rows onto content-hash ids."
+    )
+    parser.add_argument("--apply", action="store_true", help="Write changes (default: preview).")
+    args = parser.parse_args()
+
+    rows = fetch_document_rows()
+    upserts, deletes = plan_migration(rows)
+    unique = len(rows) - len(deletes)
+    print(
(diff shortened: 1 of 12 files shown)


AndresL230:
**Staging deploy is down for all branches — traces back to the build-command change from this work.** @Darkest-Teddy 👋

`frontend-staging` has failed **~16 builds in a row since 2026-07-20**, across every branch and all of us, with:

```
ERROR Could not find compiled Open Next config, did you run the build command?
```

From the Workers Builds history, the build command flipped right around this PR's timeframe — last green 07-20 03:03 on `fix/staging-deploy-env` (`f6cd3cf`), worker `modified_on` 03:21, first red 23:39:

| | Last green (`f6cd3cf`) | Every build since |
|---|---|---|
| **Build command** | `npm run cf:build` ✅ | `npx wrangler deploy --env staging` ❌ |
| Deploy command | `npx wrangler versions upload` | `npx wrangler versions upload` (unchanged) |

The new build command skips `opennextjs-cloudflare build`, so nothing produces `.open-next/` and the deploy can't find the compiled config. Looks like the build-command field got overwritten with the `--env staging` deploy line while wiring the DEPLOY_ENV single-source-of-truth config.

**Fix — Cloudflare dashboard → `frontend-staging` → Settings → Builds:**
- Build command: `npx wrangler deploy --env staging` → back to **`npm run cf:build`**
- Leave the deploy command as `npx wrangler versions upload` (it was never the problem).

One field revert and staging goes green for every branch again. 🙏

(Also flagged on #345, where the red check first surfaced.)

Darkest-Teddy:
Fixed — thanks for the precise diagnosis @AndresL230, you nailed it. 🙏

**Root cause confirmed:** the `frontend-staging` Workers Build's **Build command** field had been overwritten from `npm run cf:build` → `npx wrangler deploy --env staging`. A deploy command in the build slot skips `opennextjs-cloudflare build`, so `.open-next/` was never produced → `Could not find compiled Open Next config`. Verified locally that `npm run cf:build` produces `.open-next/worker.js` (the `main` wrangler deploys) and the bare `wrangler deploy` line does not.

**Fix applied (dashboard):**
- Build command: `npx wrangler deploy --env staging` → reverted to **`npm run cf:build`**
- Deploy command: left as `npx wrangler versions upload` (untouched — it was never the problem)

**Verified green:**
```
$ curl -sSI https://staging.saplinglearn.com/dashboard
HTTP/1.1 307 Temporary Redirect
Location: https://api.staging.saplinglearn.com/api/auth/google
```
Routing to the **staging** backend now — so the build outage *and* the original `session_expired` mixup this PR targeted are both resolved. Prod `frontend` build checked too; its Build command is clean.

**Prevention (pushed here, `f5e3b6f`):** the follow-up in ADR 0020 told operators to set a `wrangler deploy --env staging` line + `DEPLOY_ENV` build var but never said to *keep the Build command as `npm run cf:build`* — which is how it ended up in the wrong field. Reworked the ADR + `wrangler.toml` header to spell out the two distinct Build/Deploy fields and explicitly forbid a deploy command in the build slot.

AndresL230:
Containment analysis (2026-07-29): this PR's frontend/docs half is byte-identical to #409 (its own cherry-picked successor — verified via git cherry + interdiff: b662f0e/10b81c6 ≡ f6cd3cf/f5e3b6f), and its backend half (RAG chunk dedup) already landed on main in evolved form via #352 (content-hash ids, later namespaced per-kind) — that part now merge-conflicts with main. Net: after #409 merges, a rebase of this branch leaves an empty diff. Recommend closing this PR in favor of #409 once it lands; leaving that call to the author since the work itself is finished, just duplicated.

AndresL230:
Closing: all of this PR's content is now on main — the frontend/docs half merged verbatim via #409 (see containment analysis above), and the RAG dedup half landed in evolved form via #352. A rebase of this branch against main leaves an empty diff. Nothing here was dropped.

Issue #113, opened 2026-06-08
[EPIC] Frontend UI audit (10/20) — tracking


AndresL230:
Tracking issue for the frontend UI audit. Full report: `docs/frontend-ui-audit.md`.

## Health score — **10 / 20 (Acceptable — significant work needed)**

| Dimension | Score | Key finding |
|---|:---:|---|
| Accessibility | 2/4 | Shared-token contrast failures + `outline:none` strips focus from most inputs |
| Performance | 3/4 | Strong bundle discipline; weak hand-rolled runtime animation; no `next/image` |
| Theming | 2/4 | Brand green `#1B6C42` isn't a token (hard-coded ~90×); 8 competing greens |
| Responsive | 2/4 | JS-only breakpoints at 768px; Admin/Gradebook/Settings desktop-only; touch < 44px |
| Anti-Patterns | 1/4 | Landing/onboarding break **4 of the project's own hard bans** |

**Verdict:** the signed-in app shell is largely on-brand; the **pre-auth landing + onboarding** surface is where the work is — much of it is *deletion* (glass, gradient text, dark-mode orphan), not redesign.

## Issues (priority order)

> **Checklist trued up 2026-07-21.** The boxes below had drifted badly out of sync with the issue tracker — seven children were already closed but still showed unchecked. Corrected against actual issue state; only **#109, #110, #111** remain open.

**P0**
- [x] #102 — Remove the Liquid Glass (glassmorphism) system
- [x] #103 — Remove gradient-text headings on the landing page
- [x] #104 — Consolidate the brand greens into design tokens

**P1**
- [x] #105 — Remove orphaned dark-mode stylesheet + Settings toggle
- [x] #106 — Replace landing hero-metric glass cards + neon-blue accents
- [x] #107 — Fix design-token contrast failures (WCAG AA)
- [x] #108 — Restore visible focus + add a11y labels/alt text
- [ ] #109 — Make Admin / Gradebook / Settings responsive + adopt shared Dialog
- [ ] #110 — Fix sub-44px touch targets + SSR desktop FOUC

**P2**
- [ ] #111 — Optimize runtime animations + add lazy/next-image loading
- [x] #112 — Polish: side-stripe borders, white panels, overshoot easing, Inter font

### Remaining work — where it actually stands

[...]

PR #367, opened 2026-07-21
feat(ui): responsive Admin/Gradebook/Settings, 44px targets, SSR FOUC fix (#109, #110, #113)

### LINKED ISSUE #109: [P1] Make Admin / Gradebook / Settings responsive + adopt shared Dialog

### LINKED ISSUE #110: [P1] Fix sub-44px touch targets + SSR desktop FOUC

Jose-Gael-Cruz-Lopez:
Closes #109. Closes #110. Advances #111. Drives the remaining work on epic #113.

**Note on the epic's state:** #113's checklist had drifted badly — seven of its ten children were already closed but still showed unchecked, which made the epic look far less complete than it was. I've trued the issue body up separately. With #109 and #110 closed here, **#111 is the only child left open**.

## #109 — Admin / Gradebook / Settings responsive + adopt shared Dialog

Inline styles can't express a media query, which is *why* these surfaces were never made responsive. So the fixed multi-column shells moved into real CSS classes in `globals.css`:

- **Admin** — three master/detail panes → `.pane-split` / `.pane-split--wide`, one column below 900px; the metric row → `.metric-grid`, 4→2 columns. Their content track also moved from `1fr` to `minmax(0, 1fr)`: a bare `1fr` floors at min-content, which is what let wide tables push the whole grid past the viewport instead of scrolling inside their own column.
- **Settings** — both `180px 1fr` rows → `.settings-field-row`, stacking below 600px.
- **All four Gradebook modals migrated to the shared `Dialog`** — `AssignmentModal`, `EditWeightsModal`, `SyllabusUploadFlow`, `LetterScaleEditor`. None skipped. They had fixed `minWidth` of 420/520/460/360 that overflowed a 360px phone; they now inherit `min(Npx, 100vw − 32px)` plus the focus trap, Escape handling and scroll lock they never had. `AssignmentModal` additionally had **no** max-height, so an expanded bell-curve section ran its footer off-screen with nothing to scroll.

One `Dialog` change was genuinely required: it focuses the first focusable node, and its close button precedes `{children}` in the DOM — so a form dialog would open with focus on *Close*. Added an opt-in `initialFocusRef` (`autoFocus` can't do this job; React fires it before Dialog's focus pass). Additive — the existing `FlashcardImportModal` consumer is unaffected.

## #110 — touch targets + SSR desktop FOUC

- **SideNav** — rows `minHeight: 44`, collapse chevron 24→44×44, expand bar 28→44. Consequence worth knowing: 44px rows make the full rail ~870px tall, so it scrolls on shorter laptops (the `<aside>` already has `overflow-y: auto`).
- **FOUC, fixed on both axes.** `useIsMobile` is rewritten on `useSyncExternalStore` with `getServerSnapshot → false`, so server and hydrating render still agree (no React 19 mismatch) while every consumer of a breakpoint shares one `MediaQueryList`. But **the hook alone cannot fix the pre-hydration paint** — the phone has already been served desktop HTML. The actual fix is a real `@media (max-width: 767px)` rule hiding the SideNav rail via a new `data-app-sidenav` hook, because CSS reaches the first frame and JS can't.

## #111 — the safe subset only

Landing hero canvas RAF now gated behind `prefers-reduced-motion` (repaints on resize, re-arms if the pref flips); the floating-card RAF hoists its NodeList and `dataset` parsing out of the loop; spotlight rects cached on `mouseenter` and invalidated on scroll/resize; scroll handler rAF-throttled with passive listeners. `Avatar` / `AvatarFrame` got `loading="lazy"` + intrinsic dimensions.

Two items in the audit turned out to be **stale**, and are worth correcting in the record rather than silently skipping:
- `ctx.shadowBlur` does not exist anywhere in `(public)/page.tsx`. Nothing to remove.
- `.landing-spotlight-card` is currently **dead code** — no element in `src/**/*.tsx` carries that class, so the effect is a no-op today. Fixed anyway (cheap, and its CSS still ships), but it buys nothing until the class is reattached.

The `/sapling-icon.svg` logos in TopNav/SideNav were deliberately **left alone**: they're local and above the fold, so lazy-loading them would delay the logo and hurt LCP — the opposite of the issue's intent.

## Verification

| Gate | Result |
|---|---|
| `npm test` | **101 passing** (baseline 88) — 14 files |
| `npx tsc --noEmit` | 0 errors |
| `npx eslint . --max-warnings=-1` | 0 errors, 36 warnings (**exactly** baseline) |
| `npm run build` | succeeds (verified with `BACKEND_URL` set, outside the worktree) |

New tests: `useIsMobile.test.tsx` (7 cases, **mutation-checked** — swapping in the naive `useState(matchMedia(...).matches)` fails 3 of them; one case pins the query string to `MOBILE_BREAKPOINT − 1` so the CSS and JS breakpoints can't silently drift). Plus `modals.dialog.test.tsx` smoke-covering all four migrated modals — they had **zero** tests, and the migration is invisible to tsc: a modal that stops opening, loses its Cancel handler, or drops its accessible name still typechecks.

## Left as follow-ups (deliberately)

- `KnowledgeGraph2D.tsx` — the d3-tick / tooltip / edge-lookup refactor. Untouched; too risky to bundle here.
- `AtmosphericBackdrop.tsx` — offscreen-canvas orb pre-bake. Untouched.
- `next/image` adoption stays blocked: no image loader is configured for the OpenNext/Cloudflare Workers deploy. All 36 baseline eslint warnings are `@next/next/no-img-element` and will stay until that's decided separately.

These three are why **#111 stays open** after this merges.


Jose-Gael-Cruz-Lopez:
@coderabbitai review

PR #368, opened 2026-07-21
feat(dashboard): group courses by semester with an Archive for past terms (#140)

### LINKED ISSUE #140: [P2] Semesters: dashboard/course-list grouping + Archive for past terms
Part of the **Semesters** epic. **Depends on** the backend API issue.

## Goal
Group enrolled classes by semester; default to the current term, with past terms behind an Archive.

## Changes (Dashboard / course-list components, `lib/api.ts`)
- Group the user's enrolled courses **by semester**; show current-term classes by default.
- Add an **Archive** / past-semesters affordance to browse prior terms' classes (read-only entry into that term's gradebook view).
- **Leave the calendar untouched** — assignments stay date-driven with no term label; users can scroll across terms freely (per product decision).

## Acceptance
- Current-term classes show by default; past terms appear only under Archive.
- Selecting an archived class opens that semester's gradebook view.

Jose-Gael-Cruz-Lopez:
Closes #140.

Groups enrolled classes by semester, defaults to the current term, and puts past terms behind an Archive.

## The dependency was already satisfied

#140 is filed as "depends on the backend API issue", but that's stale — nothing was blocking:

- `GET /api/semesters` already exists and serves `{id, term, year, label, start_date, end_date, sort_key}`.
- `GET /api/graph/{user_id}/courses` already returns a **`term`** label per course; only the frontend `EnrolledCourse` interface hadn't declared it.

So this is frontend-only.

## It also fixes a live bug

`screens/Gradebook/Landing.tsx` read `(c as any).semester` off the courses payload — but the API sends **`term`**. The lookup therefore always came back empty and fell through to the hardcoded demo constant:

```ts
const SAMPLE_SEMESTERS = ["Spring 2026", "Fall 2025"];
```

**Every signed-in user was being shown those two fake semester chips instead of their own.** The `as any` is exactly what let the field-name mismatch through the type checker. `SAMPLE_SEMESTERS` is now strictly the logged-out preview; a signed-in user with no terms gets their own empty state (a `termsReady` gate keeps that from hanging on the skeleton). Covered by tests.

## New `src/lib/semesters.ts` — one tested source of truth

- `currentTerm(semesters, today)` mirrors `services/academics.py::current_term` **exactly**: the highest-`sort_key` term whose `[start_date, end_date]` contains today, falling back to the highest `sort_key` overall when today lands in a gap between terms. `today` is injectable. Dates compare as `YYYY-MM-DD` strings (lexicographic == chronological), and a `Date` is read in **local** time — `toISOString()` would shift the day either side of midnight.
- `termRankFromLabel("Fall 2025") -> 20253`, mirroring migration 0019's `year * 10 + ordinal`. Only consulted when no real `sort_key` is available; `sort_key` always wins.
- `groupCoursesByTerm`, `courseTermLabels`, `partitionCurrentAndArchive`.

## Degradation — the rule is "never hide a course"

If `/api/semesters` fails or returns nothing, everything collapses to one flat ungrouped list with no Archive control and nothing hidden. Courses with an empty or unparseable `term` are never dropped and never archived. Future-term courses are never archived. Only a course ranking *strictly below* the current term is archived — showing a course in the wrong bucket is recoverable, hiding it isn't.

## UI

The partition is applied to the "My courses" panel (which is also the mobile `courses` tab), the graph legend chips, the `CoursesKey` overlay, and `ManageCoursesModal`'s list. `CourseProgressRow` was extracted rather than copying that markup a third time. Archived rows route to `/gradebook?semester=<label>`, which the gradebook landing reads and preselects — so selecting an archived class lands in that semester's gradebook, per the acceptance criteria.

**The Calendar is untouched**, per the explicit product decision in the issue.

## Verification

| Gate | Result |
|---|---|
| `npm test` | **131 passing** (baseline 88) — 15 files, +43 tests |
| `npx tsc --noEmit` | 0 errors |
| `npx eslint . --max-warnings=-1` | exit 0, 36 warnings (baseline) |
| `npm run build` | exit 0, compiled successfully |

One thing worth flagging because it nearly shipped broken: fixing the `as any` removed the only `no-explicit-any` in `Landing.tsx`, which left a **stale entry in `eslint-suppressions.json`**. ESLint exits **2** on a stale suppression even with zero errors — `main` exits 0, this branch did not — so the lint gate would have failed CI. Pruned in its own commit.

## For the reviewer

1. Two files outside the intended scope got a one-line fixture addition each (`src/lib/data.test.ts`, `src/components/DocumentUploadModal.test.tsx`), forced by making `term` required on `EnrolledCourse`.
2. `Landing` reads `?semester=` from `window.location.search` in a `useState` initializer rather than `useSearchParams()`, which would require a `Suspense` boundary in the route shell this component doesn't own. It never reaches the rendered tree, so there's no hydration mismatch — but adding the boundary instead is a reasonable alternative.
3. The client's current-term rule is a hand-mirror of the Python one. Test fixtures are migration 0019's four terms verbatim so drift surfaces, but there is no automated cross-language check.
4. The knowledge graph still renders nodes from *all* terms — only the course lists and legends are term-scoped. Filtering graph data by term is a separate product call.

### DIFF
--- a/frontend/eslint-suppressions.json
+++ b/frontend/eslint-suppressions.json
@@ -91,11 +91,6 @@
       "count": 2
     }
   },
-  "src/components/screens/Gradebook/Landing.tsx": {
-    "@typescript-eslint/no-explicit-any": {
-      "count": 1
-    }
-  },
   "src/components/screens/Learn.tsx": {
     "@typescript-eslint/no-explicit-any": {
       "count": 2
--- a/frontend/src/components/DocumentUploadModal.test.tsx
+++ b/frontend/src/components/DocumentUploadModal.test.tsx
@@ -48,6 +48,7 @@ const COURSE: EnrolledCourse = {
   nickname: null,
   node_count: 0,
   enrolled_at: "2026-01-01",
+  term: "Spring 2026",
 };
 
 function renderModal() {
--- a/frontend/src/components/ManageCoursesModal.tsx
+++ b/frontend/src/components/ManageCoursesModal.tsx
@@ -14,6 +14,7 @@ import {
   type EnrolledCourse,
   type OnboardingCourse,
 } from "@/lib/api";
+import { groupCoursesByTerm } from "@/lib/semesters";
 
 const COLOR_RE = /^#[0-9a-fA-F]{6}$/;
 
@@ -63,6 +64,9 @@ export function ManageCoursesModal({ open, userId, courses, onClose, onChanged }
   }, [query, open]);
 
   const enrolledIds = React.useMemo(() => new Set(courses.map(c => c.course_id)), [courses]);
+  // No /api/semesters call here — the modal only needs a stable, readable
+  // order, which the label-derived rank already gives it.
+  const termGroups = React.useMemo(() => groupCoursesByTerm(courses), [courses]);
 
   const handleAdd = async (course: OnboardingCourse) => {
     try {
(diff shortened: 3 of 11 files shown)


Jose-Gael-Cruz-Lopez:
@coderabbitai review

PR #360, opened 2026-07-20
feat: semester-scoped learning + Courses & Semesters hub

Darkest-Teddy:
## Summary

Adds a semester switcher (a redesigned **Courses & Semesters** hub) that scopes the learning surfaces to a chosen term, and enforces a no-retake rule. Implements the approved design/plan under `docs/superpowers/`.

**Approach: Path A — read-time filtering, no schema change / no migration.** The knowledge graph stays keyed on the abstract `course_id`; a term filter is applied at the read boundary. Retakes are disallowed, so each course lives in exactly one semester.

### Backend
- `services/academics.py`: `term_id_for_label` (label → term id, mirrors gradebook) + `user_course_ids_for_term` (the courses a user is enrolled in for a term).
- `services/graph_service.py`: `get_graph(semester=)` and `get_recommendations(semester=)` filter nodes/edges/stats (and DB-side `course_id in (...)` for recs) to the term's courses; `add_course` now rejects re-adding a course enrolled in **any** term (no-retake), surfacing the existing term.
- `routes/graph.py`: optional `?semester=<term label>` on `GET /api/graph/{user}` and `/recommendations`. Absent ⇒ all terms (unchanged).

### Frontend
- `lib/api.ts`: `EnrolledCourse.term`; optional `semester` on `getGraph`/`getRecommendations`.
- `lib/useActiveSemester.ts`: persisted active-semester hook (`[value, setter, hydrated]`) + `distinctTerms`/`resolveActiveSemester` helpers.
- Dashboard / Tree / Learn / Quiz / Study scope to the active semester; graph fetches are gated on hydration to avoid an unscoped→scoped double-fetch.
- `ManageCoursesModal` → **Courses & Semesters** hub: term tabs (set the active semester), courses grouped by term, "Already taken · <term>" disabled state, disabled "Personal learning — Coming soon" placeholder.

## Test Plan
- [x] Backend: `python -m pytest tests/ -q` → 981 passed, 1 skipped (1 pre-existing unrelated OCR event-loop error).
- [x] Frontend: `npm run typecheck` clean, `npm run lint` 0 errors, `npm run test` 92 passed.
- [ ] Manual: open **Courses & Semesters**, switch term tabs, confirm Tree/Learn/Study/Quiz + dashboard graph scope to that term.
- [ ] Manual: search a course taken in a previous term in "Add a course" → shows "Already taken · <term>", disabled.

## Known non-blocking follow-ups
[...]

Issue #319, opened 2026-07-05
Clicking on a course node will result in the course name being selected as a course topic automatically


Darkest-Teddy:
<img width="1489" height="367" alt="Image" src="https://github.com/user-attachments/assets/dd211333-5284-4b79-92a5-2b12b68c3283" />

## Summary

Selecting a course node on the homepage will result in the course name being selected as a topic when redirecting to the AI Tutor page.

PR #370, opened 2026-07-21
fix(tree): don't seed course name as quiz topic on course-node click (#319)

Darkest-Teddy:
Follow-up to #369 (the AI-Tutor fix for #319), on a separate branch.

## Context
The Tree detail panel's **"Quick quiz"** button ran `onQuiz` for any node, including subject-root (course) nodes, passing `topic=<course name>`.

Unlike the tutor, the Quiz screen resolves `topic` against the concept list **with subject-roots excluded** (`Quiz.tsx`), and `QuizPanel` only ever uses the resolved `initialConceptId` — it never reads the raw topic string. So the course name never actually became the quiz topic. This is **hygiene, not a user-facing leak**: the param was dead and misleading.

## Change (`Tree.tsx` only)
- Guard `onQuiz` so a course node opens the quiz picker with no seeded topic (mirrors the tutor behavior); concept nodes keep `topic=name`.
- Drop the `course_id` query param — the Quiz screen only reads `topic`/`concept`, so it was dead.

## Verification
- `tsc --noEmit` and `eslint src/components/screens/Tree.tsx`: clean.
- (vitest 4 can't start on the local Node v20.12.1 — pre-existing rolldown `styleText` incompatibility, unrelated.)

## Base
Stacked on `feat/semester-scoped-learning` (PR #360), same base as #369, since the `onQuiz` handler only exists on that branch.

### DIFF
--- a/frontend/src/components/screens/Tree.tsx
+++ b/frontend/src/components/screens/Tree.tsx
@@ -144,8 +144,12 @@ export function Tree() {
   const onLearn = (n: GraphNode) => router.push(
     `/learn?topic=${encodeURIComponent(n.name)}&mode=socratic${n.course_id ? `&course_id=${encodeURIComponent(n.course_id)}` : ""}`,
   );
+  // Subject-root (course) nodes have no single quiz topic — open the picker
+  // rather than seeding the course name (mirrors the tutor fix, #319). The
+  // Quiz screen only reads `topic`/`concept`, so the old `course_id` param was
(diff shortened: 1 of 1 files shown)


Darkest-Teddy:
Merged into #369 (branch `fix/319-course-node-topic`) at fcb15bf so the tutor and quiz course-node fixes ship together. Closing this as its changes now live in #369.

PR #371, opened 2026-07-21
fix(topnav): close open dropdown instantly when another tab is hovered (#320)

### LINKED ISSUE #320: Taskbar menu doesn't immediately close as the cursor hovers outside of the tab
Image

## Summary
When hovering slightly outside a tab in the top navigation view, the tab's subcategories will remain open instead of closing, which can briefly result in two tabs opening at the same time.

Darkest-Teddy:
Closes #320.

## Problem
Hovering slightly outside a top-nav tab (or moving from one tab to another) left the tab's subcategory panel open instead of closing, and could briefly show **two panels at once**.

## Root cause
Each `NavGroupTrigger` owned its own `open` state **and** its own 140ms close-timer. Moving the cursor from tab A to tab B fired B's `onMouseEnter`, which cancelled only *B's* timer — A's 140ms timer kept A's panel open, so both were visible during the overlap and the old one lingered.

## Fix
Lift the open-state into a single `DesktopGroups` owner tracking one `openIndex` for the whole row. Entering any tab replaces `openIndex` synchronously, so the previously-open panel closes the instant the cursor reaches a new tab — at most one panel is ever open. Click-outside / Escape / route-change closing move up to the same owner. The 140ms grace delay is kept but now only applies when the cursor leaves the row entirely (its original anti-flicker purpose for the trigger↔panel handoff).

## Testing
- Added a regression test asserting opening a second group closes the first (`no lingering second panel`).
- `tsc --noEmit`: clean. `eslint`: 0 errors.
- Note: the local `vitest run` could not execute in this environment — vitest 4 + rolldown require Node 22 (per `.nvmrc`) and this machine has Node 20.12; CI runs on Node 22.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop navigation dropdown behavior so only one group panel remains open at a time.
  * Opening a different group now immediately closes the previously open panel.
  * Clicking outside a navigation group, pressing Escape, or changing pages closes open panels.
  * Fixed delayed hover behavior that could incorrectly close a newly opened panel.

### DIFF
--- a/frontend/src/components/TopNav.test.tsx
+++ b/frontend/src/components/TopNav.test.tsx
@@ -166,6 +166,27 @@ describe("TopNav — dropdown behavior", () => {
     expect(screen.queryByText("Tutor")).toBeNull();
   });
 
+  it("closes the previously-open group when another opens (#320)", async () => {
+    // Regression for #320: only one group panel may be open at a time.
+    // Previously each trigger owned its own open-state, so opening a
+    // second group left the first one's panel lingering (two panels at
+    // once). Opening Community must immediately close Learn.
+    render(<TopNav />);
+    const learn = screen.getByRole("button", { name: /^Learn/ });
+    const community = screen.getByRole("button", { name: /^Community/ });
+
+    fireEvent.click(learn);
+    expect(screen.getByText("Tutor")).toBeTruthy();
+
+    fireEvent.click(community);
+    // Learn's panel is gone (no lingering second panel)...
+    expect(screen.queryByText("Tutor")).toBeNull();
+    expect(learn.getAttribute("aria-expanded")).toBe("false");
+    // ...and Community's panel is the only one showing.
+    expect(screen.getByText("Social")).toBeTruthy();
+    expect(community.getAttribute("aria-expanded")).toBe("true");
+  });
+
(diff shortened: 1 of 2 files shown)


AndresL230:
### Code review

Found 1 issue:

1. Click-outside dismiss no longer works in the nav row's dead space: the check moved from each trigger's tight `wrapperRef` to the shared `rowRef`, and `DesktopGroups` renders with `flex: 1` — so the blank strip between the last trigger and the settings cluster now counts as "inside". A keyboard-focused open (which arms no 140ms timer) followed by a click in that strip leaves the panel stuck open until Escape/route change. (bug due to `frontend/src/components/TopNav.tsx` — `if (rowRef.current && !rowRef.current.contains(e.target as Node)) closeNow();` guarding on the stretched row container)

https://github.com/SaplingLearn/Sapling/blob/57c426a381f6bf4d5ffbaf13c76d27b811d75e0c/frontend/src/components/TopNav.tsx#L313-L317

AndresL230:
The click-outside issue from the review above is fixed in 8adc755: dismissal now checks `[data-nav-group]` wrapper containment instead of the stretched row, with a dead-space regression test and a fake-timers test covering the original #320 hover-timer race.

PR #375, opened 2026-07-21
feat(observability): capture LLM token usage + cost (#118)

Darkest-Teddy:
## What & why

Closes #118 — capture LLM **token usage + cost** for every model call and persist to `llm_usage`.

#118 was blocked by two unimplemented P1 prerequisites; to make it actually function (and testable) this PR delivers the full vertical slice:

- **#115 — tables (migration `0032_observability.sql`)**: `events` + `llm_usage`. Deviations from the original issue, both deliberate: **TEXT ids** (not `uuid`) because `users.id` is TEXT in this schema (`user_andres`), and the numbered-migration convention (`0032_*`) per `CLAUDE.md` rather than the issue's older `migration_*.sql` name. No raw-content columns — sensitive data is a 16-hex `content_fp` only.
- **#116 — `services/events_service.py`**: fire-and-forget write path. A bounded `queue.Queue` drained by one daemon worker that batch-inserts through `db/connection.py::table()`. `log_event` / `log_llm_usage` return in microseconds, **never raise into request handling**, and **never touch the DB on the calling thread**. Kill switch `EVENTS_LOGGING_ENABLED`, overflow drop-counter, and a `flush_now()` test hook. Worker is started/stopped from the FastAPI lifespan.
- **#118 — capture + pricing**:
  - `services/llm_pricing.py` — normalizes token fields across Pydantic AI (`input/output`, legacy `request/response`) and Gemini (`prompt/candidates_token_count`) shapes, and computes `cost_usd` from an editable `MODEL_PRICING` map (unpriced model → `NULL` + one-time warning).
  - `agents/usage.py::record_agent_usage(result, feature=, task=, user_id=)` — one line per agent run-site, reads `result.usage()` + the served model, returns the result unchanged, never raises.
  - `gemini_service.call_gemini*` thread a `feature` arg and capture `response.usage_metadata`.

## Coverage

Every Pydantic AI agent run (document, notes, quiz, chat_tutor, study_guide, social, graph, course_summary, flashcard, calendar syllabus, health) and every direct `call_gemini*` now writes one `llm_usage` row. A static **AST guard test** fails CI if a future agent run-site skips `record_agent_usage`.

## Success criteria (#118)

- [x] Every agent run writes one `llm_usage` row (feature/task/model/tokens).
- [x] Every direct Gemini call writes one row from `usage_metadata`.
- [x] `cost_usd` computed for known models; `NULL` + logged warning for unknown.
[...]

Issue #118, opened 2026-06-08
[P2] Observability: capture LLM token usage + cost (agents + Gemini)


AndresL230:
## Summary

Capture LLM **token usage and cost** for every model call and persist it to the `llm_usage` table via `events_service.log_llm_usage(...)`. Covers both LLM seams: the Pydantic AI agents (`backend/agents/*`) and the direct Gemini calls (`backend/services/gemini_service.py`). Includes a small model→price map to compute `cost_usd`.

## Specifications

**Pydantic AI agents** (`backend/agents/*`, model selection in `agents/_providers.py`):
- After each agent `.run()` / `.run_sync()`, read the result's usage (`result.usage()` → `request_tokens` / `response_tokens` / `total_tokens`) and call `log_llm_usage(feature=..., task=..., model=..., usage=..., user_id=..., request_id=...)`.
- Prefer a single wrapper/helper (e.g. in `agents/_providers.py`) so each call site is one line and no agent is missed. `feature` = the calling surface (`quiz`, `chat_tutor`, `document`, `notes`); `task` = the `model_for(task)` slot.

**Direct Gemini calls** (`backend/services/gemini_service.py`):
- In `call_gemini`, `call_gemini_multiturn`, `call_gemini_json`, capture `response.usage_metadata` (`prompt_token_count`, `candidates_token_count`, `total_token_count`) and call `log_llm_usage(provider="gemini", feature=..., model=..., usage=...)`.
- Thread a lightweight `feature` argument (default e.g. `"misc"`) through these helpers so usage is attributable.

**Cost map:**
- Add a small `MODEL_PRICING` table (per-1K input/output tokens) in a single module (e.g. `backend/services/llm_pricing.py`). Compute `cost_usd`; if the model isn't in the map, log usage with `cost_usd = NULL` and emit a one-time warning.

**Normalization:** different SDKs name token fields differently — normalize to `prompt_tokens` / `completion_tokens` / `total_tokens` before persisting.

## Success criteria

- [ ] Every Pydantic AI agent run writes one `llm_usage` row with correct `feature`, `task`, `model`, token counts.
- [ ] Every direct Gemini call (`call_gemini*`) writes one `llm_usage` row with correct token counts from `usage_metadata`.
- [ ] `cost_usd` is computed from `MODEL_PRICING` for known models; unknown models persist with `cost_usd = NULL` and a logged warning.
[...]

PR #420, opened 2026-07-26
feat(ocr): transcribe text-layer-less pages with Gemini vision

Darkest-Teddy:
Makes scanned and handwritten coursework actually usable. Companion to #419, which stops such uploads producing a *fabricated* summary; this one stops them being unreadable in the first place.

## The bug

A 13-page handwritten linear algebra practice final extracted to **0 characters**. Its eight siblings in the same folder extract 3.5k–7k each.

Docling's OCR is supposed to cover this. It crashes:

```
Stage preprocess failed for run 1, pages [13]: std::bad_alloc
```

The error is swallowed, so the pages come back empty. Docling *does* flag them in `fallback_pages` — but the only consumer of that signal was gated behind `OCR_ENGINE=auto` + `GOT_OCR_ENABLED`, and the default engine is `docling`. So the signal was computed and thrown away on every real upload.

The empty text then reached the classify prompt as an empty `Content:` block, and the model invented a document — a summary about the 1964 Berkeley Free Speech Movement, with CNNs/RNNs/Transformers as the extracted "concepts", bound for a linear algebra course's shared knowledge graph.

## The fix

A Gemini-vision backend that transcribes a rendered page image, wired to the existing `fallback_pages` signal and **deliberately not gated on `OCR_ENGINE=auto`** — that gate is exactly why the rescue never fired.

Chosen over the alternatives for handwritten mathematics specifically: Tesseract is poor at handwriting, and GOT-OCR needs a ~2GB weight download and is impractical CPU-only. Gemini already backs every other AI path in the app and returns LaTeX.

## Verified end to end

Same document, `OCR_ENGINE` left at its default — **0 chars → 4,507 chars**:

```
Problem 1
A. $A \sim \begin{bmatrix} 0 & 1 & 0 & -5 \ 0 & 0 & 1 & 1 \end{bmatrix}$
C. $$\frac{1}{\cancel{\lambda}}(2-\lambda)\lambda^2((3-\lambda)\lambda-2)$$
   $$=(\lambda-2)(\lambda^2-3\lambda+2) = \boxed{(\lambda-2)^2(\lambda-1)}$$
```

Row reductions, characteristic polynomials, `\cancel`, `\boxed`.

## Cost

Off by default (`GEMINI_VISION_OCR_ENABLED`) — one LLM call per **flagged** page. Pages with a normal text layer are never flagged, so a text PDF costs nothing; a 13-page scan costs 13 calls.

## Failure semantics

Mirrors `_apply_got_ocr_fallback`:
- **Per-page error** (quota, network) → keep whatever Docling produced for that page. A partial document beats none.
- **Unavailability** (disabled, no key) → abort the loop, rather than burning a failed call for every page of a 200-page scan.

## Two smaller things included

- **Cache key**: extracted to `_ocr_cache_key` and now includes the new flag. Without this, enabling vision OCR would keep serving the empty string cached from before it existed.
- **Corrected a false comment.** `extraction_service` claimed *"OCR is content-addressed and deterministic."* It no longer is, and that matters for content-addressed chunk ids (ADR 0019): two students uploading the same scanned handout only dedup to one embedding if they transcribe identically. Persisting OCR output content-addressed — rather than merely *caching* it, since `services/cache.py` is off by default and degrades to a clean miss — is the real fix, and is **not** attempted here.

## Tests

13 new tests, each watched failing first. Backend: enablement gating, transcription, `[BLANK PAGE]` → `""`, empty response, error propagation. Service: fires under the **default** engine, skipped when disabled, **zero calls when no pages are flagged**, one page failing doesn't lose the document, unavailability stops after one call, cache key changes with the flag.

`1077 passed`, ruff clean. Two pre-existing issues unrelated and confirmed present on clean `main`: a timing flake in `test_flashcard_import_service` (`assert 61 <= 60`) and a suite-ordering event-loop error in `test_ocr_pipeline::test_save_to_db` (passes in isolation).

## Follow-ups not in scope

- **The `std::bad_alloc` in Docling's OCR preprocessing** is a latent bug that will recur on large scans. Vision now covers for it; the crash itself is unfixed.
- **Persisting OCR output content-addressed**, per the determinism note above — needed before scanned material can dedup across students.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Gemini Vision OCR for PDF pages where standard text extraction is unavailable.
  * Supports transcription of printed and handwritten content, including LaTeX formatting.
  * Continues processing remaining pages when individual vision OCR requests fail.
  * Recognizes genuinely blank pages and avoids adding spurious text.

* **Bug Fixes**
  * Updated OCR caching to distinguish results created with Gemini Vision enabled or disabled.


AndresL230:
### Code review

Found 3 issues:

1. `_ocr_cache_key` omits `GEMINI_VISION_OCR_MODEL`, so changing the vision model keeps serving the old model's transcription for up to the 30-day TTL. The docstring one line above claims it "Includes every flag that changes the output" — but `_model()` (`gemini_vision_backend.py`, fed straight into `generate_content(model=...)`) is the one new output-changing knob this PR added, and it is not in the key. `_engine()`, `_got_ocr_enabled()` and `_gemini_vision_enabled()` are. Reachable whenever `REDIS_URL` is provisioned.

https://github.com/SaplingLearn/Sapling/blob/fe1fce6293a84254d44af8cc9ae51b93d1825ea2/backend/services/extraction_service.py#L222-L233

2. The `.env.example` comment says a scanned upload "extracts to `""` and is rejected" — it is not rejected on this PR's base. On `origin/main`, `_extract_text_or_422` and the async SSE branch in `routes/documents.py` catch only *exceptions*; an empty string is not one, so it flows into the classify prompt and the model fabricates a summary and concepts. That rejection is added by #419, which is still open. As written the comment tells an operator the failure mode is a clean 422 when it is actually silent knowledge-graph corruption — the exact bug this feature exists to prevent.

https://github.com/SaplingLearn/Sapling/blob/fe1fce6293a84254d44af8cc9ae51b93d1825ea2/backend/.env.example#L41-L43

3. The nondeterminism caveat cites "ADR 0019" for content-addressed chunk ids, but `docs/decisions/0019-sapling-model-mode-functionmodel-test-seam.md` is about the `SAPLING_MODEL_MODE` test seam — no ADR covers chunk ids. The behavioral claim is also not true on `main`: `rag_service.py` hashes `f"{doc_id}::{i}::{chunk_text}"` with a fresh `uuid4()` `doc_id` per upload, so two students uploading the same scan never collapse into one embedding regardless of how they transcribe. That only becomes true under #352 (open), which re-keys on `f"{course_code}::{chunk_text}"`. Worth marking as forward-looking and fixing the reference.

https://github.com/SaplingLearn/Sapling/blob/fe1fce6293a84254d44af8cc9ae51b93d1825ea2/backend/services/extraction_service.py#L213-L219

AndresL230:
Follow-ups for the three findings above, plus the four that fell below my reporting threshold, are in #424 — stacked onto this branch so you can take them as a discrete diff rather than having your branch rewritten.

Headline: the vision call now runs as a Pydantic AI agent through `model_for("ocr_vision")`. Beyond the CLAUDE.md convention, a raw `genai.Client` call gets no token/cost attribution from Logfire's `instrument_pydantic_ai()` — so vision OCR, plausibly the largest per-document LLM spend in the app, would have been the one call invisible to the cost dashboard #375 is building for #118.

That refactor also surfaced a latent bug: `_extract_text_or_422` is sync but called from both `async def` handlers, so `asyncio.run` raises there and the per-page `except Exception: continue` would have swallowed it — vision OCR silently doing nothing on the main upload path.

Full suite green (1099 passed); the single `test_ocr_pipeline` error reproduces identically on unmodified `fe1fce6`.

Issue #329, opened 2026-07-09
Residual #327: note_chat orchestrator + note/calendar worker agents run without usage limits


AndresL230:
Follow-up from the review of **PR #243** (`fix: restore mastery updates, persist graph_update_json, wire usage limits`). PR #243 closes the `ORCHESTRATOR_LIMITS`-dead-code finding (#327 / audit `#14`) for the **chat tutor** and **quiz** paths, but its premise that "`notes.py` … does not exist in this repo" is wrong — `routes/notes.py` exists on `main`, and several tool-using / worker agent run-sites were left unbounded. Umbrella: #127.

## Primary — `note_chat` tool-using orchestrator runs unbounded (no ceiling, no fallback)

`POST /api/notes/{note_id}/chat` runs `note_chat_agent`, which is a genuinely tool-using, **graph-mutating** orchestrator:

```python
# backend/agents/note_chat.py:47-51
tools=[
    Tool(read_active_note_tool, name="read_active_note", takes_ctx=True),
    search_course_materials_tool,
    apply_graph_update_tool,   # mutates the knowledge graph
],
```

Its run-site passes **no `usage_limits`**:

```python
# backend/routes/notes.py:264
result = await note_chat_agent.run(body.message, deps=deps)
```

So after PR #243 merges, this is the one tool-using path still running at pydantic-ai library defaults (`request_limit=50`, no token/tool-call cap) — the same class of exposure #327 was meant to close.

**Why this isn't a quick copy-paste of the #243 fix:** the chat/quiz paths degrade gracefully on `UsageLimitExceeded` by falling back to a legacy Gemini path (`_legacy_chat` / `_legacy_generate_quiz`). `note_chat` has **no legacy fallback**, so simply adding `usage_limits=ORCHESTRATOR_LIMITS` would convert an over-budget run into an uncaught 500. This needs a small design decision first:
- [ ] Wire `usage_limits=ORCHESTRATOR_LIMITS` into `note_chat_agent.run` (`routes/notes.py:264`), **and**
- [ ] Add an `UsageLimitExceeded` / `UnexpectedModelBehavior` handler that returns a graceful degraded response (e.g. a plain "couldn't complete that" message, or a non-tool one-shot), matching the chat/quiz fallback ergonomics.

## Secondary — single-shot worker agents missing `WORKER_LIMITS` (lower severity)

[...]

Issue #182, opened 2026-06-12
/api/extract/pdf and /api/extract/image are unauthenticated, unbounded OCR endpoints with no rate limit


Jose-Gael-Cruz-Lopez:
**Severity:** Medium
**Labels:** security, backend, P2
**Location:** `backend/routes/extract.py:45-92` (`extract_pdf`), `:95-121` (`extract_image`)

### Description
Neither handler takes a `Request` or any auth guard (the file has zero `require_self`/`get_session_user_id`). They accept uploads up to 200 pages and run Docling/tesseract OCR — CPU-heavy — for any anonymous caller, with no rate limiting (the only limiter in the app is an in-memory per-user one on three flashcard endpoints). There's also no byte-size cap here (unlike `/documents`). An attacker can drive unbounded OCR load/cost.

### Steps to reproduce
1. `POST /api/extract/pdf` with a large PDF, no auth, repeatedly.
2. Observe unbounded server OCR work with no throttling.

### Expected vs actual
- **Expected:** authenticated + rate-limited + size-bounded.
- **Actual:** anonymous, unbounded, unthrottled.

### Suggested fix
Require an authenticated session on both routes; add a request-size cap and a shared rate limit across OCR/LLM endpoints.

### Acceptance criteria
- Both routes return 401 unauthenticated; oversized uploads return 413; a rate limit is enforced.

Issue #119, opened 2026-06-08
[P2] Observability: activate Logfire for ops/error/LLM tracing


AndresL230:
## Summary

Activate **Logfire** for ops / error / LLM-call tracing. The SDK is already wired in `backend/main.py` (`logfire.configure(...)`, `instrument_fastapi`, `instrument_pydantic_ai`) with scrubbing in `backend/services/logfire_scrubber.py`; it currently sends nothing because no token is set (`send_to_logfire="if-token-present"`). This issue turns it on safely and documents it. Logfire complements the owned Supabase analytics — it covers request traces, latency, errors, and live per-call LLM token usage, which we do **not** rebuild in-house.

## Specifications

- Add `LOGFIRE_TOKEN` to `.env.example` with a short comment (where to get it, that it's optional, and that omitting it keeps Logfire dormant).
- Surface `LOGFIRE_TOKEN` through `backend/config.py` if config centralizes env access.
- Verify the existing scrubbing in `logfire_scrubber.py` still redacts prompts/completions/document text before egress; add a test or assertion if a gap is found. **No student content may leave the app un-fingerprinted.**
- Document the setup in the repo docs (a short section: what Logfire gives you, how to enable, what is scrubbed). Confirm `service_name="sapling-backend"` and the send mode are correct.
- Sanity-check that enabling a token does not double-count or interfere with the Supabase `llm_usage` capture (they are independent).

## Success criteria

- [ ] With `LOGFIRE_TOKEN` set, FastAPI request traces and Pydantic AI agent spans (incl. token usage) appear in the Logfire UI.
- [ ] With no token, the app behaves exactly as today (no egress, no errors).
- [ ] Scrubbing verified: prompts, completions, and uploaded document text are redacted/fingerprinted before leaving the process (test or documented manual check).
- [ ] `.env.example` and docs explain enabling Logfire and what it captures vs. what the Supabase tables capture.

## Dependencies

Independent of the other issues (no code dependency on the events tables). Can ship in parallel.

## Out of scope

The owned Supabase event/usage tables and the admin API.

AndresL230:
**Dependency graph** — independent; can ship in parallel.

PR #424, opened 2026-07-28
fix(ocr): review follow-ups for #420 — agent seam, cost ceiling, cache key, sequential rescuers

AndresL230:
Follow-ups from the review of #420. Targets `worktree-gemini-vision-ocr` so you can take them as a discrete diff rather than having your branch rewritten under you.

Two commits, disjoint file sets.

## 1. Route vision transcription through a Pydantic AI agent

The backend built a raw `genai.Client` and called `generate_content` directly. Three reasons that's wrong here, and the third is the one that matters:

- CLAUDE.md: *"New LLM-driven code should be written as Pydantic AI agents in `backend/agents/`."*
- ADR-0008 made `agents/_providers.py::model_for(task)` the single place a model is chosen. `GEMINI_VISION_OCR_MODEL` was a competing knob that bypassed it — the slot is now `SAPLING_MODEL_OCR_VISION`, like every other agent.
- **Cost attribution.** ADR-0008 notes usage/cost comes free because Logfire's `instrument_pydantic_ai()` tags every pydantic-ai span with tokens and USD. Logfire went live on main last week (#119/#406) and #375 is building usage capture for #118. A raw client call is invisible to both. Vision OCR is one metered call per scanned page — plausibly the largest per-document LLM spend in the app — so it would have been the single call the new cost dashboard couldn't see.

The run is bounded by `WORKER_LIMITS`, matching the #329/#345 sweep: an unbounded run inside a per-page loop multiplies one runaway page across the whole document.

**Latent bug this surfaced.** `_extract_text_or_422` is sync but called from both `async def` handlers (`routes/documents.py:640`, `:771`), so a bare `asyncio.run` raises there — and `_apply_gemini_vision_fallback`'s per-page `except Exception: continue` would have swallowed it. Vision OCR would have silently done nothing on the main upload path. `_run_from_anywhere` hands the coroutine to a worker thread when a loop is already running, copying the context so `agent.override` and the active span survive. Pinned by a test.

Module contract unchanged: same function name and signature, same `GeminiVisionUnavailableError` semantics, `GEMINI_VISION_OCR_ENABLED` still the switch.

## 2. Cache key, cost ceiling, sequential rescuers, accurate docs

- **Cache key omitted the model.** `_ocr_cache_key` claimed to include "every flag that changes the output" but not the vision model, so switching models kept serving the old transcription for the full 30-day TTL. Model and page cap now in the key, mixed in only when vision is enabled so vision-off deployments keep their entries. `GOT_OCR_MODEL_PATH` has the same pre-existing gap — the docstring now names it rather than overclaiming.
- **No cost ceiling.** Nothing upstream bounds the page count: `routes/extract.py` allows `min(max_pages, 50)`, the upload path has no rate limit at all, and #182's 10 req/60s was sized when a request meant one bounded *local* OCR run. `GEMINI_VISION_OCR_MAX_PAGES` caps it per document (default 10) and logs what it skipped.
- **`if/elif` made the rescuers mutually exclusive.** Enabling both meant vision never ran — including on pages GOT-OCR failed to fill, recreating the exact "signal computed then dropped" bug this feature exists to fix. They now run in sequence, GOT-OCR first (local, free), then vision over what it couldn't fill. GOT-OCR's gate is byte-for-byte unchanged.
- **Three false claims.** `.env.example` said an unreadable scan "is rejected" — not on this base; the upload paths convert only extraction *exceptions* to 422, so `""` reaches the classify prompt and the model fabricates. That rejection is #419, still open. The module docstring said vision works on "any engine"; it needs Docling to have run and succeeded. And the cache comment cited ADR 0019 (actually the `SAPLING_MODEL_MODE` test seam) for content-addressed chunk ids, whose dedup claim is untrue on main and becomes true only under #352.

## Verification

`1099 passed, 23 skipped` on the full backend suite; `ruff check` clean. The one error (`test_ocr_pipeline.py::test_save_to_db`, event-loop teardown in a live-network test) is **pre-existing** — I reproduced it identically on unmodified `fe1fce6`.

## Worth your call

- **Page cap default of 10** truncates your motivating document (the 13-page handwritten exam gets 10 pages transcribed, 3 left as Docling had them, with a warning). 20 matches `routes/documents.py`'s `max_pages` if you'd rather nothing truncates in the default window — at the cost of a 2x worse ceiling on `/api/extract`.
- **Empty model response → `""`** is now implemented by catching `UnexpectedModelBehavior`, since `output_type=str` makes pydantic-ai retry once then raise. Same document-level behaviour as before, but it does swallow that exception class. Quota and network errors still propagate.
- **The prompt moved to `system_prompt`** with the image as the user message — a different wire shape than `contents=[image, prompt]`. Worth one live-model spot check before enabling in prod.


AndresL230:
## Locked-deps verification

Ran the branch against a throwaway venv built from `requirements.lock` — the same hash-pinned, non-OCR set CI installs (`pydantic-ai-slim 1.107.0`, `google-genai 2.9.0`, `fastapi 0.138.0`), not the local dev venv, which is 18 minor versions behind on pydantic-ai and a full major behind on google-genai.

**The pydantic-ai API surface held.** `BinaryContent`, `GoogleModelSettings` timeout conversion, `usage_limits`, and the `UnexpectedModelBehavior`-on-empty-output path all behave the same at 1.107.0 as at 1.89.1. That was the risk I expected to find and did not.

**What it did surface:** `tests/test_extraction_service.py` was on CI's `--ignore` list, so nothing in it has ever gated a PR — not the OCR engine gating, not the content-addressed cache key (#97), and not #420's own `TestGeminiVisionFallback` / `TestOcrCacheKey`. It was swept in with three neighbours that genuinely need excluded deps (transformers / docling / live network); it stubs every backend it exercises and needs none of them. Dropped from the ignore list in 283fceb.

Verified with CI's exact command and env on the locked set: `1069 passed, 23 skipped, 0 failed`.

The 4 `test_extraction_backends.py` failures under locked deps are `ModuleNotFoundError: No module named 'transformers'`, reproduce identically on unmodified `fe1fce6`, and stay excluded — legitimately.

### Still unverified
No real Gemini vision call has been made anywhere. The prompt reshape (`system_prompt` + image-only user message vs the old `contents=[image, prompt]`) is untested against a live model, and there's no scanned-PDF fixture in the repo to exercise the path end to end — `tests/fixtures/sample_syllabus.pdf` is a text PDF that never populates `fallback_pages`.

AndresL230:
## First real Gemini call found a regression I introduced

Built a scanned-PDF fixture by rasterizing `tests/fixtures/sample_syllabus.pdf` — image-only, 0-char text layer, so it actually populates `fallback_pages` — while keeping the original's 231 chars of text as known ground truth.

Then made one real vision call. The facts came back perfect: every assignment, due date and type matched. But the shape did not.

| wire shape | output |
|---|---|
| prompt as `system_prompt` (this PR, before 9252497) | **743 chars** — `\documentclass{article}`, five `\usepackage` lines, `\begin{document}`, tabular, `\end{document}` |
| prompt in the user turn (original raw-client shape) | **358 chars** — clean Markdown table |
| prompt in the user turn (after 9252497) | **359 chars** — clean Markdown table |

As a system prompt, *"Use LaTeX for mathematics"* reads as a document-format directive rather than an instruction about math notation, so the model emits an entire LaTeX file. I introduced this in the agent refactor; the original raw-client implementation did not have it.

**Why the preamble matters:** `extracted_text` feeds the classify, summary and concept prompts and is chunked into `course_chunks` for RAG. Ship it and `amsmath`, `booktabs` and `graphicx` become candidate concepts on a graph shared by every student in the course — the same pollution this feature exists to prevent, arriving through a different door.

Fixed in 9252497 by restoring `[image, prompt]` in the user turn. The agent seam, the ADR-0008 model slot and the cost attribution are untouched — only placement changed. The test now pins placement and asserts the instruction is absent from any system prompt; I verified it's revert-proof by reintroducing `system_prompt` and watching it fail.

62 passed on locked deps, ruff clean.

This is the concrete argument for keeping a scanned-PDF fixture in the repo — the entire unit suite passed against this bug, because a FunctionModel will happily return whatever you tell it to.

AndresL230:
## Live stack results — including a correction

Ran the real thing: local Supabase (main's `scripts/local-up.sh`), backend on :5000, session via `POST /api/auth/test-login` as the seeded `rich-user-active`, real `POST /api/documents/upload/sync`.

### Confirmed working
- **The async seam.** A real Gemini vision call from inside a running event loop returns 358 chars of clean Markdown. This is the exact path `_run_from_anywhere` exists for — `asyncio.run` would have raised here and the per-page `except Exception: continue` would have swallowed it. Real model, real loop, no mocks.
- **No LaTeX preamble** in that output, confirming 9252497 holds against the live model.
- **The upload path is unbroken** on this branch: the scanned PDF classified as `syllabus` with an accurate summary and four correct concepts (Lab 1: Loops, Lab 2: Recursion, Midterm Project, Final Exam).

### Correction: that upload did NOT exercise vision OCR
Docling ships **RapidOCR** and read the rasterized page itself — 341 chars, `fallback_pages = []`. The correct summary and concepts came from Docling, not Gemini. My fixture was too easy.

Worth knowing precisely when this feature actually spends money. `docling_backend.py:104` flags a page when:

```python
if char_count < LOW_CHAR_THRESHOLD or math_flag:   # threshold = 40
```

So vision fires only when RapidOCR **also** fails (<40 chars) **or** the page has math-shaped content that isn't LaTeX. That matches #419's evidence, where all 13 pages of a handwritten linear-algebra exam were flagged — RapidOCR handles clean rasterized text fine and fails on handwriting. `docling>=2.15,<3` is in `requirements.txt`, so RapidOCR is present in prod too.

Consequence: a plain scanned text PDF costs nothing, which is the intended design — but it also means the fixture needed to regression-test this path must contain handwriting or unLaTeXed math, not just a missing text layer.

### Still open
No end-to-end run has pushed a document through the *vision* branch of the pipeline. Both halves are proven independently (real model + real loop; Docling flagging + rescue sequencing under unit test), but not composed. That needs a genuinely hard scan as a fixture.

AndresL230:
## The live test found a second, worse bug

Ran the composed path for real. It works — and in doing so it exposed something no mocked test could reach.

Calling the vision seam four times in one process:

```
call 1: OK  302 chars
call 2: RuntimeError: Event loop is closed
call 3: OK  308 chars
call 4: RuntimeError: Event loop is closed
```

`_providers._provider` is a module-level `GoogleProvider`; its async httpx client binds to the first loop `asyncio.run` creates and dies with it. **This is #354** — every `run_agent_sync` caller shares it (`study_guide`, `course_context_service`, `flashcard_import_service`), and the sweep is still open in **#358**, not on main.

Transcription is the only caller that runs in a **loop**, which turns a latent bug into an unusable feature: a 10-page scan alternates success/failure page by page, and `_apply_gemini_vision_fallback`'s per-page `except Exception: continue` keeps Docling's text silently. Half a document degrades back to the mangled OCR this feature exists to replace, with nothing logged.

Fixed in 6a6c555 with a per-run provider passed as a `model=` override — scoped to this agent, leaves the shared `_provider` alone, so it can't conflict with whatever #358 lands. Four consecutive live calls now pass.

## The composed path, verified

Same code, same fixture, only the flag differs:

| | vision OFF | vision ON |
|---|---|---|
| chars | 230 | 302 |
| problem 3 (√) | **dropped** — `<!-- formula-not-decoded -->` | **recovered** — `$\sqrt{x^2 + 16} \leq 5$` |
| `∫x^2` | `∫x~2 dxfrom 0 to 3` | `$\int x^2 dx$` |
| `∂f/∂x` | `af/ax` | `$\partial f/\partial x$` |

Docling losing an entire problem is the point — content it drops never reaches the concept prompt, so it never reaches the student's graph.

A full upload through the async handler classified it `assignment`, summarized all four problems including the dropped one, and produced concepts with real LaTeX and no `\documentclass` wrapper.

## About the fixture

A missing text layer is **not** enough to reach vision. Docling ships RapidOCR and reads rasterized prose perfectly — my first fixture (a rasterized syllabus) came back clean with `fallback_pages = []`, so vision never ran. The committed fixture is reached because `_detect_math_without_latex` flags math-shaped content carrying no LaTeX, which is the actual scanned/handwritten-math case. Regenerate with `tests/fixtures/make_scanned_math_ps4.py`.

Tests are in the **`live_llm`** lane, not `tests/integration/` — they need Docling and a real model, not Postgres, and that lane's conftest mandates a running Supabase stack. Opt-in via `RUN_LIVE_OCR=1` plus a real key, so CI's dummy key is a clean skip. One test guards the premise and fails loudly if Docling stops flagging the fixture, since the other two would otherwise pass vacuously.

**Verification:** live lane `3 passed`; full suite on locked deps `1069 passed, 26 skipped, 0 failed`; ruff clean.

PR #428, opened 2026-07-28
test(e2e): Playwright harness and fixtures (#385)

AndresL230:
## Summary

Browser-lane foundation for epic #402: a Playwright harness in `frontend/e2e/` with authed global setup, per-test DB isolation, failure artifacts, and machine-readable retry reporting — plus exactly one smoke spec proving the harness end-to-end. The six user journeys (#386/#387/#392–#395) build on this and are deliberately **not** included.

- **`frontend/playwright.config.ts`** — chromium-only, `workers: 1` (serial to start), `retries: 2` gated on `process.env.CI`, trace/video/screenshot retained on failure, `timezoneId: "America/New_York"` + `locale: "en-US"` pinned so the frozen #383 clock (2026-03-11T12:00:00Z) renders identical local-time copy (greeting, calendar "today") everywhere.
- **`e2e/global-setup.ts`** — health-checks the stack, mints a session for `rich-user-active` via `POST /api/auth/test-login` (#381) through the same-origin Next proxy, and persists it as `storageState` (`e2e/.auth/`, gitignored). The token is stateless HMAC, so per-test truncation can't invalidate it.
- **`e2e/support/db.ts`** — the single DB seam; journeys never hand-roll SQL. `TRUNCATE … RESTART IDENTITY CASCADE` over all mutable public tables with the #397 denylist (annotated KEEP IN SYNC with `backend/tests/integration/conftest.py::_TRUNCATE_DENYLIST`), then re-seed by running the canonical `db/seed_local_rich.py` through the backend venv. Loopback-exact host guard that raises (never skips) before any connection opens, mirroring #397.
- **`e2e/support/fixtures.ts`** — specs import `test`/`expect` from here; an automatic fixture resets the DB **before** each test (reset-before, so a crashed run can't leak state forward).
- **`e2e/smoke.spec.ts`** — one harness proof: storageState session → middleware-gated `/dashboard` renders `data-testid="app-shell"`. Not a journey.
- **`package.json`** — `build:test` now bakes `NEXT_PUBLIC_TEST_MODE=1` (the composition #383 built the flag for; it was missing), new `e2e` script, devDeps `@playwright/test` / `pg` / `@types/pg` (lockfile regenerated with npm 10.9.2, `lockfileVersion: 3` check passes).
[...]

Issue #402, opened 2026-07-22
[EPIC] Chapter 1 — E2E Regression Suite


AndresL230:
## Problem

The audit (verified against `6765f49`, 2026-07-22) found **970 backend tests, all against per-module `MagicMock` DB stubs; zero browser tests; and CI that never starts a database or applies a migration**. The persistence layer is therefore structurally untested — a misspelled column, a NOT NULL violation, a wrong `.eq()` chain, a broken upsert conflict target, or an encrypt-on-write / decrypt-on-read mismatch would all pass today, because the mock agrees with whatever the caller asserts. Negative-authz tests have the same problem: RLS does not exist (zero `ROW LEVEL SECURITY` statements in `backend/db/migrations/`; the backend connects with the service-role key via `db/connection.py:10`, bypassing it), all access control lives in `services/auth_guard.py`, and every IDOR test runs against mocks — so a guard that is correct in mock-land but wrong about PostgREST filter semantics passes. Meanwhile the local Supabase stack is running and used by no test (open PR #377 starts closing that gap for the subcutaneous lane).

## What this chapter is

The lane split is by *how each suite is run and trusted*, not by subject matter. Chapter 1 asserts conformance to a spec we wrote, finds regressions, requires determinism, runs on every PR, gates merge, tolerates zero flake, and outputs pass/fail. Two lanes:

- **subcutaneous** — HTTP → FastAPI → real Postgres; assertions read back with raw SQL (#396, #397, #398)
- **browser** — Playwright → Next → FastAPI → Postgres (#385–#395)

plus a **plumbing** lane that unblocks both (#378–#384).

## Success criteria

- ≥6 browser journeys
- Flake <1% over 20 consecutive CI runs
- E2E lane <8 min wall clock
- 100% of encrypted columns covered by a raw-SQL round-trip assertion
- Migration replay from empty gated on every PR

## Non-goals

- Replacing the existing mocked suite (it is fast and it stays)
- Adopting RLS policies
- Visual regression / screenshot diffing
- Automating real Google OAuth
- LLM-as-judge scoring as a PR gate

## Dependency graph

Regenerated from the per-issue Dependencies sections:

```
#378 ─┬─> #379 ─> #391 ─────> #392, #393 (each also needs #385)
      └─> #396 ─> #397 ─> #398
                     └──────> #400 (Ch2 oracles; also needs #399)

#380 ─> #384 ─┐
[...]

PR #429, opened 2026-07-28
test(e2e): journey — seeded session → dashboard (#386)

### LINKED ISSUE #386: test(e2e): journey — seeded session → dashboard
## Context
First journey: prove the auth fixture + harness end to end with the smallest possible assertion surface, so every later journey builds on a known-good base.

## Scope
- Inject the session cookie → land on dashboard → assert a known seeded course is visible.
- No other assertions — this is the harness smoke test.

## Acceptance criteria
- [ ] 10 consecutive local runs green.

## Dependencies
Depends on: #385
Blocks: #388

---
Chapter: 1 (Regression Suite) | Lane: browser | Size: S | Suggested owner: Andres
Part of epic #402

AndresL230:
## Summary

The smallest browser journey (#386), the known-good base for the rest of the chapter-1 lane: storageState session → `/dashboard` → one assertion that a known seeded course is visible.

- **`frontend/e2e/dashboard.spec.ts`** (new) — opens the "My courses" key via `dashboard-courses-key-toggle` (it starts collapsed on the default sidebar layout), then asserts the seeded course inside the testid-anchored row labels: `getByTestId("dashboard-course-code").filter({ hasText: "MATH210" })`. MATH210 (Linear Algebra, `rich-course-math210` from `db/seed_local_rich.py`) is `rich-user-active`'s spring-2026 enrollment, and spring 2026 is the current term under the frozen test-mode clock (2026-03-11), so its row always sits in the current — not Archive — bucket. Nothing else is asserted — smoke.spec.ts owns the URL/shell proof. Zero `waitForTimeout`; the click's auto-wait doubles as the data-arrived gate.
- **`frontend/e2e/global-setup.ts`** (harness fix, additive) — the minted storageState now also carries the `sapling_user` localStorage entry (mirroring exactly what `setActiveUser` persists on real sign-in). The cookie alone is not a signed-in browser: `UserContext` reads identity from localStorage only, so every authed screen rendered its shell but never fetched user data — this journey found `/dashboard` stuck on its skeleton for the full test timeout. Every sibling journey needs this, so it belongs in the shared setup rather than per-spec boilerplate.

## Testids — `dashboard` surface minted (review follow-up)

Per `docs/frontend-testids.md` "Adding a surface" (review flagged that the doc's testid-anchor rule is blanket, and #428 set the precedent):

- `dashboard-courses-key-toggle` on the key's expand/collapse button; `dashboard-course-code` on each course-row label (repeated, **no suffix** — a documented deviation from the suffix rule: journeys select rows by seeded content, so no per-row identity is exposed).
- Registry table row + `### dashboard` inventory added to `docs/frontend-testids.md`.
- `src/components/screens/Dashboard.tsx` added to the eslint enforcement `files` array. That surfaced **25 pre-existing untagged intrinsic elements**; instead of a blind testid sweep they are absorbed by the repo's documented legacy-debt mechanism — the ESLint bulk-suppressions baseline (`eslint-suppressions.json`, regenerated via `npm run lint:baseline`; see the header comment in `eslint.config.mjs`). The baseline diff is exactly one entry (`no-restricted-syntax: 25` for Dashboard.tsx), so any NEW interactive element in the file still fails CI without a testid. The doc's enforcement section records this (mirroring the existing landing-page note). `npm run lint` and `tsc --noEmit` pass; Dashboard unit tests pass.

## Verification — fresh 10 consecutive local runs of the spec as it will merge, all green

Full cycle under the shared stack lock (`make e2e-up` → 10 × `npx playwright test e2e/dashboard.spec.ts` → `make e2e-down`, exit 0), run **after** the testid rework:

| Run | Result |
| --- | --- |
| 1/10 | 1 passed (6.7s) |
| 2/10 | 1 passed (4.8s) |
| 3/10 | 1 passed (3.8s) |
| 4/10 | 1 passed (4.0s) |
| 5/10 | 1 passed (4.1s) |
| 6/10 | 1 passed (4.4s) |
| 7/10 | 1 passed (4.8s) |
| 8/10 | 1 passed (4.9s) |
| 9/10 | 1 passed (4.7s) |
| 10/10 | 1 passed (4.8s) |

History: cycle 1 failed run 1 with the skeleton hang described above (which surfaced the global-setup gap); cycle 2 (text/role anchors) was 10/10 green; cycle 3 (the tally above) is the reworked, merge-state spec.

## Product observation (reported, not papered over)

A browser holding a **valid `sapling_session` cookie but no `sapling_user` localStorage** renders `/dashboard` as an infinite skeleton: middleware lets the request through, but `UserContext` only bootstraps identity from localStorage (written solely by the sign-in flows) and never falls back to cookie-based `/api/auth/me` — even though that endpoint can identify the user from the cookie alone (the auth callback already calls it without `user_id`). Real-world shape: any state where the HttpOnly cookie outlives localStorage. Worth a follow-up issue; this PR only mirrors the real sign-in state in the harness rather than changing app behavior.

Part of #402, closes #386

### DIFF
--- a/docs/frontend-testids.md
+++ b/docs/frontend-testids.md
@@ -74,6 +74,7 @@ renders the element.
 | Knowledge graph | `graph` | `frontend/src/components/KnowledgeGraph.tsx` (wrapper only — not the 2D/3D internals) |
 | App shell | `app` | `frontend/src/components/ShellFrame.tsx` (the authed layout frame every `(shell)` route renders inside) |
 | Study rooms | `social` | `frontend/src/components/screens/Social.tsx` (rooms sidebar, chat, overview, study match, directory — added with the #394 two-context journey) |
+| Dashboard | `dashboard` | `frontend/src/components/screens/Dashboard.tsx` (rendered by `(shell)/dashboard/page.tsx`) |
 
 Two surfaces do **not** carry their testids in the screen file named by the
 route:
@@ -203,6 +204,13 @@ route:
 | `social-member-kick-{userId}` | "Kick" on that member row (leader only) |
 | `social-match-run` | "Find matches" on the study-match tab |
 
+### `dashboard`
+
+| testid | element |
+| --- | --- |
+| `dashboard-courses-key-toggle` | expand/collapse toggle of the "My courses" key overlay on the graph panel (default sidebar layout; the key starts collapsed) |
+| `dashboard-course-code` | a course row's code/name label inside the expanded key — repeated per course with **no suffix** (deliberate deviation from the suffix rule above: journeys select a row by seeded content, `getByTestId(…).filter({ hasText })`, so no per-row identity is exposed) |
+
 ## Enforcement
 
 `frontend/eslint.config.mjs` has a per-file `no-restricted-syntax` block scoped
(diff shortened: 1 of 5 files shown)


AndresL230:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance. (One convention finding — copy-based anchors — was fixed in-PR: the dashboard testid surface was minted per docs/frontend-testids.md and re-verified with a fresh 10/10 run streak.)

Issue #405, opened 2026-07-22
bug(social): /api/social room routes 500 — rooms table missing topic/course/owner_id/updated_at/is_public columns


AndresL230:
## Summary

`backend/routes/social.py` reads columns from the `rooms` table that **no migration ever creates**, so every `/api/social` endpoint that lists or fetches a room fails with a hard Postgres error. This reproduces on a fresh local Supabase (and would on any DB whose schema comes only from the migrations).

## The drift

`rooms` is defined once, in `backend/db/migrations/0001_baseline_schema.sql:286`:

```sql
CREATE TABLE IF NOT EXISTS rooms (
    id          TEXT PRIMARY KEY,
    name        TEXT NOT NULL,
    invite_code TEXT UNIQUE NOT NULL,
    created_by  TEXT NOT NULL REFERENCES users(id),
    created_at  TIMESTAMPTZ DEFAULT now()
);
```

No later migration alters `rooms` (verified: `grep -ri "alter table rooms" backend/db/migrations` → nothing).

But `social.py` selects a wider column list in **four** places (lines ~44, ~80, ~99, ~268):

```python
table("rooms").select(
    "id,name,topic,course,owner_id,created_by,invite_code,created_at,updated_at,is_public",
    ...
)
```

Five of those columns — **`topic`, `course`, `owner_id`, `updated_at`, `is_public`** — do not exist on the table. Additionally, the room-create path (`create_room`, ~line 30) only ever inserts `id, name, invite_code, created_by`, so even the columns that "should" exist are never populated.

## Reproduction (fresh local Supabase, per docs/local-supabase.md)

```python
from db.connection import table
table("rooms").select("id,name,topic,course,owner_id,created_by,invite_code,created_at,updated_at,is_public", limit=1)
```

→ PostgREST 400:
```json
{"code":"42703","message":"column rooms.topic does not exist"}
```

Affected endpoints (all under `/api/social`): join-by-invite, list-my-rooms, get-room, and room-delete authorization — anything going through those four `select()` calls returns 500 instead of the intended response.

## Why it was hidden until now

The backend test suite is fully mocked (`tests/conftest.py` stubs Supabase), so no test exercises the real column list. This surfaced while building the local Supabase dev stack + integration suite (#372 / the #362-#363 follow-ups): the `/api/social` room routes were deliberately excluded from the new integration E2E because of exactly this drift.

## Fix options

[...]

Issue #382, opened 2026-07-22
feat(frontend): data-testid convention on six core E2E surfaces


AndresL230:
## Context
The repo has exactly two `data-testid` attributes and both live inside `vi.mock` blocks in `src/components/TopNav.test.tsx` (:53, :56) — shippable code has zero (verified repo-wide at `6765f49`). Browser tests would have to anchor on CSS classes and copy text, both of which churn.

## Scope
- Document a `data-testid` naming convention.
- Apply it to six surfaces: sign-in, approval gate, upload modal, tutor input + send, quiz answer, graph container.
- An ESLint rule preventing drift on new components in those surfaces.
- Not a whole-app sweep — six surfaces only.

## Acceptance criteria
- [ ] Convention documented.
- [ ] Six surfaces carry stable testids.
- [ ] ESLint rule prevents drift on new components.

## Dependencies
Depends on: none
Blocks: #385

---
Chapter: 1 (Regression Suite) | Lane: plumbing | Size: M | Suggested owner: Jose
Part of epic #402

Issue #231, opened 2026-06-13
[security] Realtime + Storage anon-key surface bypasses route-code authorization (RLS / Realtime-Auth gap)


Jose-Gael-Cruz-Lopez:
## Summary
The backend enforces ownership entirely in **route code** (`require_self` + `user_id`-scoped queries) using the **service-role key**, so PostgreSQL **RLS is disabled project-wide** (`docs/architecture.md`, `docs/decisions/0017`). But the **frontend talks to Supabase directly with the public anon key** for realtime and storage — paths that never touch route code and therefore have **no authorization**. The `room_id`/message filters used client-side are conveniences, not security boundaries.

Discovered while fixing #124 (realtime chat rendered ciphertext). #124 fixes the *display*; this issue is the *authorization model* behind it.

**Severity: metadata-only for realtime** — `room_messages.text` / `room_reactions` content the attacker would receive is column-encrypted (ciphertext), so message *bodies* stay protected. What leaks is **metadata** (who posted/reacted, when, in which room, reply/edit/delete activity, presence/typing) for **arbitrary rooms the user isn't a member of**. Storage is a separate anon-write / public-read concern.

## Full anon-key surface audit (every route-code-bypassing path)
Client obtained via `getSupabase()` (`NEXT_PUBLIC_SUPABASE_ANON_KEY`).

**Realtime (`frontend/src/components/screens/Social.tsx`):**
1. `postgres_changes` INSERT/UPDATE on **`room_messages`**, filtered client-side by `room_id=eq.<roomId>` — the filter is attacker-controlled; any room's message stream is subscribable.
2. `postgres_changes` INSERT/DELETE on **`room_reactions`** — **no room filter at all** (only a client-side "is this message loaded" check), so reaction events for **all** rooms stream to any subscriber. Broadest exposure.
3. **Presence** channel `presence:<roomId>` (typing indicators) — also unauthenticated.

**Storage (anon-key upload + public getPublicUrl):**
4. `Social.tsx` → `chat-images` bucket (chat attachments).
5. `ReportIssueFlow.tsx` → issue-report screenshots bucket.
6. `Admin.tsx` → cosmetics bucket.

The frontend-audit docs already flag "Verify bucket RLS policy" / "Bucket permissions must allow anon writes." Risk: anon upload abuse + public read of all objects.

**No direct anon-key _table_ queries exist** — all DB reads/writes go through FastAPI (route-code enforced). The gap is bounded to realtime + storage above.

## Membership model
[...]

Issue #124, opened 2026-06-08
[P0] Realtime room chat renders encrypted ciphertext for incoming messages


AndresL230:
**Severity: CRITICAL** · Source: `docs/backend-contract-bug-audit.md` (finding #2)

`frontend/src/components/screens/Social.tsx:142-159,422` · `backend/routes/social.py:335,356`

`room_messages.text` is column-encrypted at rest; only the REST read paths decrypt it. The RoomChat realtime subscription stores `{ ...row }` and renders `payload.new.text` directly, so messages from other users (which arrive only via realtime) display as **unreadable ciphertext** until a manual reload. The UPDATE handler has the same defect for edits.

### Fix
Treat realtime as a "something changed" signal — re-fetch affected messages through the decrypting REST endpoint (or always source text from the API) rather than trusting `payload.new.text`.

_Likely a root cause of #85 (Social page doesn't work)._

PR #431, opened 2026-07-28
test(e2e): journey — study room with two browser contexts (#394)

### LINKED ISSUE #394: test(e2e): journey — study room with two browser contexts
## Context
Study rooms are the only multi-user surface, and `room_messages.text` is column-encrypted. Two browser contexts are the only honest way to observe propagation.

## Scope
- Two browser contexts, two seeded users, one room.
- Assert chat propagates between contexts and both graphs render.

## Acceptance criteria
- [ ] Both contexts assert receipt of the other's message; 10 consecutive local runs green.

## Dependencies
Depends on: #385
Blocks: none

---
Chapter: 1 (Regression Suite) | Lane: browser | Size: M | Suggested owner: Jose
Part of epic #402

AndresL230:
## Summary

Two-context study-room journey: `rich-user-active` and `rich-user-second` sign into separate Playwright browser contexts, meet in the seeded `rich-room-study-group`, and **each context asserts receipt of the other's message** (attributed to the sender by display name) plus its own knowledge-graph render (`graph-container` on `/dashboard` for both users). `room_messages.text` is column-encrypted at rest, so the assertions prove the full write → encrypt → realtime signal → decrypting re-fetch → render loop. Zero `waitForTimeout` — every wait is event-based.

- `frontend/e2e/study-room.spec.ts` — the journey.
- `frontend/e2e/support/session.ts` — NEW additive helper minting a second user's `storageState` via `POST /api/auth/test-login` (same flow as `global-setup.ts`, returned as an object for `browser.newContext`). It carries both the `sapling_session` cookie AND the `sapling_user` localStorage identity `UserContext` requires; `global-setup.ts` takes the `test/386-journey-dashboard` branch's localStorage fix **verbatim** so the sibling PRs converge on identical content.
- `USER_SECOND` appended to `e2e/support/stack.ts`.
- `Social.tsx` joins the #382 `data-testid` convention per the registry's "Adding a surface" process: full `social-*` sweep, inventory appended to `docs/frontend-testids.md`, file appended to the eslint enforcement array.

## Room creation — chosen path

The journey uses the **seeded room** (`db/seed_local_rich.py`: `rich-room-study-group`, both users members, 3 encrypted messages) rather than creating one via UI/API. Most robust: no create/join dance, deterministic ids, and the per-test truncate + re-seed restores it. Rendering the seeded history doubles as proof of the decrypting read path. UI create/join stays reachable via the new `social-create-room` / `social-join-room` testids for later specs.

## Propagation mechanism — findings

Room chat propagates via **Supabase Realtime `postgres_changes`** over a websocket — not polling, not SSE. `Social.tsx` subscribes on `room_messages` filtered by room; a FOREIGN insert event carries **ciphertext** `text` (#124), so the client treats it purely as a "something changed" signal and re-fetches through the decrypting REST endpoint (`GET /api/social/rooms/{id}/messages`). Own sends render optimistically and reconcile from the POST response. There is **no polling fallback** — realtime is the only cross-client path — so the spec waits for each context's server-side subscription confirmation (the `Subscribed to PostgreSQL` frame on the room topic) before any cross-context send: an insert landing before the receiver's subscription is live would never be delivered.

## Bugs found → unblocking migrations

1. **#405 confirmed live on the local schema** (verified at runtime pre-migration, not assumed): `rooms` had only `id/name/invite_code/created_by/created_at`, and the wide select `routes/social.py` uses returned PostgREST `42703 — column rooms.topic does not exist`, so every room-listing endpoint 500'd and the journey was impossible. **Migration 0032** adds the five drift columns (`ADD COLUMN IF NOT EXISTS`, nullable, unpopulated) — schema reconciliation only; `create_room` population semantics stay open in #405.
2. **New finding, same drift family:** the `supabase_realtime` publication existed but published **zero tables** locally (verified: `pg_publication_tables` empty) — nothing in the repo ever added `room_messages`, the deployed environments got it via dashboard configuration (the production behavior #124 fixed proves it's published there). Without it, `postgres_changes` never fire and cross-client chat silently does not propagate at all on a migrations-only stack. **Migration 0033** publishes it, guarded to no-op where the publication is absent or the table already published. `room_reactions` deliberately stays unpublished (#231).

## Verification — 10 consecutive local runs green

Single lock hold: pre-migration evidence capture → `make e2e-up` (applies 0032/0033) → 10 runs → `make e2e-down`.

| run | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| result | pass 11.9s | pass 10.3s | pass 10.7s | pass 10.6s | pass 10.3s | pass 10.7s | pass 9.7s | pass 10.6s | pass 10.6s | pass 9.9s |

Post-migration state checks: `rooms` carries all selected columns; `pg_publication_tables` lists `supabase_realtime → room_messages`. `npm run typecheck`, `eslint` (incl. the new enforcement on `Social.tsx`), and the `Social.realtime.test.ts` vitest suite all pass.

Part of #402, closes #394.

### DIFF
--- a/backend/db/migrations/0032_rooms_missing_columns.sql
+++ b/backend/db/migrations/0032_rooms_missing_columns.sql
@@ -1,0 +1,24 @@
+-- 0032: add the rooms columns routes/social.py already reads (#405).
+--
+-- `rooms` was created in 0001 with only id/name/invite_code/created_by/
+-- created_at, but social.py selects `topic, course, owner_id, updated_at,
+-- is_public` in four places (join-by-invite, list-my-rooms, room overview,
+-- kick authorization). On any schema that replays purely from migrations
+-- (local Supabase, the E2E stack) PostgREST rejects the select with 42703
+-- ("column rooms.topic does not exist") and every one of those endpoints
+-- 500s — which blocks the study-room E2E journey (#394).
+--
+-- This migration only reconciles the schema with the column list the code
+-- reads. The columns stay nullable and unpopulated: deciding their semantics
+-- (owner_id vs created_by, is_public default, and populating them in
+-- create_room) remains open product work tracked in #405.
+--
+-- IF NOT EXISTS keeps this a no-op on any environment where the columns
+-- already exist via historical dashboard DDL drift.
+
+ALTER TABLE rooms
+    ADD COLUMN IF NOT EXISTS topic      TEXT,
+    ADD COLUMN IF NOT EXISTS course     TEXT,
+    ADD COLUMN IF NOT EXISTS owner_id   TEXT,
+    ADD COLUMN IF NOT EXISTS updated_at TIMESTAMPTZ,
+    ADD COLUMN IF NOT EXISTS is_public  BOOLEAN;
--- a/backend/db/migrations/0033_realtime_publish_room_messages.sql
+++ b/backend/db/migrations/0033_realtime_publish_room_messages.sql
@@ -1,0 +1,36 @@
+-- 0033: publish room_messages on the supabase_realtime publication.
+--
+-- Study-room chat propagation (frontend/src/components/screens/Social.tsx)
+-- rides Supabase Realtime `postgres_changes` on `room_messages`: a foreign
+-- INSERT/UPDATE event is treated as a "something changed" signal and the
+-- client re-fetches through the decrypting REST endpoint (#124 — the payload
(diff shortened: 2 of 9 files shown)


AndresL230:
### Code review

Note for maintainers: open PR #375 also adds a migration numbered 0032 — whichever merges second must renumber (see comment there).

Issue #393, opened 2026-07-22
test(e2e): journey — quiz answer → mastery update (UI + DB)


AndresL230:
## Context
Quiz answers drive mastery, which since migration 0023 is an append-only `node_mastery_events` table (never a `mastery_events` column). A correct answer must never lower a score — a property no mocked test can currently falsify.

## Scope
- Answer a quiz through the UI; assert mastery changes in the UI **and** in the DB (`node_mastery_events` rows).
- Assert monotonicity: a correct answer never lowers a score.
- Deterministic quiz generation via #391.

## Acceptance criteria
- [ ] UI and DB agree on the mastery change; 10 consecutive local runs green.
- [ ] Monotonicity assertion in place.

## Dependencies
Depends on: #385, #391
Blocks: none

---
Chapter: 1 (Regression Suite) | Lane: browser | Size: M | Suggested owner: Jack
Part of epic #402

Issue #391, opened 2026-07-22
feat(backend): pydantic-ai test seam via FunctionModel (SAPLING_MODEL_MODE)


AndresL230:
## Context
`agents/_providers.py:86` builds one process-wide provider at import (`_provider = GoogleProvider(api_key=GEMINI_API_KEY or "dummy-key-for-import")`); `model_for()` (`:89-97`) returns `GoogleModel(name, provider=_provider)` with no injection point — so today's tests stub whole agent objects, and prompt/tool wiring is never exercised. *This is Chapter 1, not Chapter 2 — without it the tutor and quiz journeys can't be deterministic.*

## Scope
- Dispatch on `SAPLING_MODEL_MODE`: `real` → `GoogleModel`; `function` → pydantic-ai `FunctionModel` (returns scripted tool calls, so tool registration, arg schemas, and the retry loop all run for real); `cassette` → replay from `tests/evals`.
- No change to production behavior (default remains `real`).

## Acceptance criteria
- [ ] An integration test drives a real agent with `FunctionModel` and asserts on tool call arguments.

## Dependencies
Depends on: #379
Blocks: #392, #393, #399

---
Chapter: 1 (Regression Suite) | Lane: browser | Size: M | Suggested owner: Jack
Part of epic #402

AndresL230:
**Unblocked as of 2026-07-23** — this issue's dependency chain (`#378 → #379 → #391`) is now fully satisfied: both closed by PR #411, merged to `main` @ `5dee71d`.

Two things that landed with #379 change how you should build this seam:

- **`backend/tests/conftest.py` now has an autouse `_hermetic_llm_transport` fixture** that patches `google.genai._api_client.BaseApiClient` at the **class** level (methods: `request` / `request_streamed` / `async_request` / `async_request_streamed`, the `_request*` chokepoints, and the file upload/download paths) so any unstubbed LLM call raises `UnstubbedLLMEgress`. pydantic-ai's `GoogleModel` wraps a `google.genai.Client`, so **the agents lane already rides this same seam** — there is a test asserting exactly that in `backend/tests/test_hermetic_llm_guard.py`.

  Practical consequence: `SAPLING_MODEL_MODE` must swap in `FunctionModel` **above** the transport, not by unpatching the guard. If your seam ends up needing the guard disabled to work, that is a signal the substitution is happening at the wrong layer.

[...]

PR #434, opened 2026-07-28
test(e2e): journey — tutor conversation persists to messages (#392)

### LINKED ISSUE #392: test(e2e): journey — tutor conversation persists to messages
## Context
The tutor is **not** token-streamed client-side: `src/lib/api.ts:114` (`sendChat`) returns a full `{ reply, graph_update, mastery_changes }` object — so no SSE handling is needed in this journey. Turns persist to `messages`, whose `content` column is encrypted at rest.

## Scope
- Send a message through the UI; assert a response renders and the turn persists to `messages` (raw-SQL readback: column is ciphertext, decrypts to the sent text).
- Deterministic model responses via #391 (`SAPLING_MODEL_MODE=function`).

## Acceptance criteria
- [ ] Response renders and the turn is persisted; 10 consecutive local runs green.

## Dependencies
Depends on: #385, #391
Blocks: none

---
Chapter: 1 (Regression Suite) | Lane: browser | Size: M | Suggested owner: Jack
Part of epic #402

AndresL230:
## Summary

Browser journey for the tutor chat (#392): resume the seeded `rich-sess-cs-recursion` session on `/learn`, send a message through the real composer (`tutor-input` → `tutor-send`), assert the deterministic reply renders in `tutor-messages`, then raw-SQL readback of `messages` proving the encryption boundary:

- exactly 2 new rows for the session (seeded 4 → 6; a double-write from an accidental legacy fallback would fail the count),
- `content` at rest is **ciphertext** (≠ the sent plaintext / rendered reply), and
- both rows **decrypt** — via the backend's own `decrypt_if_present`, shelled through the backend venv — to exactly the sent text and the deterministic reply. Both directions are asserted because `decrypt_if_present` echoes plaintext input back (legacy tolerance), so decrypt-equality alone can't prove encryption.

No SSE handling and zero `waitForTimeout`: `sendChat` is a plain fetch returning the full `{ reply, … }` object; assertions wait on the rendered reply locator.

### Boot contract

```
SAPLING_MODEL_MODE=function SAPLING_FUNCTION_HANDLERS=agents.function_handlers_e2e make e2e-up
```

`scripts/e2e-up.sh` does not scrub the environment — exported vars flow through `make` into the `setsid venv/bin/python -m uvicorn` launch, and `backend/.env` doesn't define `SAPLING_*`, so no sanctioned-passing workaround was needed. The second var is new (see below).

## Seam findings (#391 coverage of the tutor path)

1. **`POST /api/learn/chat` IS covered by the seam** — it runs `chat_tutor` (pydantic-ai) via `_chat_via_agent`, whose model comes from `model_for("chat_tutor")`.
2. **…but the browser could never reach the seam before this PR.** `routes/learn.py::_resolve_model_pref` built a live `GoogleModel` via `google_model(name)` for the per-request fast/smart pref, passed as `agent.run(model=…)` — overriding the FunctionModel. The frontend **always** sends a pref (`ModelToggle`'s `useModelPref` defaults to `"fast"`), so in function mode every browser tutor turn would still have dialed real Gemini. Fixed here: in any non-real mode the pref resolves to `None` (the agent default, already mode-correct). `real` mode is byte-for-byte unchanged.
   - **Same bypass exists in `routes/quiz.py::_resolve_model_pref` (lines ~127–145)** — left untouched (out of scope), flagged for the #393 quiz journey.
3. **`POST /api/learn/start-session` is NOT covered** — it still runs the legacy `services/gemini_service.py::call_gemini_multiturn` path (explicit `TODO(refactor-3 follow-up)` in the route). The journey therefore enters the chat by **resuming a seeded session**, never via "Start learning". Same applies to `/action` and `/mode-switch` (both legacy).
4. **Handler registration for out-of-process runs did not exist.** #391's `register_function_handler` is in-process only; ADR 0019 anticipated E2E lanes would "set the env at process start, then register per-task handlers" but shipped no mechanism for a separate uvicorn process. Added additively: `SAPLING_FUNCTION_HANDLERS=<module>` is imported lazily, **once, only on a dispatch miss** — the pytest lane (env unset, explicit registration) is unchanged, a missing handler still raises the pointed `LookupError`, explicit registrations always win, and a typo'd module path fails loudly (ImportError) instead of running handler-less. `agents/function_handlers_e2e.py` holds the fixed-reply `chat_tutor` handler; other journeys should append theirs there.
5. **Residual live-Gemini *attempt* (not a dependency): RAG query embedding.** `_chat_via_agent` calls `rag_service.retrieve_chunks` whenever the session's course has a `course_code` (seeded CS101 does), and `_embed_query` hits `gemini-embedding-001` outside any seam. `retrieve_chunks` swallows every exception and returns `[]`, so the journey's outcome does not depend on it — but each E2E tutor turn still *attempts* one billed embedding call when `backend/.env` carries a real key. Worth a follow-up (seam or env kill-switch for embeddings).

## Verification

- **10 consecutive local runs green** (single lock hold, fresh function-mode stack boot from this branch's worktree): runs 1–10 all `1 passed`, 4.5–5.9s each, zero failed/flaky/retried, clean teardown. Boot line used: the contract above.
- Backend unit suite in the worktree: **1075 passed, 1 skipped** (includes 6 new tests in `tests/test_e2e_function_handlers.py`: env-module autoload end-to-end through the real `socratic_agent`, unset-var LookupError posture, bad-module loud failure, explicit-registration precedence, and `_resolve_model_pref` mode behavior both ways). `ruff check` clean on changed files.
- `tsc --noEmit` and `eslint` clean on changed frontend files (one pre-existing baselined warning in `Learn.tsx`).
- Encrypt→decrypt round-trip of the exact helper snippet verified against the local stack's `ENCRYPTION_KEY` before the runs.

## Testid note (deviation, with precedent)

The journey needs one new testid, `tutor-session-resume-{sessionId}`, on the Learn screen's session rows — but `screens/Learn.tsx` has ~20 intrinsic interactive elements outside any browser journey, so adding it to the eslint `files` array would force testids on all of them. Followed the documented `signin-trigger` exception instead: testid added + inventoried in `docs/frontend-testids.md` with the exception noted, file stays out of the lint block.

The AI-disclosure `DisclaimerModal` (fixed overlay, no testids) is pre-acked via `addInitScript` (models a returning user); dismissing it by copy would violate the no-copy-anchoring rule and tagging it was out of scope.

## Cross-PR convergence notes

- `frontend/e2e/global-setup.ts` gained the `sapling_user` localStorage entry **copied verbatim from `test/386-journey-dashboard`** (per the harness-owner heads-up): the cookie-only storageState never populates `UserContext` (it bootstraps identity solely from that localStorage key), so the Learn screen's session list — which this journey resumes from — stays empty without it. Identical content on both branches, so the merge is clean whichever lands first.
- At least one sibling journey PR boots with the same `SAPLING_FUNCTION_HANDLERS=agents.function_handlers_e2e` contract and will carry its own version of the handlers module / `_providers.py` autoload. These were authored in isolated worktrees — whoever merges second should converge to one module (append handlers per task) rather than duplicate the mechanism.

## Files

- `frontend/e2e/tutor.spec.ts` — the journey (new)
- `frontend/e2e/support/decrypt.ts` — backend-venv decrypt seam (new)
- `frontend/e2e/support/db.ts` — additive `queryRaw` (parameterized SELECTs behind the existing loopback guard)
- `frontend/src/components/screens/Learn.tsx` — `tutor-session-resume-{sessionId}`
- `docs/frontend-testids.md` — inventory + lint-exception note (append-only)
- `backend/agents/_providers.py` — `SAPLING_FUNCTION_HANDLERS` lazy autoload on dispatch miss
- `backend/agents/function_handlers_e2e.py` — deterministic `chat_tutor` handler (new)
- `backend/routes/learn.py` — `_resolve_model_pref` respects `SAPLING_MODEL_MODE`
- `backend/tests/test_e2e_function_handlers.py` — 6 unit tests (new)

Part of #402, closes #392

## Summary by CodeRabbit

* **New Features**
  * Added support for loading function-mode tutor handlers during startup and dispatch.
  * Added deterministic tutor behavior for end-to-end testing.
  * Added a test identifier for resuming recent tutor sessions.

* **Bug Fixes**
  * Prevented live model overrides from being created in non-real modes.
  * Improved missing or invalid handler errors.

* **Tests**
  * Added coverage for handler loading, routing, session persistence, and encrypted message storage.

* **Documentation**
  * Updated tutor end-to-end testing and test-identifier guidance.

### DIFF
--- a/backend/agents/_providers.py
+++ b/backend/agents/_providers.py
@@ -155,8 +155,10 @@ def clear_function_handlers() -> None:
 # process start ... then register per-task handlers". SAPLING_FUNCTION_HANDLERS
 # names a module (e.g. `agents.function_handlers_e2e`) whose import registers
 # handlers. It is consulted lazily, once, and ONLY on a dispatch miss — so the
-# pytest lane (which registers explicitly and never sets the var) is unchanged,
-# and a test that forgot to register a handler still gets the loud LookupError.
+# normal (hermetic) pytest lane, which registers explicitly and does not set
+# the var, is unchanged, and a test that forgot to register a handler still
+# gets the loud LookupError. (The seam's own tests DO set the var, resetting
+# the latch around each case.)
 
 _ENV_HANDLERS_LOADED = False
 
@@ -164,17 +166,20 @@ _ENV_HANDLERS_LOADED = False
 def _load_env_handlers_module() -> None:
     """Import the SAPLING_FUNCTION_HANDLERS module (once per process).
 
-    A bad module path raises ImportError loudly at first dispatch — a broken
-    E2E boot must fail the run, never silently fall through to LookupError
-    with the wrong story.
+    A bad module path raises ImportError loudly at EVERY dispatch, not just
+    the first — a broken E2E boot must fail the run, never silently fall
+    through to LookupError with the wrong story. That is why the latch is
+    set only after a successful import (or when there is nothing to load):
+    latching before the import would permanently cache the failure as
+    "loaded" and downgrade every later dispatch to the generic LookupError.
     """
     global _ENV_HANDLERS_LOADED
     if _ENV_HANDLERS_LOADED:
         return
-    _ENV_HANDLERS_LOADED = True
     module = (os.getenv("SAPLING_FUNCTION_HANDLERS") or "").strip()
     if module:
-        importlib.import_module(module)
+        importlib.import_module(module)  # raises before the latch on failure
+    _ENV_HANDLERS_LOADED = True
 
 
 def _function_model_for(task: AgentTask) -> "Model":
--- a/backend/agents/_providers.py
+++ b/backend/agents/_providers.py
@@ -17,6 +17,7 @@ the Pro tier for the conversational tutor where reasoning depth matters.
 
 from __future__ import annotations
 
+import importlib
 import os
 from typing import TYPE_CHECKING, Callable, Literal
 
@@ -145,6 +146,37 @@ def clear_function_handlers() -> None:
     _FUNCTION_HANDLERS.clear()
 
 
+# ── Boot-time handler registration for out-of-process runs (#392) ──────────
+#
+# `register_function_handler` covers in-process tests, but the E2E browser lane
+# boots the backend as a separate uvicorn process (`SAPLING_MODEL_MODE=function
+# make e2e-up`) where no pytest fixture can reach the registry. ADR 0019
+# anticipated this: full-journey lanes "must set SAPLING_MODEL_MODE=function at
+# process start ... then register per-task handlers". SAPLING_FUNCTION_HANDLERS
+# names a module (e.g. `agents.function_handlers_e2e`) whose import registers
+# handlers. It is consulted lazily, once, and ONLY on a dispatch miss — so the
+# pytest lane (which registers explicitly and never sets the var) is unchanged,
+# and a test that forgot to register a handler still gets the loud LookupError.
+
+_ENV_HANDLERS_LOADED = False
+
+
+def _load_env_handlers_module() -> None:
+    """Import the SAPLING_FUNCTION_HANDLERS module (once per process).
+
+    A bad module path raises ImportError loudly at first dispatch — a broken
+    E2E boot must fail the run, never silently fall through to LookupError
+    with the wrong story.
+    """
+    global _ENV_HANDLERS_LOADED
+    if _ENV_HANDLERS_LOADED:
+        return
+    _ENV_HANDLERS_LOADED = True
+    module = (os.getenv("SAPLING_FUNCTION_HANDLERS") or "").strip()
+    if module:
+        importlib.import_module(module)
+
+
 def _function_model_for(task: AgentTask) -> "Model":
     """Build a FunctionModel that dispatches to the task's registered handler at
     run time. The lookup is deferred to the call (not capture time) so a handler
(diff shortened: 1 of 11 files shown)


AndresL230:
Review fixes pushed (`30161b2`, rebased onto `385a534`/#431 — the verbatim `global-setup.ts` commit dropped as already-upstream):

1. **Latch bug (required): fixed.** `_ENV_HANDLERS_LOADED` now sets only after a successful import (or when the var is unset), so a bad `SAPLING_FUNCTION_HANDLERS` path raises `ModuleNotFoundError` on **every** dispatch instead of downgrading to the generic `LookupError` after the first. Regression test `test_bad_env_module_path_fails_loudly_on_every_dispatch` dispatches twice within one latch lifetime — verified **red against the old latch order, green on the fix**.
2. **Testid mechanism: converged on the suppressions baseline.** `screens/Learn.tsx` joins the eslint `files` array; `npm run lint:baseline` produced exactly one new entry (`no-restricted-syntax: 21` for Learn.tsx — 22 intrinsic elements minus the newly tagged resume button, no other counts moved). The prose lint exception is gone and the Tutor table row now names the ChatPanel/Learn.tsx split, Sign-in-row style.
3. **Comment accuracy:** the "pytest lane never sets the var / never imports it" claims are qualified to the normal (hermetic) lane, with the seam's own tests called out as the deliberate exception.

Re-verification: backend **1076 passed, 1 skipped** (+1 regression test), `ruff`/`tsc`/`eslint` (0 errors)/`vitest` (204 passed) clean, and a fresh lock-held function-mode confirmation cycle on the rebased branch: **3/3 consecutive `e2e/tutor.spec.ts` runs green** (7.4s/5.9s/5.5s), clean teardown. The original 10/10 tally stands as acceptance evidence.

AndresL230:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance. (Three review findings were fixed in-PR: the env-handlers failure latch now sets only after a successful import — regression-tested red→green with a dispatch-twice test; Learn.tsx joined eslint testid enforcement via the suppressions baseline; comment claims qualified.)

Issue #129, opened 2026-06-08
[P2] Quiz scoring & idempotency: double-submit double-counts, free point on malformed item


AndresL230:
**Severity: MEDIUM + LOW** · Source: `docs/backend-contract-bug-audit.md` (findings #10, #29)

- **#10 [MEDIUM] Quiz submit has no idempotency — resubmission double-counts mastery/streak** — `routes/quiz.py:355-436`. `submit_quiz` sets `completed_at` (`:432`) but never reads it; re-POSTing the same `quiz_id` re-applies the mastery delta, increments `times_studied`, appends another event, re-runs `update_streak`. *Fix:* if `attempt.get("completed_at")` is set, return the stored result (or 409).
- **#29 [LOW] Free point for an unanswered, no-correct-option question** — `routes/quiz.py:372-379`. Both `selected` and `correct_label` default to `''`, so `is_correct = '' == ''` is True for a malformed item, inflating score and mastery delta. *Fix:* `is_correct = bool(correct_opt) and selected == correct_label`.

AndresL230:
Closed by PR #464 (merged). Evidence: resubmitting a completed quiz 409s, and the gate is an ATOMIC conditional update (completed_at is.null) so two concurrent submits can't both apply mastery (the review caught that a plain pre-read was check-then-act); malformed items with no correct option no longer grade a missing answer as correct ('' == ''). Pinned by hermetic tests (409, race-loser, is.null idiom, malformed item) and a promoted journey that replays the captured wire body and DB-asserts exactly one node_mastery_events row (lane 20/20 pre-merge).

PR #437, opened 2026-07-28
test(e2e): journey — quiz answer → mastery update (UI + DB) (#393)

### LINKED ISSUE #393: test(e2e): journey — quiz answer → mastery update (UI + DB)

AndresL230:
## Summary

Browser journey for the quiz → mastery path (#393): sign in as the seeded `rich-user-active` (storageState), deep-link to `/quiz?concept=rich-node-cs-recursion`, answer a deterministic three-question quiz through the real UI, and assert the mastery change **in the UI and in the database**:

- `graph_nodes.mastery_score` moves 0.25 → 0.34 (3 correct × +0.03), `times_studied` bumps 0 → 1, `last_studied_at` set;
- exactly **one** new row appends to `node_mastery_events` (the 0023 table — never a `mastery_events` column) with `delta ≈ +0.09` and reason `Quiz: 3/3 correct`;
- the journey's `quiz_attempts` row completes (`score=3, total=3, completed_at` set; scoped `id NOT LIKE 'rich-%'` because the seed carries a completed baseline attempt on the same node);
- the percentages the student *saw* (`quiz-results-mastery`) equal the DB state read back over the raw-SQL seam — writes go through the app, assertions never ride the layer that wrote.

**Monotonicity** — the property no mocked test can falsify — is pinned three ways: a fully correct submission must never lower the rendered score (`uiAfter >= uiBefore`), the persisted score (`masteryAfter >= masteryBefore`), or the event delta sign (`delta >= 0`). If the scoring math in `routes/quiz.py` ever regresses to subtract on correct answers, this journey fails.

## Product bugs found (2, both fixed here)

1. **Every UI quiz submission has 422'd since the shell revamp.** `QuizPanel` submits `{question_id, selected}`; backend `models.AnswerItem` requires `selected_label`. Introduced by `399eae3` ("frontend: ship revamp shell, screens, and API client"), which dropped the old `selected_label` payload; verified live on `origin/main`. No mocked test on either side could catch it — backend tests construct `AnswerItem` directly, frontend tests mock `fetch`. The journey caught it on its first live run (`POST /api/quiz/submit -> 422`, `.e2e/backend.log`). Fixed by renaming the request field to the backend contract (the response items keep their separate `selected` key). Distinct from the known #129 scoring bug, which this journey does not exercise (single submit per answer, well-formed items).
2. **The quiz half of the #392-discovered seam bypass.** `routes/quiz.py::_resolve_model_pref` built a live `GoogleModel` for a `fast`/`smart` pref regardless of `SAPLING_MODEL_MODE` — same bug the #392 sibling fixed in `routes/learn.py`, deliberately left to this PR. The quiz UI sends no pref today (`generateQuiz` omits `model_pref`, which is why this journey's runs stayed deterministic regardless), but any client that did would silently dial Gemini in function mode. Fixed with the same pattern + the same test shape as learn's.

## Seam findings (#391 / ADR 0019 / #392 convergence)

- **Quiz generation IS covered by the FunctionModel seam** — it does not route through `gemini_service.call_gemini_json`; `/api/quiz/generate` runs `agents/quiz.py::quiz_agent` built via `model_for("quiz")`. Verified with evidence, not assumption: the spec's mode guard asserts the scripted fixture's question text before answering (a real-Gemini stack fails loudly there), generate round-trips in ~530 ms (`.e2e/backend.log`), and the only Gemini egress in a full run is the one below-seam embedding call (next bullet).
- **Env propagation confirmed:** `SAPLING_MODEL_MODE=function SAPLING_FUNCTION_HANDLERS=agents.function_handlers_e2e make e2e-up` reaches uvicorn by plain env inheritance; `load_dotenv` (main.py, config.py) never overrides exported vars. No script changes needed.
- **Handler registration converged on #392's mechanism — now upstream.** After #434 merged (including the review-found latch fix in `_load_env_handlers_module` and its dispatch-twice regression test), this branch was rebased onto it: the shared seam files (`_providers.py`, `learn.py`, `db.ts`, `global-setup.ts`) collapsed into main entirely, and the PR diff is down to the quiz-specific surface — the quiz handler **appended** to `agents/function_handlers_e2e.py` (per that module's own instruction) and the quiz contract tests appended to its test file. The scripted quiz contract (labels B, C, A = `E2E_QUIZ_CORRECT_LABELS`) is pinned hermetically through the real agent + real wire mapping.
- **Below-the-seam egress (follow-up candidate):** `/api/quiz/generate`'s best-effort RAG grounding (`_course_material_block` → `retrieve_chunks` → `_embed_query`) sits below the model seam: with a real key in `backend/.env` each generate makes one live `gemini-embedding-001:batchEmbedContents` call even in function mode (observed in `.e2e/backend.log`). It cannot affect the journey's outcome (the scripted handler ignores prompt content; `course_chunks` is empty post-reset; failures are swallowed by design) but it is incidental egress and a bounded (60 s) latency source the spec budgets for. Worth a follow-up to short-circuit grounding in non-real modes.
- **`quiz_context` deliberately unregistered:** `submit_quiz`'s background context update runs in `try/except pass`; without a handler it fails fast with no post-response DB write racing the next test's truncate + re-seed.

## 10-run tally

**10/10 consecutive local runs green** in one lock-held cycle (fresh `SAPLING_MODEL_MODE=function SAPLING_FUNCTION_HANDLERS=agents.function_handlers_e2e make e2e-up` boot, `make e2e-down` after; cycle exit rc=0):

```
RUN  1/10  1 passed (7.9s)      RUN  6/10  1 passed (6.6s)
RUN  2/10  1 passed (6.2s)      RUN  7/10  1 passed (6.2s)
RUN  3/10  1 passed (6.4s)      RUN  8/10  1 passed (5.2s)
RUN  4/10  1 passed (6.7s)      RUN  9/10  1 passed (6.7s)
RUN  5/10  1 passed (6.6s)      RUN 10/10  1 passed (5.4s)
```

The consecutive greens double as isolation proof: every run asserts the seeded 0.25 baseline before playing, which would fail if the prior run's 0.34 write leaked past the per-test truncate + re-seed.

**Post-rebase confirmation** (after re-syncing onto main's merged #434 seam, incl. the latch fix): fresh lock-held cycle, **3/3 green** — 8.4 s / 6.9 s / 6.8 s.

## Harness changes (additive)

- `frontend/e2e/quiz.spec.ts` — the journey. Zero `waitForTimeout`; locator/event waits only. Pre-acks the AI disclaimer via `addInitScript` (the modal intercepts every click for a browser that never acked; scoped to the spec so `global-setup.ts` stays byte-identical to the #386/#394 content already on main).
- `frontend/src/components/QuizPanel.tsx` + `docs/frontend-testids.md` — one new testid on the existing `quiz` surface, `quiz-results-mastery` (inventory row appended; the file is already in the eslint `files` array), plus the `selected_label` fix above. DB readback uses main's `queryRaw` helper (merged with #434) — no harness-file changes remain in this diff.

Backend hermetic suite: 1080 passed, 23 skipped on the rebased branch. Frontend: tsc clean, eslint 0 errors, vitest 204 passed.

Part of #402, closes #393

### DIFF
--- a/backend/agents/function_handlers_e2e.py
+++ b/backend/agents/function_handlers_e2e.py
@@ -29,7 +29,7 @@ handlers here rather than growing parallel modules.
 
 from __future__ import annotations
 
-from pydantic_ai.messages import ModelResponse, TextPart
+from pydantic_ai.messages import ModelResponse, TextPart, ToolCallPart
 
 from agents._providers import register_function_handler
 
@@ -46,3 +46,53 @@ def _chat_tutor_handler(messages, info) -> ModelResponse:
 
 
 register_function_handler("chat_tutor", _chat_tutor_handler)
+
+
+# ── Quiz (#393) ────────────────────────────────────────────────────────────
+#
+# Fixed three-question quiz. The correct options sit at indexes 1, 2, 0;
+# routes/quiz.py assigns wire labels in options order, so the correct labels
+# are B, C, A. frontend/e2e/quiz.spec.ts clicks exactly that sequence and
+# backend/tests/test_e2e_function_handlers.py pins it — KEEP ALL THREE IN
+# SYNC (E2E_QUIZ_CORRECT_LABELS is that contract, exported like
+# E2E_TUTOR_REPLY above). The count deliberately ignores the requested
+# num_questions: the route stores whatever the agent returns, and a fixed
+# count keeps the journey's mastery math (3 correct × +0.03 = +0.09)
+# byte-stable across runs. The `concept` field is fixed too (the route
+# derives mastery from the DB node it looked up, never from this field).
+#
+# The single ToolCallPart below is the agent's OUTPUT tool (the structured
+# Quiz result) — not a function tool, so the no-tool-calls constraint above
+# holds. `quiz_context` stays deliberately unregistered: submit_quiz updates
+# it in a BackgroundTask wrapped in `except Exception: pass`, so an
+# unscripted handler fails fast with no post-response DB write racing the
+# next test's truncate + re-seed.
+
+E2E_QUIZ_CORRECT_LABELS = ("B", "C", "A")
+
+_E2E_QUIZ_QUESTIONS = [
+    {
+        "question": f"E2E deterministic question {n}: which option is marked correct?",
+        "type": "multiple_choice",
+        "difficulty": "medium",
+        "options": [f"Q{n} option A", f"Q{n} option B", f"Q{n} option C", f"Q{n} option D"],
+        "correct_answer": f"Q{n} option {label}",
+        "explanation": f"Scripted E2E fixture: option {label} is the marked answer for question {n}.",
+        "concept": "Recursion",
+    }
+    for n, label in zip((1, 2, 3), E2E_QUIZ_CORRECT_LABELS)
+]
+
+
+def _quiz_handler(messages, info) -> ModelResponse:
+    return ModelResponse(
+        parts=[
+            ToolCallPart(
+                tool_name=info.output_tools[0].name,
+                args={"questions": _E2E_QUIZ_QUESTIONS},
+            )
+        ]
+    )
+
+
+register_function_handler("quiz", _quiz_handler)
--- a/backend/routes/quiz.py
+++ b/backend/routes/quiz.py
@@ -138,10 +138,19 @@ def _resolve_model_pref(model_pref: str | None):
     """
     if not model_pref:
         return None
+    from agents._providers import _model_mode, google_model
+    if _model_mode() != "real":
+        # #391 seam: the per-request fast/smart override must not bypass
+        # SAPLING_MODEL_MODE by constructing a live GoogleModel. Same fix as
+        # routes/learn.py (#392) — there the browser always sends a pref;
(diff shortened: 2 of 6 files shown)


AndresL230:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance. (Panel verified the #438 fix restores the exact pre-revamp submit contract, the quiz seam handler validates against the real output schema, and the mastery math matches the seeded baseline.)

Issue #436, opened 2026-07-28
Pre-existing flake on main: test_ocr_pipeline::test_save_to_db fails with 'Event loop is closed'


AndresL230:
Reproduces identically on untouched main (verified during PR #432's backend suite run: 1082 passed, 23 skipped, 1 error — this test).

Belongs in the epic #402 flake ledger: a deterministic-lane test erroring on main undermines the 'mocked suite is trustworthy' baseline. Likely an async teardown ordering issue in the OCR pipeline test.

Discovered-by: #387 / PR #432 verification.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #447, opened 2026-07-28
fix(graph): dedupe subject-root synthesis per distinct course (#355)

### LINKED ISSUE #355: Duplicate subject_root node in /api/graph → React duplicate-key warnings in KnowledgeGraph2D

AndresL230:
Fixes #355.

## Root cause

Subject-root synthesis in `graph_service.py` iterates the user's **enrollments**; a user with two offerings of the same abstract course got `subject_root__<course>` synthesized twice — a duplicate node id plus duplicated hub-spoke edges (+1 node / +5 edges against the rich seed, reproduced by the e2e_oracles graph check on every run).

## Fix

- `backend/services/graph_service.py`: guard subject-root synthesis with a `seen_course_ids` set placed before any node/edge construction — one subject root (and one set of spokes) per **distinct abstract course**.
- `backend/tests/test_graph_service.py`: TDD regression — a user with two offerings of the same course now yields unique `nodes[].id`, exactly one `subject_root__c1`, and no surplus spokes (RED before the fix: duplicate ids / 3 == 2).

## Promotion (Chapter 2 → Chapter 1, promotion 1 of 3)

- `frontend/e2e/graph.spec.ts`: the #355 acceptance test rode in as `test.fixme` — now un-fixme'd with **no assertions relaxed**; stale "bug is open" comments refreshed. The e2e_oracles graph check must go clean after this lands.

## Verification

- Hermetic suite: 1155 passed, 26 skipped, 1 error = the pre-existing #436 flake (`test_ocr_pipeline::test_save_to_db`, untouched by this PR).
- `ruff check` clean; frontend `tsc --noEmit` clean.
- Playwright graph lane + oracles run against the live local stack before merge (verification pass on this branch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate subject hub nodes and connections in graph views when users have multiple offerings or enrollments for the same course.
  * Graphs now display a single subject hub per abstract course, with accurate connections to concept nodes.

* **Tests**
  * Added coverage confirming correct node and connection counts in this scenario.
  * Updated end-to-end checks to validate the resolved behavior.

### DIFF
--- a/backend/services/graph_service.py
+++ b/backend/services/graph_service.py
@@ -227,12 +227,21 @@ def get_graph(user_id: str) -> dict:
         if cid and cid in course_color_map:
             n["course_color"] = course_color_map[cid]
 
-    # Build subject root hubs from enrolled courses
+    # Build subject root hubs from enrolled courses — one root per DISTINCT
+    # abstract course, not per enrollment. A user can hold two offerings of the
+    # same abstract course (e.g. re-taking it, or two sections); the graph stays
+    # keyed on the abstract course_id, so without this dedup the synthesized
+    # subject_root node (and its hub-spoke edges) would be duplicated once per
+    # extra enrollment (#355).
     subject_nodes = []
     subject_edges = []
+    seen_course_ids: set[str] = set()
 
     for enrollment in enrolled_courses:
         course_id = enrollment["course_id"]
+        if course_id in seen_course_ids:
+            continue
+        seen_course_ids.add(course_id)
         course = enrollment.get("courses", {}) if isinstance(enrollment.get("courses"), dict) else {}
         course_code = course.get("course_code", "")
         course_name = course.get("course_name", "")
--- a/backend/tests/test_graph_service.py
+++ b/backend/tests/test_graph_service.py
@@ -261,6 +261,40 @@ class TestGetGraph:
         assert len(node["mastery_events"]) == 2
         assert result["stats"]["avg_learning_velocity"] > 0
 
+    def test_dedupes_subject_root_for_two_offerings_of_same_course(self):
+        """#355: a user enrolled in TWO offerings of the same abstract course
+        (e.g. re-taking it, or two sections) must get exactly ONE subject-root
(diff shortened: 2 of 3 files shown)


AndresL230:
Scoping note: merging this auto-closes #355. The #355 thread's related get_courses per-enrollment fan-out (duplicate course_id rows on /tree etc.) is deliberately NOT in this PR — deduping that payload interacts with enrollment_id-keyed consumers (gradebook) and needs its own analysis. Tracked in https://github.com/SaplingLearn/Sapling/issues/449.

Issue #151, opened 2026-06-08
[P1] Agent migration: retire call_gemini* + gemini_service.py (final cutover)


AndresL230:
Part of the **Agent migration** epic — the **final cutover** (roadmap milestone: end of July, "gemini_service retired as primary LLM seam"). Owner: Jack.

## Goal
Once all feature surfaces are migrated (the issues above), deprecate and remove the raw Gemini helpers.

## Scope
- Remove `call_gemini`, `call_gemini_json`, `call_gemini_multiturn` (and `MODEL_LITE`/`MODEL_SMART` constants) from `backend/services/gemini_service.py`, or reduce the file to the shared provider only.
- Retire the **legacy chat fallback** last: `learn.py:417,538,847` (`_legacy_chat` + action path) — kept intentionally per **ADR 0001**; this issue is where that ADR is revisited and the fallback removed once the agent path is proven in production.
- Update/remove `backend/tests/test_gemini_service.py` accordingly.

## Acceptance
- No production code imports `services.gemini_service` helpers.
- `python -m pytest backend/tests/ -q` green.
- A short decision record in `docs/decisions/` recording the retirement.

AndresL230:
Part 1 of 2 landed: PR #472 (merged; full lane + oracles green incl. the first live greeting-turn journey). The learn.py/streaming half is agent-only — rung ladder redesigned (nonstream fast-tier fallback with the writes-guard reaching inside the fallback via the sapling_wrote stamp; retryable/413 gating on the client), /start-session JSON and /action agent-ified, legacy chat paths + the five prompt templates deleted. Part 2 (documents.py legacy pipelines, deleting services/gemini_service.py, ADR 0024) is in flight and closes this issue.

AndresL230:
Closed via #472 (#151a — agent-only rung ladder in learn.py/chat_stream) + #473 (#151b — documents.py cutover + deletion of `services/gemini_service.py`). Evidence:

- The module and its tests are gone; every production LLM call now runs as a Pydantic AI agent behind `agents/_providers.py`. ADR 0024 records the decision and supersedes ADR 0001's legacy-fallback clause (the framework adoption stands).
[...]

PR #461, opened 2026-07-29
fix(learn): consume /learn?resume= deep links (#164) + ADR-0020 interrupted-turn Retry + promoted streaming journeys (#356)

### LINKED ISSUE #164: Dashboard 'Where you left off' and Tree session links are dead — /learn?resume= and ?session= are never read
**Severity:** High
**Labels:** bug, frontend, P1
**Location:** `frontend/src/components/screens/Learn.tsx:86-105` (reads only `topic`/`mode`/`suggest`); broken callers `frontend/src/components/screens/Dashboard.tsx:571` (`/learn?resume=<id>`), `frontend/src/components/screens/Tree.tsx:228` (`/learn?session=<id>`).

### Description
The refactored Learn screen reads only `searchParams.get("topic"|"mode"|"suggest")` and never reads `resume` or `session`. Dashboard's "Where you left off" cards push `/learn?resume=<sessionId>` and Tree's recent-session rows push `/learn?session=<sessionId>`, so both land on the Learn **entry** screen with no session resumed. Learn already has `handleResume`/`resumeSession` — it's simply never wired to the URL.

### Steps to reproduce
1. From `/dashboard`, click a "Where you left off" card.
2. Land on the Learn "Start a session" screen; the past conversation is not resumed. Same from Tree's session rows.

### Expected vs actual
- **Expected:** the named past session opens and hydrates its messages.
- **Actual:** the param is ignored; a blank start screen renders — the button looks dead.

### Suggested fix
Add an effect in `LearnInner` reading `resume`/`session` on mount and calling `handleResume(id)`; unify callers on a single `?resume=` contract.

### Acceptance criteria
- Clicking a Dashboard "Where you left off" card or a Tree session row opens that exact session with its messages loaded.

### LINKED ISSUE #356: Complete PR #349 streaming-tutor manual e2e smoke (items 3,4,6,7 + Stop UX decision)
## Summary

PR #349's manual e2e smoke was partially completed during the streaming-crash investigation. Items **1 (progressive render), 2 (progress card moves), 5 (Stop control present), 8 (greeting streams)** were verified live in the browser, and the crash on tool-calling turns was found and fixed. The items below remain **unverified** and need a live run on a supported runtime.

## Remaining smoke items (from the #349 description)

- [ ] **3 — Force Rung 1 (legacy fallback):** break the agent so it fails *before any token*; confirm the legacy reply arrives as a single message and **exactly one** user+assistant row pair persists (no double-write).
- [ ] **4 — Kill backend mid-turn:** confirm the bubble shows interrupted and, after refresh, there is **no phantom partial assistant message** in history.
- [ ] **6 — Rung 3 (stream fails to open):** make `POST /api/learn/chat/stream` fail to open; confirm the client transparently completes the turn via the non-streaming `/api/learn/chat` JSON fallback.
- [ ] **7 — Switch sessions mid-stream:** confirm no leaked stream in the Network tab and no stale bubble.
- [ ] **5 — Product decision:** Stop mid-turn (and during a Rung-3 fallback) currently **discards the partial reply and offers no Retry** — confirmed live in the smoke. Decide: keep the partial + mark interrupted + offer Retry (per the spec), or amend the spec.

## Verified environment recipe (from the #349 investigation)

> WARNING: Do NOT run the backend on Python 3.14 — it is unsupported and independently throws `Event loop is closed`. Use 3.12/3.13 (Docker base is `python:3.12`, CI is 3.13).

1. **Backend:** `uv venv --python 3.13`; install `backend/requirements.txt` minus `torch`/`docling`/`transformers` (lazy OCR deps, not needed for chat); then
   `dotenv -f backend/.env.staging run -- python -m uvicorn main:app --host 127.0.0.1 --port 5001`
2. **Frontend:** `NEXT_PUBLIC_LOCAL_MODE=false BACKEND_URL=http://localhost:5001 NEXT_PUBLIC_API_URL=http://localhost:5001 SESSION_SECRET=<staging value> npx next dev`
   (LOCAL_MODE **must** be false — mock mode never opens a stream and the middleware auth is bypassed.)
3. **Auth:** mint a `sapling_session` HMAC cookie for `seed-user-demo` with the staging `SESSION_SECRET`. Format: `base64url(json{"user_id","exp"}) + "." + base64url(HMAC_SHA256(payload_b64))`. Set it via `document.cookie`. (The backend also accepts `?auth_token=<token>` for curl.)
4. **Align client identity:** `localStorage.sapling_user = {"id":"seed-user-demo","name":"Demo Student","avatar":""}` then reload — otherwise the client sends `user_id=local-user-001` and every `require_self` 403s (and the course list shows empty).
5. Target `seed-user-demo` has a seeded CS101/MATH210/BIO110 knowledge graph (`python -m db.seed_staging`), so item-2-style mastery bumps land on existing concepts.

## Acceptance criteria
- [ ] Each item above verified with a screenshot / captured evidence.
- [ ] Item 5 product decision recorded (and implemented, or the spec corrected).

## Notes
Depends on the streaming-crash fix in #349 (fresh google client per stream) — these items only pass on a runtime where that fix is present.

_Filed by Claude Code during the PR #349 e2e-smoke investigation._

AndresL230:
## What

Bundle B1 of the backlog clear — the tutor-resume P1 plus the streaming smoke tail, in one PR because they share files and journeys.

### #164 — dead resume deep links
`Learn.tsx` read only `topic`/`mode`/`course`/`suggest` off the URL; Dashboard's "Where you left off" cards push `/learn?resume=<id>` and Tree's session rows pushed `/learn?session=<id>`, so both stranded users on the session picker. Now:

- `readResumeParam` (exported, unit-tested) accepts `resume` (canonical) and `session` (legacy alias); Tree unified on `?resume=`.
- A mount effect consumes the param **once per id** (guard ref — the mode-sync URL rewrite preserves params and must not re-trigger it) and calls `resumeSessionById`, which is **server-authoritative**: topic/mode/course come off the resume payload, so deep links work for sessions outside the 10-row recent list.
- A quiet loading state replaces the picker flash while the deep link hydrates.

### #356 item 5 / ADR 0020 — interrupted turns keep the partial + offer Retry
The merged #349 code still discarded partials on Stop and rendered Rung-2 failures as an `Error:` assistant bubble. Per ADR 0020:

- `ChatMsg` gains `interrupted`/`retryText`; ChatPanel renders the partial (dimmed, never blanked) + an "Interrupted" marker + Retry.
- `send`'s ladder: Stop, Rung-2, and a failed Rung-3 JSON fallback all append the interrupted bubble (error detail goes to a toast, not the transcript). Retry drops the failed pair (`removeInterruptedTurn`, pure + unit-tested) and re-sends — safe because the routes persist only on completion.
- A turn aborted by **switching sessions** appends nothing (`sessionIdRef` guard) — otherwise the fix itself would have introduced the item-7 stale bubble.
- Scope note added to ADR 0020: the greeting turn (`beginSession`) returns to the entry screen with the topic draft intact — there is no transcript turn to mark.

### #356 items 4/6/7 — promoted journeys (`frontend/e2e/streaming.spec.ts`)
- **item 6**: `/api/learn/chat/stream` route-aborted → transparent JSON fallback → deterministic reply + exactly one encrypted user/assistant pair (decrypt-verified).
- **items 4+5**: Stop mid-stream → partial kept + interrupted + Retry, **zero rows persisted**; Retry completes and persists exactly one pair.
- **item 7**: switch sessions mid-stream → SSE request torn down (`requestfailed` observed), no stale bubble, both sessions' row counts untouched.
- **#164**: journey enters via the REAL dashboard card (new `dashboard-resume-{id}` testid) in `tutor.spec.ts`.

Mid-stream windows are real: the function-mode seam paces streamed replay (`set_function_stream_delay_ms` — 24-char chunks, reset by `clear_function_handlers`, **no new lane env var**: the e2e handlers module opts in at import with 150ms) and serves a long deterministic reply on the `E2E_SLOW_STREAM` message trigger. Constants stay byte-identical between the JSON and streamed lanes; handler-constant ↔ spec ↔ `test_e2e_function_handlers.py` sync maintained.

### #356 item 3 — live evidence instead of a lane journey
The server-side Rung-1 legacy fallback (`call_gemini_multiturn`) has **no function-mode gate by design** (fallback-only, deleted in #151), so it cannot run deterministically in the e2e lane. New `tests/test_streaming_rung1_live.py` (opt-in `RUN_LIVE_STREAM_RUNGS=1`, billable, `live_llm` lane) proves the real failure shape degrades to a live legacy reply arriving as a single token + done with `on_complete` never called — run green locally. Mechanics stay pinned hermetically in `test_chat_stream.py` / `test_learn_stream_routes.py`.

## Verification

- Backend: full hermetic suite **1316 passed** on the local venv (pydantic-ai 1.89) **and** the streaming-adjacent files re-run green on a lock-pinned scratch venv (**pydantic-ai 1.107.0**, the CI universe).
- Frontend: vitest **254 passed** (12 new), `tsc --noEmit` clean, lint 0 errors (suppressions pruned — the two new testids removed suppressed occurrences).
- Live Rung-1 test: 1 passed against real Gemini.
- Full local e2e cycle (Playwright + oracles under the stack flock): to be run pre-merge; results will be posted below.

Closes #164. Closes #356.

### DIFF
--- a/backend/agents/_providers.py
+++ b/backend/agents/_providers.py
@@ -280,6 +280,31 @@ FunctionModelHandler = Callable[
 ]
 _FUNCTION_HANDLERS: dict[str, FunctionModelHandler] = {}
 
+# Streamed-replay pacing (#356). 0 = off: text parts replay as one whole
+# delta each, exactly the pre-pacing behavior. When a handlers module opts
+# in (set_function_stream_delay_ms), _stream_dispatch re-chunks text into
+# _FUNCTION_STREAM_CHUNK_CHARS-sized deltas with this delay between them, so
+# browser journeys have a real mid-stream window to act in (Stop a turn,
+# switch sessions). Only the streamed lane paces; the JSON lane and the
+# joined stream text stay byte-identical to the handler's constant.
+_FUNCTION_STREAM_DELAY_MS = 0.0
+_FUNCTION_STREAM_CHUNK_CHARS = 24
+
+
+def set_function_stream_delay_ms(ms: float) -> None:
+    """Opt the function-mode STREAMED replay into pacing (see the comment on
+    _FUNCTION_STREAM_DELAY_MS above). Called by boot-time handler modules
+    (agents/function_handlers_e2e.py) rather than wired to another lane env
+    var; `clear_function_handlers()` resets it so in-process tests always
+    start unpaced."""
+    global _FUNCTION_STREAM_DELAY_MS
+    _FUNCTION_STREAM_DELAY_MS = max(0.0, float(ms))
+
+
+def function_stream_delay_ms() -> float:
+    """Public read of the streamed-replay pacing knob (tests pin it)."""
+    return _FUNCTION_STREAM_DELAY_MS
+
 
 class UnregisteredHandlerError(LookupError):
     """Raised by the function-mode dispatch when a task has no registered
@@ -326,9 +351,12 @@ def register_function_handler(task: AgentTask, handler: FunctionModelHandler) ->
 
 
 def clear_function_handlers() -> None:
-    """Drop all registered handlers. Tests call this around each case so
-    process-global registrations never leak between them."""
+    """Drop all registered handlers and reset the streamed-replay pacing
+    knob. Tests call this around each case so process-global registrations
+    (and a handlers module's pacing opt-in) never leak between them."""
+    global _FUNCTION_STREAM_DELAY_MS
     _FUNCTION_HANDLERS.clear()
+    _FUNCTION_STREAM_DELAY_MS = 0.0
 
 
 # ── Boot-time handler registration for out-of-process runs (#392) ──────────
@@ -401,12 +429,32 @@ def _function_model_for(task: AgentTask) -> "Model":
         # as one delta, each tool call as a DeltaToolCall — so a task's handler
         # constants stay byte-identical between the JSON and streamed lanes
         # (spec assertions compare against the same E2E_* constants either way).
+        #
+        # Pacing (#356): when set_function_stream_delay_ms opted this process
+        # in, text is re-chunked into small deltas with a sleep between them —
+        # a real mid-stream window for browser journeys. The delay is read
+        # AFTER _resolve_handler(), which is what imports a boot-time handlers
+        # module that sets the knob, so the very first paced stream paces.
         response = _resolve_handler()(messages, info)
+        delay_s = _FUNCTION_STREAM_DELAY_MS / 1000.0
+        emitted_text = False
         for index, part in enumerate(response.parts):
             kind = getattr(part, "part_kind", "")
             if kind == "text":
-                if part.content:
-                    yield part.content
+                if not part.content:
+                    continue
+                if delay_s > 0:
+                    chunks = [
+                        part.content[i : i + _FUNCTION_STREAM_CHUNK_CHARS]
+                        for i in range(0, len(part.content), _FUNCTION_STREAM_CHUNK_CHARS)
+                    ]
+                else:
+                    chunks = [part.content]
+                for chunk in chunks:
+                    if emitted_text and delay_s > 0:
+                        await asyncio.sleep(delay_s)
+                    emitted_text = True
+                    yield chunk
             elif kind == "tool-call":
                 args = part.args
                 json_args = args if isinstance(args, str) else json.dumps(args or {})
--- a/backend/agents/function_handlers_e2e.py
+++ b/backend/agents/function_handlers_e2e.py
@@ -31,7 +31,21 @@ from __future__ import annotations
 
 from pydantic_ai.messages import ModelResponse, TextPart, ToolCallPart
 
-from agents._providers import FunctionModelHandler, register_function_handler
+from agents._providers import (
+    FunctionModelHandler,
+    register_function_handler,
+    set_function_stream_delay_ms,
+)
+
+# Streamed-replay pacing (#356): re-chunk streamed text into small deltas with
+# 150ms between them, giving the mid-stream journeys (Stop a turn, switch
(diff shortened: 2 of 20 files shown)


AndresL230:
### Code review

Found 2 issues:

1. The two new DB-readback tests in `streaming.spec.ts` decrypt-verify `messages.content` but skip the ciphertext-at-rest assertion (`expect(row.content).not.toBe(plaintext)`). `support/decrypt.ts` says "Specs must always make both" — `decrypt_if_present` echoes plaintext back unchanged, so decrypt-equality alone cannot prove the column was encrypted; if the encryption layer silently broke, both tests would still pass. The sibling `tutor.spec.ts` journey makes both checks.

https://github.com/SaplingLearn/Sapling/blob/7d76204ee51c0d2618799f6af36674f3e20dabfd/frontend/e2e/streaming.spec.ts#L114-L124

2. The `/learn?resume=` deep link still paints the "Start a session" picker for the first frame(s): `resuming` starts `false` and only flips inside a post-paint `useEffect`, so the first committed render with a resume param takes the picker branch — the exact flash the new loading state exists to avoid. A lazy initializer (`useState(() => !!readResumeParam(searchParams))`) closes it deterministically.

https://github.com/SaplingLearn/Sapling/blob/7d76204ee51c0d2618799f6af36674f3e20dabfd/frontend/src/components/screens/Learn.tsx#L357-L359

Lower-confidence items (below threshold, noted for the author): the success-path append after the un-abortable Rung-3 `sendChat` is not guarded by the new `sessionIdRef` check, so a session switch during a Rung-3 fallback can still append the old session's reply into the new transcript; the "three-rung fallback ladder" summary comment above `send` still describes the pre-ADR-0020 Stop/Rung-2 behavior; ADR 0020's "Scope / where it lands" section contradicts its updated Status line.

Issue #117, opened 2026-06-08
[P2] Observability: instrument capture seams (middleware, auth, feature routes)


AndresL230:
## Summary

Instrument the application seams so real user activity, auth events, and errors flow into the `events` table via `events_service.log_event(...)`. This is the cross-cutting wiring that turns the foundation into actual **user-usage analytics + audit + error** data. One-line calls at each seam; no business-logic changes.

## Specifications

Add `log_event` calls at these seams. Each event is **metadata only** (ids, counts, status, durations) — never raw content.

**Middleware** — `backend/services/request_context.py` (`RequestIDMiddleware`):
- On responses with status `>= 400`, emit `error.4xx` / `error.5xx` (`category="error"`) with `payload`: `{ path, method, status_code, duration_ms }` and the `request_id`. Do not log 2xx here (curated domain events cover success).

**Auth** — `backend/services/auth_guard.py`:
- `auth.login` (`category="audit"`) on successful session decode at a login boundary.
- `auth.permission_denied` (`category="audit"`) when `require_self`/`require_admin`/`require_role` reject, with `payload`: `{ reason, route }`.

**Feature routes** (`category="usage"`, explicit one-liners):
- `document.upload` and `document.processed` — `backend/routes/documents.py` (include `category`, `course_id`, page/char counts in payload; no text).
- `quiz.started` / `quiz.completed` — `backend/routes/quiz.py` (include `score`, `total`, `concept_node_id`).
- `chat.message_sent` — `backend/routes/learn.py` (include `mode`, `session_id`; pass the message text via `content=` so only its fingerprint is stored).
- `note.created` — `backend/routes/notes.py`.
- `session.started` / `session.ended` — wherever sessions are opened/closed.

Keep the taxonomy consistent and documented in a short comment block (or `events_service` docstring): `domain.action`, lower-case, dotted.

## Success criteria

- [ ] A 4xx/5xx response produces exactly one `error.*` event with path/method/status/duration; 2xx produces none.
- [ ] A permission rejection produces an `auth.permission_denied` audit event.
- [ ] Uploading a document, taking a quiz, sending a tutor message, creating a note, and starting/ending a session each produce the corresponding `usage` event with the specified payload fields.
- [ ] No raw content appears in any event — `chat.message_sent` stores only `content_fp`.
- [ ] All instrumentation calls are non-blocking (they go through `events_service`) and wrapped so a logging failure can't break the endpoint.
[...]

PR #470, opened 2026-07-30
feat(agents): structured-output retry + validation hardening (#153)

### LINKED ISSUE #153: [P2] Agent platform: structured-output retry + validation hardening
Part of the **Agent migration** epic #152 (platform hardening). Owner: Jack.

## Goal
Make agent structured-output robust: systematic retry-on-validation and defensive handling so malformed/partial model output and provider schema rejections never surface as raw 500s.

## Why
- The `orchestrator-schema-complexity` attempt showed Gemini's structured-output API **rejects complex schemas** before the model even runs — outputs must stay small/flat. That rule is folklore today, not enforced.
- The audit's agents/LLM pass flagged JSON-mode parsing that can crash on malformed output.

## Scope
- **Retry-on-validation:** set `output_retries` (and use `ModelRetry`) on the agents so a transient invalid output retries instead of failing the request.
- **Codify the output-type rule** from the attempt: keep agent output types under ~5 flat fields, prefer enums to constrained strings, no nested optional models. Add a short note in `backend/agents/` (or `docs/decisions/`) so it isn't relearned.
- **Defensive parsing** for any remaining raw `call_gemini_json` paths until they migrate (#143–#147): guard against non-JSON / partial output.
- **Map failures to clean errors:** `UnexpectedModelBehavior` / `UsageLimitExceeded` / provider schema-rejection → a typed error response (and the terminal SSE `error` event on streaming routes), not a raw 500. Reuse the `request_id` correlation.

## Files
`backend/agents/*.py` (output types + `output_retries`), `backend/agents/_providers.py`, `backend/services/gemini_service.py` (JSON helpers, until retired by #151).

## Acceptance
- A deliberately malformed model output retries and then degrades to a typed error, not a 500 — covered by a test in `backend/tests/`.
- Output-type guidance documented and the current agents conform.

AndresL230:
## What

Second link of the B7 seam-endgame chain. The failure-mapping half of the issue was already true on this base (re-verified route by route); what landed:

- **The schema budget, codified and enforced**: a structural test auto-discovers all 20 registered agents and walks every structured output's JSON schema against the documented budget — with the exact schema Gemini rejected in the 2026-05-03 attempt kept as a negative control. A future rich schema fails in CI, not against provider 400s.
- **Output retries (2) on all 14 structured-output agents** — including a real dep-universe catch: pydantic-ai 1.107's recommended `retries={"output": 2}` dict form **silently breaks 1.89 at retry time** (reproduced empirically); the correct-everywhere kwarg is used and the structural test rejects the dict form. `WORKER_LIMITS` bumped so the last retry surfaces as `UnexpectedModelBehavior`, not a misfiled `UsageLimitExceeded`. Free-text agents (the streaming tutor especially) deliberately untouched — that's the rung ladder's jurisdiction. Recovered retries get a warning log; the #117 taxonomy is untouched.
- **The ADR-0023 bare-newline tutor reply**, fixed at the stream layer: whitespace-only completed replies prefer the joined streamed chunks, else take the Rung-1 ladder — never an empty persisted assistant row, `on_usage` still fires, and the at-most-one-of persistence invariant holds (all 13 prior invariant tests untouched-green).
- **Legacy `call_gemini_json` sites** degrade cleanly on non-JSON instead of 500ing the last-resort path (both retire in #151).

## Verification

- Backend **1490 passed** + ruff clean (1.89 venv); **354 passed** across all affected files under the lock-pinned **1.107** venv; evals replay green across all six datasets **under both venvs** (cassettes untouched by design — no prompt changes).
- 7 red-first tests (blank-reply ladder ×4, blank JSON-path fallback, non-JSON degrades ×2) + a function-mode acceptance suite pinning exactly-3-requests → typed 502.
- Full local e2e cycle pre-merge; results below.

Closes #153.

## Summary by CodeRabbit

* **New Features**
  * Added bounded validation retries for generated summaries, quizzes, flashcards, study materials, and other structured content.
  * Improved recovery when generated content is malformed or incomplete.

* **Bug Fixes**
  * Prevented blank responses from being saved as successful chat messages.
  * Added safe fallbacks for invalid model responses, including typed temporary-unavailability errors.
  * Improved streamed chat handling when responses contain only whitespace.

* **Reliability**
  * Added warnings when content succeeds after validation retries.
  * Expanded automated coverage for schema limits, retries, fallbacks, and streaming behavior.

### DIFF
--- a/backend/agents/__init__.py
+++ b/backend/agents/__init__.py
@@ -4,12 +4,55 @@ Exposes the per-run cost guardrails every agent call site passes via
 the explicit `usage_limits=` kwarg. We keep two shapes: workers don't
 call tools (tool_calls_limit=0) and never need many requests; the
 orchestrator does call tools and may iterate.
+
+OUTPUT-SCHEMA BUDGET (#153, codifying
+docs/attempts/2026-05-03-orchestrator-schema-complexity.md)
+---------------------------------------------------------------------
+Gemini's native structured-output API compiles the output schema into a
+constrained-decoding automaton and REJECTS rich schemas outright ("too
+many states for serving") — the request fails before the model runs.
+The original document-orchestrator output (classification + summary +
+concepts + optional syllabus in one model) died exactly this way. Every
+agent `output_type` must therefore stay small and flat:
+
+- at most ~8 fields per object (aim for < 5), and one level of object
+  nesting below the root (root -> list[Item] is the ceiling);
+- no optional NESTED models (`Sub | None` fields — optional scalars are
+  fine); if a sub-result is conditional, run a second agent or compose
+  in route code;
+- enums as flat string Literals, not constrained/patterned strings;
+- rich results are composed deterministically in code from several
+  small agent runs (see agents/document.py), never asked for whole;
+- `PromptedOutput` is the escape hatch when a *specific* Pydantic
+  feature trips the provider (e.g. `date` fields on syllabus
+  extraction) — it moves enforcement into the prompt + local
+  validation but does NOT exempt the schema from this budget.
+
+`tests/test_agent_output_schemas.py` walks every registered agent and
+fails CI when an output schema exceeds the budget, so the next rich
+schema dies in review, not against Gemini's 400s.
+
+Validation-retry policy (#153): idempotent structured-output generation
+agents run with an output-retry budget of 2 (`retries=2` on tool-less
+workers; `output_retries=2` on quiz, whose tool retries stay at the
+default) so a transiently malformed output re-rolls instead of failing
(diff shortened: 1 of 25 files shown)


AndresL230:
Review pass complete: two reviewers fully clean (one re-verified the retry semantics against both installed pydantic-ai sources); the history pass found two real gaps in the new blank-reply path, both fixed — whitespace deltas no longer count as 'content seen' for the client ladder, and a blank turn whose tools already wrote mastery ends in a terminal error rather than a fallback that would double-apply (the no-writes case keeps the fallback). Contract tests rewritten accordingly. All gates green incl. lockvenv. e2e cycle next.

Issue #149, opened 2026-06-08
[P1] Agent migration: graph-grounded tutor retrieval + tool-use loop


AndresL230:
Part of the **Agent migration** epic (roadmap: August). Owner: Jack.

## Goal
Make the tutor genuinely graph-aware: graph-grounded retrieval for the tutor system prompt plus a tool-use loop for on-the-fly graph updates.

## Scope
- Graph-grounded retrieval feeding `build_system_prompt` (`backend/routes/learn.py:239`) — pull the relevant subgraph for the topic/course instead of dumping the whole graph.
- Tool-use loop so the tutor reads/writes graph state mid-turn via tools.

## Coordination
- **Depends on / overlaps #127** (the chat tutor currently emits no mastery deltas — its mastery-update tool must exist for the write half to be meaningful).
- Pairs with #70/#74 (streaming + SSE graph deltas, Andres).

## Acceptance
- Tutor prompt is scoped to a retrieved subgraph, not the full graph.
- The tutor can update mastery/concepts mid-conversation via tools, reflected in the graph.

AndresL230:
Closed by PR #469 (merged; full lane 28/28 + oracles clean at the rebased head). Evidence: the tutor gained its graph read surface (read_graph_neighborhood — course-scoped, name-seeded, depth-1, ids never leave the tool — and read_concepts_for_user), every agent turn carries a deterministic ≤1.5k-char GRAPH CONTEXT seed block (the agent path previously inlined NO graph context; the full-graph dump was legacy-only and user-wide), the three legacy prompt sites now use the same compact course-scoped serializer (killing the cross-course leak), TUTOR_LIMITS prevents legitimate multi-tool turns from tripping into the silent legacy fallback, and chat_tutor joined the offline eval harness through the new TutorRetrieval seam (ADR 0023; ADR 0021 decision 4 superseded). Eval deltas before→after: ExpositoryHasStructure 0.812→1.000, GraphToolUsed 0.938→1.000, GroundedConcept 0.812→0.875, MasteryUpdateEmitted 0.688→1.000 — the pre-change cassettes live in the PR's first commit for audit. Note: AC(2) (mid-turn mastery/concept writes) had already shipped via #127; this PR completes AC(1) + the retrieval seam.

PR #471, opened 2026-07-30
feat(agents): prompt-injection hardening on student content (#150)

### LINKED ISSUE #150: [P2] Agent migration: prompt-injection hardening on student-supplied content
Part of the **Agent migration** epic (roadmap: September). Owner: Jack.

## Goal
Guardrails and prompt-injection hardening on student-supplied content (uploaded docs, notes, chat) before it reaches agent prompts.

## Scope
- Treat student/document text as untrusted: delimit/quote it in prompts, add instruction-injection resistance, and constrain tool use so injected text can't trigger unintended tool calls or data exfiltration.
- Pairs with the usage-limit guardrails from #127 (`ORCHESTRATOR_LIMITS`).

## Acceptance
- A documented set of injection test cases (in `backend/tests/`) that the agents resist.
- No path where document/note content can override system instructions or invoke tools it shouldn't.

AndresL230:
## What

Third link of the B7 seam-endgame chain; gates the public beta. Full detail in the commit message. The shape:

- **Containment, single source of truth**: `services/prompt_safety.py` — delimited untrusted-content envelopes with data-not-instructions framing and delimiter-forgery neutralization (embedded BEGIN/END copies are defanged, case/whitespace-insensitively, idempotently). Applied at every assembly boundary where student or peer-derived text enters a prompt: RAG chunks, the graph seed block, the legacy COURSE MATERIALS + shared-context JSON, and the tool→LLM boundary. Storage stays raw.
- **Instruction hardening**: a shared injection-guard paragraph across the tutor's three preambles, note_chat, quiz, and the legacy preamble — plus the academic-integrity block held out of #149 for exactly this PR.
- **Tool-use constraint**: the mastery-delta schema is clamped to the instructed band, so an injected "set my mastery to 1.0" fails validation into #153's retry loop; a contract test freezes that no tool signature exposes ids to the model.
- **Honest eval handling**: replay keys on case names and would have stayed silently green on prompt changes — chat_tutor and quiz_generation were re-recorded (the re-record also landed ADR-0023's tracked prompt-shape fix after the bare-newline quirk reproduced live). All evaluators 1.000 on both datasets; GroundedConcept ratcheted 0.875 → 1.0.
- **Documented scope calls**: the student's own message channel stays unwrapped (it *is* their instruction channel), catalog chunks stay trusted, the misconceptions tool stays unregistered pending real consent enforcement, and the tool-less document workers are a noted follow-up.

## Verification

27-case red-first injection suite (14 red at base) including an end-to-end FunctionModel test proving injected tool returns reach the model enveloped. Backend **1518 passed** + ruff clean; **278 passed** under the lock-pinned 1.107 venv; evals replay green across all six datasets. Full local e2e cycle pre-merge; results below.

Closes #150.


AndresL230:
Review pass complete: findings all fixed — the convergent one (sibling tools on the same agent returning raw content: session history could replay a student-planted REAL forged delimiter; both graph readers returned raw concept names) now neutralized with tests; zero-width delimiter forgeries stripped before matching; three doc staleness items squared; and the reviewers' FYI (last_session_summary unwrapped) folded in since it's the same class. send_to_tutor's preface deliberately stays unwrapped (the student's own instruction channel, consistent with the PR's documented scope call). Backend 1521 + lockvenv + evals ×6 green. e2e cycle next.

AndresL230:
Pre-merge gate: full lane **28/28 passed** + oracles clean. Merging.

PR #472, opened 2026-07-30
refactor(learn): agent-only rung ladder — retire the legacy chat paths (#151a, 1/2)

AndresL230:
## What

Part **1 of 2** of the final `gemini_service` cutover (#151) — the learn.py/streaming half, i.e. the rung-ladder redesign the batch plan flagged as the reason #151 is not a deletion chore. Part 2 (documents.py legacy pipelines, deleting `services/gemini_service.py`, ADR 0024) follows; **the issue closes with part 2**.

Highlights (full detail in the commit message):

- **Rung 1 redesigned**: the seam keeps its exact contract under a new name (`nonstream_fallback`) — but the fallback is now a fresh **non-streaming agent turn on the fast tier** (a different, faster model = a genuinely better second chance), via the extracted `_chat_turn_json` / new `_start_session_agent`.
- **The writes-guard, generalized**: no fallback ever runs after tools wrote graph/mastery — terminal error with an additive `retryable: false`, which the client now honors (closing the pre-existing hole where Learn's ladder re-ran blank turns over `/chat` and double-applied mastery, defeating #470's server-side guard).
- **Honest failure semantics**: UsageLimitExceeded → 413 (deterministic; the client does not retry it), UnexpectedModelBehavior → 502 retry-friendly, on `/chat`, `/start-session`, `/action`.
- **`/start-session` JSON gets its first agent implementation** (legacy was its *primary*, not a fallback) and `/action` — which had **no** agent path at all — is agent-ified in place. Both covered by the existing `chat_tutor` function-mode handler via task dispatch, pinned by a new route test.
- **New greeting-turn journey** (the scoping pass found zero journeys touching `/start-session`) exercising the entry screen, the lazy-session contract, and the DB-polled transcript.

## Verification

- Backend **1511 passed** + ruff clean; **192 passed** across all touched files under the lock-pinned pydantic-ai 1.107; frontend **350 passed** + tsc clean; evals replay green ×6 (prompts untouched by design).
- Red-first throughout (18 backend + 7 frontend red pre-implementation); the redesigned live Rung-1 test still proves the cross-version exception-wrapping seam with real Gemini.
- Full local e2e cycle pre-merge (also the first live run of the new greeting journey); results below.

Part of #151 — do not auto-close.

## Summary by CodeRabbit

* **New Features**
  * Tutor sessions now use an agent-based learning flow across starting sessions, chatting, streaming, and actions.
  * Sessions are created lazily after the first chat message.
  * Streaming errors now indicate whether retrying is safe.
  * Added topic-picker test hooks and improved custom-topic learning coverage.

* **Bug Fixes**
  * Prevented unsafe retries that could duplicate learning updates.
  * Improved fallback behavior for interrupted or failed tutor responses.
  * Added clearer handling for usage limits and temporary tutor failures.


AndresL230:
Review pass complete: one confirmed finding (both deep reviewers converged) — the writes-guard protected entry to the fallback but the fallback is itself a tool-calling run, and its own write-then-fail path hardcoded retryable:true, letting the client re-run a turn whose writes had landed. Fixed by stamping write-state on the raised exception at both route helpers and reading it in the fallback's terminal error; red-first tests at both layers. ADR 0020's stale 'wrote nothing' safety argument amended. Backend 1515 + lockvenv green. e2e cycle next.

AndresL230:
Pre-merge gate: full lane green (incl. the FIRST live run of the new greeting-turn journey) + oracles clean. Merging — #151 stays open for part 2.

Issue #484, opened 2026-07-30
Decision needed: course_chunks.chunk_text is plaintext while documents.extracted_text is encrypted


AndresL230:
Split from the #323 audit (verified @ d1d724f) — needs Andres's call, not a unilateral patch. The same string is encrypted into `documents.extracted_text` (documents.py:1066-1070, migration 0030) and chunked into `course_chunks.chunk_text` in the clear (rag_service.py:190), cross-user shared by design (content-addressed ids). Coherent-by-necessity — pgvector similarity can't run over ciphertext — but the encryption boundary is byte-derivably defeated and 0017's encrypted-column list doesn't mention it. Options: searchable-encryption design, chunk-store isolation/scoping, or an explicit accepted-risk ADR. Recommend at minimum an ADR recording whichever stance is chosen.

AndresL230:
Decided and recorded as **ADR 0025** (merged in #503). Andres delegated the call; this is the reasoning behind it.

## The issue's premise is wrong

> pgvector similarity can't run over ciphertext

`match_course_chunks` (migration 0039) computes `1 - (c.embedding <=> query_embedding)`, orders by `c.embedding <=> query_embedding`, and filters `WHERE c.embedding IS NOT NULL`. `chunk_text` is only SELECTed as payload — the ranking never reads it. Nothing queries it by content either (no `ILIKE`/`LIKE`/FTS anywhere in `services/`, `routes/`, `scripts/`).

So encrypting it does not block retrieval, and this is **cheaper** than the issue implies.

## But the same fact makes it partial — and that's the part worth recording

The `embedding` column cannot be encrypted, because pgvector must compute distance over it. An embedding is a lossy but real representation of its source text, and embedding-inversion recovers substantial content from vectors alone.

So encrypting `chunk_text` **does not make the row opaque**. Without that written down, the next person reads "chunk_text is encrypted" and concludes chunk content is confidential. It isn't — this closes trivially-readable plaintext and restores consistency with `documents.extracted_text`, which is worth doing on its own, but it is defense in depth, not confidentiality. The ADR states it as a consequence rather than burying it.

## The decision

Encrypt **uniformly** — document *and* catalog chunks — with ids still computed on **plaintext**.

[...]

Issue #481, opened 2026-07-30
RAG vector store has no migration — course_chunks/match_course_chunks exist only as dashboard DDL


AndresL230:
Split from the #323 audit (verified @ d1d724f). `course_chunks`, the `match_course_chunks` RPC, and `CREATE EXTENSION vector` appear in ZERO .sql files (425 checked — migrations/, seed.sql, archive/). A database replayed purely from `python -m db.migrate` has RAG silently dead end-to-end: `retrieve_chunks` swallows the RPC failure into `[]` (rag_service.py:137-139), `_get_catalog_chunk` degrades to "", and indexing failures vanish into a fire-and-forget log line — the tutor answers ungrounded with no error, metric, or user-visible signal, and nothing in the suites or oracles asserts the table exists.

Fix: a numbered migration (IF NOT EXISTS, mirroring 0032's reconcile pattern) codifying the extension + table + RPC to match the live dashboard DDL (shape per docs/superpowers/specs/2026-07-03-rag-document-chunking-design.md — verify against staging before writing). This is the highest-leverage remaining RAG item and it is a migration, not a feature.

AndresL230:
Fixed in #495 (`9ca30d4`). Closing — backend-only, no UI surface to verify.

**Migration 0039** codifies `CREATE EXTENSION vector`, `course_chunks` (with its indexes) and the `match_course_chunks` RPC. Shape verified against **live production and staging** by reading a real row and calling the RPC — not from the design doc, which describes intent while the database is what the code actually talks to.

Two things that probe turned up, worth recording:

- The live RPC is **768-dim** and the code's `_OUTPUT_DIM` is 768 — they match, so there was no latent dimension bug hiding behind the missing migration.
- Direct psycopg to staging is unreachable from here (the IPv6-only endpoint refuses and the pooler hosts reject), so the probe went through PostgREST.

Every statement is `IF NOT EXISTS` / `CREATE OR REPLACE`, mirroring 0032's reconcile pattern — a no-op where the objects already exist.

**The `ragstore` oracle** closes the gap this issue names ("nothing in the suites or oracles asserts the table exists"). It checks the extension, the table and the RPC — deliberately **not** their contents, since an empty `course_chunks` is normal on a fresh stack and a missing one is the bug.

## Proven the only way that counts

[...]

Issue #330, opened 2026-07-09
Add ability for users to manually add a concept (graph node) from the frontend


AndresL230:
## Summary

Users can **remove** their own knowledge-graph concepts from the site, but there is **no way to manually add a single named concept**. Today concepts (graph nodes) only appear via agent/pipeline flows (notes extract-concepts, document upload, tutor chat, quizzes). We want a first-class "add a concept" capability, surfaced on the Tree.

## Current state (grounded in code)

**Remove — already shipped:**
- `Tree.tsx:144` → `deleteGraphNode(userId, selected.id)` (confirm-guarded) → `DELETE /api/graph/{user_id}/nodes/{node_id}` → `graph_service.delete_node` (`backend/services/graph_service.py:381`), which also cascades the node's edges and refreshes offering analytics.

**Add — not supported directly:**
- `routes/graph.py` has no `POST .../nodes` route. Every node is created through `apply_graph_update` (`backend/services/graph_service.py:461`), fed `new_nodes` only from documents (`routes/documents.py`), tutor (`routes/learn.py:752`), and notes-extract (`agents/tools/graph.py:30` `apply_concepts_to_graph`).
- `Tree.tsx` has zero add/create-node UI (grep-verified).

**Good news:** the backend engine already exists — `apply_graph_update` merges `new_nodes` with case/whitespace-insensitive dedup backed by the `UNIQUE (user_id, course_id, concept_name)` constraint (migration 0023). So re-adding an existing concept merges rather than duplicates. We only need a thin route + UI.

## Capabilities to build

**Backend**
- [ ] Add `graph_service.add_node(user_id, concept_name, course_id=None, initial_mastery=0.0)` — a thin wrapper over `apply_graph_update({"new_nodes": [...]})` (parallels the existing `add_course`/`delete_node` helpers), returning the created-or-merged node (id, concept_name, course_id, mastery).
- [ ] Add `POST /api/graph/{user_id}/nodes` in `routes/graph.py`, `require_self`-guarded, body `{ concept_name: str, course_id?: str }`. Trim/validate non-empty; return the node. Signal merge-vs-create so the UI can toast appropriately.

**Frontend**
- [ ] `lib/api.ts`: `addGraphNode(userId, conceptName, courseId?)` wrapper (mirrors `deleteGraphNode`).
[...]

Issue #323, opened 2026-07-07
RAG retrieval system for the tutor (roadmap M4)


AndresL230:
## Context

Roadmap milestone **M4** ("Graph-aware RAG tutor end to end") now explicitly names the **RAG retrieval system** as a deliverable, but there's no dedicated tracking issue. The closest existing issue is #149 (graph-grounded tutor retrieval + tool-use loop), which covers the *retrieval + tool-use loop inside the tutor agent* — not the RAG system as a whole.

This issue tracks the RAG retrieval system that grounds the tutor: how student documents, notes, and knowledge-graph context get indexed and retrieved to build the tutor's prompt context.

## Scope (to refine)

- Retrieval over student-supplied content (documents, notes) + knowledge-graph nodes/edges
- Chunking / indexing strategy and where embeddings live (Supabase / pgvector vs. external store)
- Retrieval interface consumed by the tutor agent — feeds the `build_system_prompt` seam in `backend/routes/learn.py`
- Query-performance work on `graph_nodes` / `graph_edges` (shared with #149)
- Retrieval-quality / relevance eval harness (relates to #148 extraction-accuracy eval harness)
- Respect column-level encryption: decrypt document/note content at the read boundary before indexing/injecting (see `services/encryption.py`)

## Relationships

- Parent roadmap milestone: **M4** (canopy roadmap — canopy.saplinglearn.com/roadmap)
- Sibling: #149 — graph-grounded tutor retrieval + tool-use loop (the agent-side loop that consumes this system)
- Adjacent epic: #152 — agent migration

_Filed as a tracking issue from a roadmap-planning session; scope to be refined by the backend owner._

AndresL230:
**Epic audit — delivered vs remaining, verified against origin/main @ d1d724f (post-#149/#151/#154, the seam endgame).** Closing this epic: its load-bearing scope is delivered or superseded, and everything real that remains is now tracked in sharper follow-ups (#481–#484).

| Scope bullet | Status | Evidence |
|---|---|---|
| Retrieval over **documents** | 🟡 Delivered, hardening remains | Single-sited via `_prepare_chat_run` (learn.py:505-590) for ALL five tutor entry points + quiz grounding; indexing on streaming upload (fire-and-forget). Hardening → #482 |
| Retrieval over **notes** | 🔴 Not started | notes.py has zero RAG references; only the 1500-char send-to-tutor paste bridge → #483 |
[...]

PR #485, opened 2026-07-30
feat: B10 stragglers — manual add-concept (#330) + rooms semantics (#405)

AndresL230:
Part of #330 and part of #405 (both left open per closure policy — Andres verifies the UI surfaces and closes). Completes B10 alongside the #323 audit (closed with follow-ups #481–#484).

## #330 — manually add a concept
- `graph_service.add_node`: thin wrapper over `apply_graph_update` (dedup/edges/analytics-refresh stay in one place); resolves the anchor id to its NAME (the edge machinery is name-keyed); reports merge-vs-create; returns the canonical row. `POST /api/graph/{user_id}/nodes` (require_self, 422 on blank name; `course_id` required — NULL-course dedup is always-distinct).
- Learn's rail composer now actually PERSISTS (it was client-only — concepts vanished on refresh): optimistic node kept, canonical id swapped via pure unit-tested helpers, rollback + toast on failure, merge toast on dedup.
- Tree gains '+ Add concept' (single-course filter only; anchors to the selected node). Testids registered doc+eslint; Tree.tsx entered the lint enforcement list with its 8 pre-existing untagged elements baselined per the documented recipe.
- Journey `tree-add-concept.spec.ts`: persist → render (a11y list by node id) → case-drifted re-add merges, original row survives.

## #405 — rooms semantics (Andres's decision: real ownership + public rooms)
- **Migration 0038**: `owner_id := created_by` backfill + NOT NULL (ownership is real and transferable later; `created_by` stays the immutable creator record), `is_public` DEFAULT false NOT NULL, `updated_at` DEFAULT now() + backfill.
- `create_room` populates owner/topic/course/is_public; kick authorization keys on **owner_id**; membership changes touch `updated_at`.
- **Public surface**: `GET /api/social/public-rooms` (explicit projection — cannot leak `invite_code`) + invite-less `POST /api/social/public-rooms/{id}/join` (403 private, idempotent). Frontend: create form gains topic/course/Public, sidebar shows topic + Public badge, and a Public-rooms discovery list with one-click Join. A fuller discovery surface stays future work per the decision.
- Journey `room-public.spec.ts`: UI create with topic+public → DB semantics assert → second user joins invite-less → membership row.

## Tests
Backend: 17 new (add_node service/route ×8, rooms semantics ×9); suite 1525 passed; ruff clean. Frontend: 397 passed / tsc / lint 0 errors (rebased over #480's suppressions prune — converged). Two promoted journeys. Full local e2e cycle before merge.

## Summary by CodeRabbit

* **New Features**
  * Added public study rooms with optional topics and courses.
  * Users can discover and join public rooms without invite codes.
  * Added manual concept creation for selected courses, including duplicate merging and optional relationships.
  * Added public-room labels, member counts, and discovery controls.

* **Bug Fixes**
  * Improved room ownership permissions and membership update tracking.
  * Added safer defaults and validation for room data.

* **Tests**
  * Added coverage for public rooms, concept creation, merging, and related user flows.

### DIFF
--- a/backend/db/migrations/0038_rooms_semantics.sql
+++ b/backend/db/migrations/0038_rooms_semantics.sql
@@ -1,0 +1,26 @@
+-- 0038: give the #405 rooms columns real semantics ("real ownership +
+-- public rooms").
+--
+-- 0032 added topic/course/owner_id/updated_at/is_public nullable and
+-- unpopulated (its header deliberately deferred the semantics to #405).
+-- This migration backfills and constrains them:
+--   * owner_id := created_by for existing rows — ownership starts with the
+--     creator and is TRANSFERABLE later; created_by stays the immutable
+--     creator record. Authorization (kick, future transfer/delete) keys on
+--     owner_id from here on.
+--   * is_public: DEFAULT false + NOT NULL (NULL was neither public nor
+--     private). false = invite-only; true = joinable without an invite via
+--     POST /api/social/public-rooms/{id}/join and listed by
+--     GET /api/social/public-rooms.
+--   * updated_at: DEFAULT now() + backfill from created_at; membership
+--     changes touch it in code (routes/social.py::_touch_room).
+
+UPDATE rooms SET owner_id = created_by WHERE owner_id IS NULL;
+ALTER TABLE rooms ALTER COLUMN owner_id SET NOT NULL;
+
+UPDATE rooms SET is_public = FALSE WHERE is_public IS NULL;
+ALTER TABLE rooms ALTER COLUMN is_public SET DEFAULT FALSE;
+ALTER TABLE rooms ALTER COLUMN is_public SET NOT NULL;
+
+UPDATE rooms SET updated_at = created_at WHERE updated_at IS NULL;
+ALTER TABLE rooms ALTER COLUMN updated_at SET DEFAULT now();
--- a/backend/db/seed_local_rich.py
+++ b/backend/db/seed_local_rich.py
@@ -450,7 +450,12 @@ def seed_rooms() -> None:
     for room_id, name, invite_code, created_by in _ROOMS:
         h.upsert(
             "rooms",
-            {"id": room_id, "name": name, "invite_code": invite_code, "created_by": created_by},
+            # owner_id mirrors create_room's semantics (#405): ownership starts
+            # with the creator. 0038 makes it NOT NULL and SQL has no
+            # cross-column default, so the seed must set it explicitly or a
+            # from-empty replay fails on the INSERT.
+            {"id": room_id, "name": name, "invite_code": invite_code,
+             "created_by": created_by, "owner_id": created_by},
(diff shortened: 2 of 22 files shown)


AndresL230:
### Code review

Found 1 issue:

1. Migration 0038 makes `rooms.owner_id` NOT NULL with no default, but the rich seed never sets it — a from-empty replay (`make e2e-up`, and CI's `e2e.yml`, which both run migrations then seed) fails the first rooms INSERT and aborts the whole stack boot. Found independently by two reviewers and verified directly. Masked on any existing local volume, where the upsert becomes an UPDATE.

https://github.com/SaplingLearn/Sapling/blob/60cbdcc9c8542422300900200a88a6a96fc42313/backend/db/migrations/0038_rooms_semantics.sql#L21-L26

https://github.com/SaplingLearn/Sapling/blob/60cbdcc9c8542422300900200a88a6a96fc42313/backend/db/seed_local_rich.py#L448-L456

Fixed, along with eight sub-bar findings the same pass raised (all fixed regardless of score, per the standing rule):

- **`join_public_room` / `join_room` check-then-insert** raced to a raw 500 on the `room_members` PK — now `upsert(on_conflict="room_id,user_id")`, with an in-flight guard on the Join button.
- **`addConcept` could render a duplicate node**: a live tutor turn's `stream-<name>` placeholder wasn't absorbed by the canonical row, so the client showed two nodes for one row until the next full refetch. `reconcileNodes` now absorbs placeholders by normalized name (real-id nodes are left alone), and its edges are retargeted too.
- **Anchor resolution wasn't course-scoped** — an anchor from another course resolved by name *inside* the target course and could wire the edge to the wrong node.
- **Leader/kick UI still keyed on `created_by`** while the backend switched to `owner_id`; latent today (they're equal) but breaks the moment transfer ships. `RoomOverviewData` now declares `owner_id`.
- **`PublicRoomsList` swallowed its fetch error** (a broken fetch was indistinguishable from "no public rooms" — the #463/#185 smell) and never revalidated its own list after a join.
- **Temp-id collision**: `node-new-${Date.now()}` is ms-resolution with no in-flight guard; now carries a monotonic suffix.
- Two doc/comment corrections: the `AddNodeBody` rationale inverted `UNIQUE NULLS NOT DISTINCT` semantics, and the testids doc miscategorized Tree's baselined elements (they're the search input, five detail-panel actions and two fullscreen toggles — `graph-zoom-*` lives in `KnowledgeGraph2D.tsx`).
- `room-public.spec.ts` used the wrong seeded display name for `rich-user-second` ("Casey" → "Sam").

One finding was **documented rather than fixed**: `apply_graph_update`'s dedup folds case in Python while the 0023 constraint is case-*sensitive*, so two concurrent adds differing only in case can both land. That's pre-existing behavior inherited by `add_node`, not introduced here; the docstring no longer claims it's "UNIQUE-backed", and a citext/functional index is the real fix.

Issue #341, opened 2026-07-15
[P2] /settings overflows <main> by 56px under top-nav — account deletion controls below the fold


AndresL230:
Found while reviewing #335 (which fixes this class of bug on five other screens but misses this one).

### The bug

`frontend/src/components/screens/Settings.tsx:232`:

```jsx
<div style={{ display: "flex", height: "calc(100vh - 112px)" }}>
```

`112px` is the `TopBar` height. Root total = `112 + (100vh - 112)` = **exactly `100vh`** — correct for the *sidebar* layout where `<main>` is `100vh`.

But in the **top-nav** layout, `ShellFrame` stacks a 56px `TopNav` above `<main>`, so `<main>` is `100vh - 56px`. The screen overflows by exactly 56px.

### What the user sees

The inner pane at `:233` already has `overflowY: auto`, so `/settings` shows **two nested scrollbars**, and the bottom 56px of the pane sits below the fold — **including the account-deletion controls on the `data` tab**.

Mobile always takes the top-nav path (`useSidebar = pref === "sidebar" && !isMobile`), so this is the mobile experience.

### Fix

Use `FullHeightScreen` / `height: 100%` per #335's approach, or reference the exported `TOP_NAV_HEIGHT` (`TopNav.tsx:75`) which is already available and unused here.

### Same class, lower severity — worth folding in

- `Tree.tsx:311` — `height: calc(100vh - 240px)`; overflows top-nav `<main>` by ~29px, and the 3D graph canvas is sized off this box via `ref`/`size.h` so it renders oversized too.
- `Gradebook/Landing.tsx:218` and `Gradebook/Course.tsx:555` — `minHeight: calc(100vh - var(--row-h))`. `--row-h` is a **density token** (40/34/48px per `globals.css:144,157,160`), not a nav height — so the offset is semantically wrong *and* density-dependent (spurious scroll of 16/22/8px).

### Root cause

Screens hand-roll viewport math instead of resolving against `<main>`. #335 collapses this onto ShellFrame's two `100vh` declarations, which is the right shape — these are the stragglers.

AndresL230:
**Scope narrowed — the headline bug is fixed** (verified @ 7fa6cb6).

`Settings.tsx` no longer hard-codes `height: calc(100vh - 112px)`; it now wraps in `<FullHeightScreen>` (imported :5, opens :221, closes :633). The 56px top-nav overflow and the buried account-deletion controls on the `data` tab are gone.

[...]

PR #487, opened 2026-07-31
fix(ui): replace viewport-math residuals with FullHeightScreen (#341)

AndresL230:
Replaces the three surviving `calc(100vh - …)` residuals with `FullHeightScreen` + flex, and adds the journey that proves the one with a user-visible consequence.

## Why these were wrong

`ShellFrame` renders two layouts: a sidebar (its `<main>` is the full `100dvh`) or a horizontal `TopNav` above `<main>` (so `<main>` is `100dvh - 56px`). Any constant subtracted from `100vh` can only be calibrated for one of them — and none of them accounted for the density preference, which retunes the padding tokens of the chrome being subtracted.

## The consequential one — Tree.tsx

`height: calc(100vh - 240px)` wrapped the element a `ResizeObserver` watches; its `contentRect` is passed straight through as `<KnowledgeGraph width height>`. So this was not a stray scrollbar — the graph **canvas itself** rendered at the wrong size in whichever layout the constant wasn't tuned for.

Now: `FullHeightScreen` root + `flex: 1` / `minHeight: 0` on the graph row, so it absorbs exactly what `<main>` has left below the TopBar and filter row. Correct in both layouts, at any density, by construction rather than by calibration.

## The semantic ones — Gradebook Landing + Course

`minHeight: calc(100vh - var(--row-h))` subtracted a **density token** (40/34/48px per `globals.css`) as if it were a nav height. Both now sit in a `FullHeightScreen` with `flex: 1 0 auto` on their `<main>` — fill the remaining space, keep growing when content is taller (the `minHeight` semantic, preserved).

## Test

`frontend/e2e/tree-viewport.spec.ts` asserts the invariant that was violated: on `/tree` the graph canvas fits inside the shell scrollport — no overshoot, no scrollable overflow — in **both** layouts, driven off the `sapling_layout` localStorage pref. It is a new spec rather than an addition to `graph.spec.ts`, which declares itself data-only and deliberately reads no x/y.

`FullscreenGraph` (the overlay variant) already used `flex: 1` and is untouched.

## Gates

- `tsc --noEmit` — clean
- `npx vitest run` — 55 files, 399 tests passed

part of #341

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screen sizing across sidebar and top-navigation layouts.
  * Fixed Tree and Gradebook views to fill available space without unwanted overflow.
  * Preserved legitimate scrolling for content that exceeds the available height.

* **Tests**
  * Added end-to-end coverage validating viewport fit, shell sizing, canvas settling, and content scrolling.


AndresL230:
### Local e2e cycle — red at base, green at fix

Run as two phases inside one `flock` of the stack lock, at `3ee666c` vs base `432c207`, with `SAPLING_MODEL_MODE=function`.

Both phases matter: a journey that passes on unfixed code proves nothing, so phase 1 runs the new spec against the **base** build and requires it to fail.

**Phase 1 — base `432c207` + the new spec: 2 failed (expected)**

```
layout "topnav": canvas 480px vs scrollport 664px — overshoots by 27px
layout "sidebar": <main> is 686px but needs only 647px
                  (647px available, 606px of content) — 39px is phantom
```

Two things worth noting in those numbers:

- The tree failure is **topnav only** — sidebar passed at base, which is exactly what a `240px` constant calibrated for the sidebar layout would do. 27px measured against the issue's estimate of ~29px.
- The gradebook phantom is **39px**, which is `--row-h` (40px at the default density) to within rounding — the density-token-as-nav-height bug, measured rather than argued.

**Phase 2 — fix `3ee666c`: 35/35 journeys, oracles clean**

```
35 passed (1.3m)
```

### Two corrections this caught

The gradebook assertion took two revisions before it was worth keeping, both caught by the base run rather than by review:

1. First version asserted the page must not scroll **at all**. Wrong — under topnav the scrollport is 56px shorter and the seeded content genuinely needs more than it, so scrolling there is correct. It failed the *fixed* build for a legitimate reason.
2. Second version derived content height from `main.scrollHeight`, which is floored at the element's own client height — so any phantom height was mirrored into the number it was compared against, and it **passed at base**. Vacuous.

It now measures the content extent from the in-flow children (skipping `AmbientOrbs`, which is `position: fixed`), which the box's own sizing cannot influence.

### Gates

`tsc --noEmit` clean · `npm run lint` 0 errors · `npx vitest run` 55 files / 399 tests passed

Jose-Gael-Cruz-Lopez:
Automated post-merge review pass (Claude Code) on `caf04b0` (the review ran while this merged — treat as follow-up input). The core verdict is positive: all three residuals from #341 are properly root-caused, the `flex: 1 0 auto` replacement expresses what the old `minHeight: calc(...)` was failing to say, the residual sweep is genuinely complete for the bug class, and the red-at-base regression journey (both layouts, overshoot *and* undershoot, real scrollport measurements) is exemplary.

**One real regression worth a follow-up fix:**

1. **Sticky TopBar travel is now clipped on the two gradebook screens** (`Gradebook/Landing.tsx:229`, `Course.tsx:519`). At base, TopBar's sticky containing block was the shell `<main>`'s full scrollable content, so it stayed pinned for the whole scroll. At head its parent is `FullHeightScreen`'s fixed `height: 100%` box — exactly one scrollport tall — so once page content exceeds ~2× the scrollport, the pinned header (breadcrumb, Curve/Letter Scale actions) detaches and scrolls away mid-page. These are precisely the screens that deliberately still page-scroll, and long assignment tables plausibly cross the threshold. Verified: `FullHeightScreen.tsx` sets `height: "100%"` with the `...style` spread after it, so passing `style={{ height: "auto", minHeight: "100%" }}` on just these two screens restores full sticky travel while keeping the fill/phantom-height behavior the journey asserts (a min-height flex column still distributes free space to a `flex: 1 0 auto` child). The pinned-chat screens must keep fixed height — a `grow` prop on FullHeightScreen would make the two modes explicit.

**Minor follow-ups:**

2. One same-class residual survives: `(shell)/notetaker/page.tsx:1028` — `minHeight: calc(100vh - 280px)` on the fullscreen editor textarea (defanged by the enclosing scroll pane, but the same blind-constant pattern).
3. `viewport-fit.spec.ts` measures `/gradebook` Landing only; `Course.tsx` got the identical fix but no measurement — extending the loop to the seeded `/gradebook/[courseId]` is cheap, and a scroll-then-assert that TopBar is still pinned on a tall Course page is the assertion that would have caught #1.
4. `Tree.tsx:280-300` — the filter row got `flexShrink: 0` but the TopBar header keeps default shrink; at very short viewports the TopBar is now the only compressible child and squashes (previously the page scrolled). `flexShrink: 0` on the header keeps the chrome intact.

Issue #290, opened 2026-06-30
frontend: re-home pending onto app surfaces + add confirmation beat


AndresL230:
[P1-D] `app/pending/page.tsx` is a bare centered flex (no card, no shell). Give it the app surfaces (`.card`) and add a brief 'you're in' confirmation so beta→pending isn't glow→silence. Deferred from token-unification Phase 1 (PR #286). Tracker: docs/superpowers/followups/2026-06-30-token-unification-followups.md · Audit: docs/frontend-rhythm-audit.md

AndresL230:
## Proposal — re-home pending + add a confirmation beat (B11, verified @ d1d724f)

**Not implemented — proposal only.**

### Current state

`app/pending/page.tsx` is 42 lines and has **no container at all** — the sprout SVG (`:29-32`), heading (`:33`), copy (`:36`), and sign-out button (`:39`) float directly on a radial gradient (`:16-27`), the same one Onboarding uses. There is no "you're in" beat, so the beta flow really does go glow → silence, as the issue says. (The landing's beta modal *does* have a confirmation at `page.tsx:765-770`, but that fires before this page and is unrelated to arriving here.)

`.card` (`globals.css:210-213`) gives `--bg-panel`, a 1px `--border`, `--r-lg` (16px), and `--shadow-sm` — **but no padding**, which is why callers set their own.

### Proposal

Wrap the content in `.card` + `--pad-lg`, and add a brief "you're in ✓" beat — event-driven, ≤600ms, no infinite loop (the audit's motion rule). Keep the copy. Align the backdrop with whatever #289 settles on so the two transition-zone pages match.

### Scope and risk

- A few hours.
- **Minimal risk**, one hard constraint: `data-testid="pending-gate"` (`:17`) and `data-testid="pending-signout"` (`:39`) must survive unchanged — the file is in the testid enforcement list and journeys select on them.

AndresL230:
Landed via #488 (`9bbacc6`). **Leaving this open for your UI verification.**

## What changed

`/pending` was 42 lines with no container — sprout, `<h1>`, `<p>` and a button floating directly on the radial gradient. Now:

- **A surface.** Content sits on a real `.card` with `padding: var(--pad-lg)` (`.card` supplies bg/border/radius/shadow but no padding — that part is the caller's job).
- **A confirmation beat.** The sapling draws itself once, then the message steps in behind it. Settles at ~560ms, single pass, no loop — inside the ≤600ms rule from your proposal.

[...]

PR #488, opened 2026-07-31
feat(ui): give the approval gate a surface and a confirmation beat (#290)

AndresL230:
Sign-up **ends** on `/pending`, and the page was 42 lines with no container — a sprout, an `<h1>`, a paragraph and a button floating directly on a radial gradient. Beta glow into silence, with no "you're in" moment anywhere in it.

## What changed

**A surface.** The content now sits on a real `.card`. `.card` supplies bg/border/radius/shadow but no padding — that's the caller's job — so the card sets `var(--pad-lg)`.

**A confirmation beat.** The sapling draws itself once, then the message steps in behind it. Everything settles by ~540ms: single pass, no loop.

**Copy that does the confirmation work the layout can't.** "Your account is ready" states what actually happened; "We'll email you" names the channel that the old "We'll reach out" left vague.

## Why CSS and not JS

`globals.css` already carries a global `prefers-reduced-motion` reset, so a CSS entrance is automatically safe — a JS-driven one would have to re-implement that guard itself.

It reuses the motion vocabulary that already exists (the `fade-in`/`slide-up` keyframes and the `.anim-d*` delay utilities) instead of inventing a parallel one. The new `.pending-*` rules are declared **above** `.anim-d*` deliberately: the `animation` shorthand resets `animation-delay`, so the delay utilities have to come later to win.

The sprout's stroke carries `pathLength={1}` so the draw keyframe can dash it without measuring, and its three subpaths run stem-then-veins — so it reads as the sapling growing rather than a generic fade.

## Anchors

`data-testid="pending-gate"` and `data-testid="pending-signout"` are unchanged. Both are already registered in `docs/frontend-testids.md` (surface table and inventory) and `app/pending/page.tsx` is already in the eslint enforcement list, so neither half needed updating.

`src/app/pending/page.test.tsx` (new, red-first) pins the anchors, the sign-out behaviour, the card surface, and the beat's **finiteness** — it fails if a future pass introduces an `infinite` animation or drops a fill mode from a delayed step.

## Gates

- `tsc --noEmit` — clean
- `npx vitest run` — 56 files, 403 tests passed

part of #290

### DIFF
--- a/frontend/src/app/globals.css
+++ b/frontend/src/app/globals.css
@@ -224,6 +224,28 @@ body { font-size: 14px; line-height: 1.5; }
 .fade-in { animation: fade-in var(--dur) var(--ease); }
 .slide-up { animation: slide-up var(--dur-slow) var(--ease); }
 .fade-up { animation: fade-in 0.5s var(--ease); animation-fill-mode: both; }
+
+/* Approval gate (#290) — the "you're in" beat.
+   The sapling draws itself once and the message steps in behind it; the
+   whole sequence settles by ~540ms. Finite by construction: every rule runs
+   a single pass, and `both` holds each step at its start frame so a delayed
+   element never flashes at its resting value first. Declared ABOVE the
+   .anim-d* utilities on purpose — these shorthands reset animation-delay, so
+   the delay utilities have to come later to win. The global
+   prefers-reduced-motion reset collapses the whole sequence. */
+@keyframes pending-draw  { from { stroke-dashoffset: 1; } to { stroke-dashoffset: 0; } }
+@keyframes pending-swell { from { opacity: 0; transform: scale(0.7); } to { opacity: 0.2; transform: none; } }
+.pending-card { animation: slide-up var(--dur-slow) var(--ease) both; }
+.pending-sprout-body {
+  transform-box: fill-box; transform-origin: 50% 100%;
+  animation: pending-swell 340ms var(--ease) 60ms both;
+}
+.pending-sprout-stem {
+  stroke-dasharray: 1;
+  animation: pending-draw 420ms var(--ease) 120ms both;
+}
+.pending-step { animation: fade-in var(--dur) var(--ease) both; }
+
 .anim-d0 { animation-delay:   0ms; }
 .anim-d1 { animation-delay:  80ms; }
 .anim-d2 { animation-delay: 160ms; }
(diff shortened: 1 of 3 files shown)


Jose-Gael-Cruz-Lopez:
Automated review pass (Claude Code). Verified findings only — overall this is a careful, well-tested change: scope matches #290 exactly, middleware gating is untouched (`middleware.ts:91` still bounces unapproved users; `/pending` stays outside the protected matcher), the `.pending-*`-before-`.anim-d*` cascade reasoning is correct and documented, testids are pre-registered, and 4/4 unit tests + tsc + eslint reproduce clean at head.

**Should fix before merge**

1. **`pending/page.tsx:100` — the copy promises an email the system never sends.** "We'll email you as soon as your access is approved" names a specific channel, but approval is `PATCH /users/{user_id}/approve` (`backend/routes/admin.py`), which only flips `is_approved` — there is no email-sending code anywhere in the repo (verified: the only smtp/mailer-ish grep hits are `SESSION_SECRET` false positives). If admins manually email approved users, ignore this; otherwise soften to "We'll let you know…" — a user waiting on an email that never arrives is worse than vague copy.

**Minor**

2. `globals.css:511-518` — the reduced-motion reset zeroes `animation-duration` but not `animation-delay`, so under reduced motion the `both`-fill steps still stagger-reveal over ~240ms (held invisible, then snapping in). Consistent with existing `.fade-up` surfaces, but the comments claiming the sequence fully collapses overstate it. Cheap global fix: add `animation-delay: 0.01ms !important` to the reset.
3. `pending/page.tsx:102-108` — the sole interactive element (`.btn`, ~37px) is under the 44px touch-target convention. Pre-existing class, not a regression, but this page is likely hit on mobile right after sign-up; a min-height on this instance is a cheap win.

**Follow-up worth an issue (pre-existing, not this PR):** nothing on `/pending` ever re-checks approval — an approved user with the tab open (or even refreshing `/pending`) stays on the waitlist screen until they manually navigate elsewhere. A one-shot "check again" against `/api/auth/me` would close the loop the confirmation beat opens. Also: no e2e journey directly visits `/pending` (only negative-space assertions); a seed-unapproved → `pending-gate` → sign-out journey would make coverage direct.

Verdict: **ready to merge once the email-copy question is resolved.**

AndresL230:
### Local e2e cycle — green

Full cycle inside one `flock` of the stack lock, at `30d38f7` (rebased on `main`), `SAPLING_MODEL_MODE=function`.

```
35 passed (1.3m)
```

Gates: `tsc --noEmit` clean · `npm run lint` 0 errors · `npx vitest run` 56 files / 403 tests.

Note there is no journey that drives `/pending` directly (only incidental references in `auth-session.spec.ts` and `smoke.spec.ts`), so the cycle here is a regression check that the `globals.css` additions didn't disturb anything else — the beat itself is covered by `page.test.tsx`, which pins its finiteness rather than its appearance.

Issue #288, opened 2026-06-30
frontend: extract shared hero-card from --surface-hero (de-dup beta/sign-in modal)


AndresL230:
[P1-E] Replace the duplicated inline 24px warm-gradient box in the beta modal (`app/(public)/page.tsx`) and `components/SignInModal.tsx` with one shared component backed by `--surface-hero` / `--surface-hero-shadow`. Deferred from token-unification Phase 1 (PR #286). Tracker: docs/superpowers/followups/2026-06-30-token-unification-followups.md · Audit: docs/frontend-rhythm-audit.md

AndresL230:
## Proposal — shared hero-card (B11, verified @ d1d724f)

**Not implemented — proposal only.**

### Current state

The tokens this issue asks us to adopt exist and have **zero consumers**: `--surface-hero` and `--surface-hero-shadow` (`globals.css:765-766`). Meanwhile the literal gradient is copy-pasted **five times** — `page.tsx:756`, `:785`, `:819`, `:871`, and `SignInModal.tsx:226` — and has already drifted:

- shadow alphas: `0.12` / `0.15` / `0.18` (and the token's own `0.12`)
- radii: `20` vs `24`
- **hue**: every inline copy uses cool slate `rgba(15,23,42,…)`; the token uses green `rgba(19,38,16,…)`

The close-button markup (`page.tsx:801-814` ≡ `SignInModal.tsx:238-252`) and the logo+wordmark row are duplicated verbatim too.

### Proposal

One `<HeroCard>` component — a **named variant of `.card`** backed by the two tokens (this is what the rhythm audit's P1-E prescribes) — adopted at all five sites, with the close button and logo row folded in.

### Scope and risk

- ~half a day plus a screenshot review.
- **Call this out explicitly:** adopting the token is a *visible* change, not a pure refactor — the shadow hue shifts from slate to green. That's the intended correction (the app is warm), but you should see a before/after and approve it.
- Low risk otherwise; `SignInModal.tsx` is already in the testid enforcement list, so no lint surprises.

AndresL230:
Landed via #489 (`b335dfb`). **Leaving this open for your UI verification** — this one is a deliberate *visible* change, so it needs your eyes.

## What changed

[...]

PR #489, opened 2026-07-31
refactor(ui): extract the shared hero-card surface (#288)

AndresL230:
The warm gradient behind the sign-in and beta modals was inlined as a literal at **five** sites, and had already drifted at four of them. `--surface-hero` / `--surface-hero-shadow` sat in `globals.css` with **zero** consumers.

| Site | Role | Radius | Shadow |
|---|---|---|---|
| `page.tsx:756` | beta success modal | 20 | `rgba(15,23,42,0.15)`, no inset |
| `page.tsx:785` | beta signup shell | 24 | `rgba(15,23,42,0.12)` + inset |
| `page.tsx:819` | modal left panel (inner) | — | background only |
| `page.tsx:871` | modal right panel (inner) | — | background only |
| `SignInModal.tsx:226` | sign-in modal | 24 | `rgba(15,23,42,0.18)` + inset |

## What replaced it

- **`.card--hero`** — a named variant of `.card`, used as `class="card card--hero"` (declared after `.card` so it wins).
- **`.hero-surface`** — the gradient *alone*, for the two inner panels, which must not restate the parent's border/radius/shadow.
- **`<HeroCard>`** — a thin wrapper that owns those two classes and forwards ref, className and every other prop. Callers keep their own layout and dialog semantics.

Both classes are now the tokens' only consumers.

## This is a visible change, and an intended one

Adopting the token moves the shadow hue from cool slate `rgba(15,23,42)` to the warm `rgba(19,38,16)` the rest of the app uses, unifies the alphas at `0.12`, and gives the beta success modal the 24px radius and inset highlight its two siblings already had. Before/after screenshots in a comment below.

## On the literal fallbacks

`var(--surface-hero, <literal>)` in both rules is deliberate. The tokens are scoped to `.public-surface, .landing-page`. I verified every consumer sits inside that subtree today — `SignInModal` is mounted only from the landing page, and nothing here portals to `document.body` — so the fallback is **defensive, not load-bearing**. For a shared component it is the cheap guard against the first mount that isn't.

## Deliberately not folded in

The close button and the logo/wordmark row. The close buttons are not duplicated verbatim after all — they differ in offset (18 vs 14) and only one carries a testid. The wordmark appears at **20** sites repo-wide with no shared component, so extracting it for 2 of them would leave a half-migration; #111 touches every icon site anyway, which is the coherent place for it.

## Tests

`HeroCard.test.tsx` pins the class contract and prop/ref forwarding, and adds two anti-drift guards: **no inlined copy of the gradient may exist anywhere in `src/`**, and both CSS rules must keep a literal fallback (including the inset half, or token and fallback would render as two different surfaces).

`data-testid="signin-modal"` and `"signin-close"` are unchanged; no testid surface was added, so neither half of the registry needed updating.

## Gates

- `tsc --noEmit` — clean
- `npx vitest run` — 56 files, 403 tests passed

part of #288

### DIFF
--- a/docs/superpowers/followups/2026-06-30-token-unification-followups.md
+++ b/docs/superpowers/followups/2026-06-30-token-unification-followups.md
@@ -22,10 +22,18 @@ Each is its own plan / PR. Severity tags map to `docs/frontend-rhythm-audit.md`.
       (`.btn .btn--primary .btn--xl`) — app button language, no motion. The glow was **deleted
       outright rather than converted** to a `--brand-glow` entrance, which satisfies the intent
       (no perpetual motion, one button species); `--brand-glow` consequently has no consumers.
-- [ ] **[P1-E] Extract one hero-card primitive from `--surface-hero`.** Replace the duplicated
+- [x] **[P1-E] Extract one hero-card primitive from `--surface-hero`.** Replace the duplicated
       inline 24px warm-gradient box in the beta modal (`app/(public)/page.tsx`) and
       `components/SignInModal.tsx` with a single shared component backed by `--surface-hero` /
       `--surface-hero-shadow`.
+      ✅ **Done (verified 2026-07-30):** landed with #288. `.card--hero` (a named variant of
+      `.card`) plus `.hero-surface` (the gradient alone, for panels nested inside a hero card)
+      now consume both tokens, and a thin `<HeroCard>` wrapper owns the pair. All five literal
+      sites are converted — four in `app/(public)/page.tsx`, one in `components/SignInModal.tsx`
+      — and `HeroCard.test.tsx` fails if a literal copy of the gradient reappears anywhere in
+      `frontend/src`. The literal fallbacks (`var(--surface-hero, …)`) exist because the tokens
+      are scoped to `.public-surface, .landing-page`; every consumer is inside that subtree
+      today, so they are defensive rather than load-bearing.
 - [ ] **[P1-C] Re-home onboarding** (`components/screens/Onboarding.tsx`) into the app's
       layout/spacing/type — `--pad-*` spacing, the shell type scale — instead of a centered
       card floating in a radial-gradient void.
--- a/frontend/src/app/(public)/page.tsx
+++ b/frontend/src/app/(public)/page.tsx
@@ -7,6 +7,7 @@ import { useScrollLock } from '@/lib/useScrollLock';
 import { Network, Sparkles, FilePlus2, Brain, CalendarClock, Users, PenSquare } from 'lucide-react';
 import HowItWorks from '@/components/HowItWorks';
 import SignInModal from '@/components/SignInModal';
+import { HeroCard } from '@/components/HeroCard';
 import { BRAND_FOREST } from '@/lib/brand';
 import { Button } from "@/components/ui";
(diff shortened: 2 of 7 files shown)


AndresL230:
### Local e2e cycle — green (and it caught a real break)

Final cycle at `8e6d2b4`, one `flock`, `SAPLING_MODEL_MODE=function`:

```
35 passed (1.4m)
```

Gates: `tsc --noEmit` clean · `npm run lint` 0 errors · `npx vitest run` 58 files / 409 tests.

### Worth flagging: the first cycle failed, and no cheaper gate could have caught it

The review-fix commit closed the hero-surface block comment early — the original `*/` stayed put while a new paragraph was inserted above it, so the tail of the comment became raw stylesheet text. `make e2e-up` died building the frontend:

```
CssSyntaxError: globals.css:1007:13: Unclosed string
```

780 lines below the actual mistake, at the first quote postcss happened to reach.

**CI's `lint + tsc + vitest` job was green on that broken file.** Nothing in the fast lane reads `globals.css` as CSS: eslint lints JS/TS, tsc checks types, vitest never imports the stylesheet. The first thing that parses it is the Next production build.

So this PR also adds `src/app/globals.test.ts` — a dependency-free structural scan (comments, strings, brace depth) that fails in the fast lane and points at the right line. It's hand-rolled rather than a postcss import because postcss is only a *transitive* dependency here and the installed copy already drifts from CI's. It tests itself too: a second case feeds it each broken shape to prove it doesn't silently return "fine".

### Rebase note

Rebased onto `main` after #488 landed. Both PRs touch `globals.css`; the two ordering invariants were re-verified after the merge — `.card--hero` (`:242`) still follows `.card` (`:210`), and `.pending-*` (`:278`) still precedes `.anim-d*` (`:306`).

Jose-Gael-Cruz-Lopez:
Automated review pass (Claude Code), originally run against `f8eb2ac` and re-checked after the force-push to `8e6d2b4`. Independently verified at the original head: extraction is genuinely complete (zero residual gradient literals outside the token/fallbacks/test constant), the PR body's drift table is accurate row-for-row against base, `.card--hero` overrides every `.card` property at equal-and-later specificity so `card` adoption is cascade-neutral, testids preserved, tsc clean + full vitest suite green. The e2e-cycle comment and the `[P1-E]` tick landed since — both good.

**Still outstanding**

1. The PR body promises "Before/after screenshots in a comment below," but no comment contains any image. Since this PR *declares* a deliberate visible change (cool→warm shadow hue, unified alphas, success modal gaining radius+inset), the visual sign-off it leans on isn't actually on the record — post the screenshots or amend the body.

**Minor**

2. `HeroCard.test.tsx` — the anti-drift walk still scans only `.ts`/`.tsx` (its own comment at ~:65 notes globals.css "is not walked (only .ts/.tsx)"). A future inlined copy of the gradient in a *new* `.css` file (e.g. a CSS module) would evade the guard. The legitimate fallback literals are an argument for excluding `app/globals.css` by path, not the whole `.css` extension.

**Note for future edits:** `.card--hero` currently overrides everything `.card` declares, so the pairing is purely nominal today — any future addition to `.card` (padding, hover elevation, transition) will silently restyle all three modals. Worth one sentence in the globals.css comment block.

Verdict: **ready to merge once the screenshots are posted** (or the body reference removed).

PR #492, opened 2026-07-31
perf(frontend): cut per-frame work in the hero canvas, backdrop and graph (#111)

AndresL230:
Four hot spots, no visual change. **Half of #111 was already fixed on main** and is untouched here — the landing canvas already respects reduced-motion, per-node `shadowBlur` is gone (0 hits repo-wide), the floating-cards RAF already resolves its NodeList once, and the spotlight rect is already cached. Those claims in the issue body are stale; this PR is only the genuinely remaining work.

## 1. Hero canvas link pass — measured 0.353ms → 0.097ms per frame (3.6×)

An uncapped double loop measured all **25,425** pairs of the 226 projected nodes every frame, to draw the ~**213** that actually qualify. 99.2% of the distance checks were wasted.

A node can only link within `70 * sc` px, so binning into a grid whose cell is the *largest* such radius means every possible partner sits in the 3×3 block around it and nothing else is ever measured.

Extracted to `lib/linkPairs.ts` so "identical output" is **provable rather than asserted**: `linkPairs.test.ts` compares it to the all-pairs reference across 40 random clouds plus dense, sparse, negative-coordinate, asymmetric and degenerate cases — and checks the emission **order** matches, because these are translucent overlapping strokes and order decides the composited image. One test also guards against a future "simplification" quietly restoring the double loop.

## 2. AtmosphericBackdrop — 14 gradients per frame → 14 blits, at half the rate

It rebuilt 14 full-viewport radial gradients *every frame at 60fps*, app-wide, mounted twice by `ShellFrame`. Everything a gradient depends on — colour, radius, depth-derived alpha — is fixed when the orb is created; only x/y move.

Each orb is baked once into an offscreen canvas and blitted; the loop is capped at ~30fps (ambient drift moves well under a pixel per frame). Re-bakes only when the device pixel ratio actually changes.

The reduced-motion still-frame path is preserved exactly, **including** the existing behaviour that switching reduced-motion on mid-session stops the loop.

## 3. KnowledgeGraph2D — three problems

- **Tick storm.** `.on("tick", forceRerender)` reconciled the whole SVG per physics tick. Ticks now coalesce into at most one render per animation frame. This does **not** touch the deterministic settle: `simulation.tick()` does not dispatch tick events (only the internal timer does), so the reduced-motion/test-mode path still renders once, synchronously, and stays deterministic.
- **Tooltip churn.** `setTooltipPos` was the unconditional first statement of `onPointerMove` — moving the mouse over empty canvas re-rendered the entire graph to update a value nothing reads unless `hovered` is set. Now gated, with the position seeded on pointer-enter so the first frame still lands correctly.
- **O(E·N) lookups.** Edge rendering used `.find()` per endpoint per edge per frame; the link filter did two `.some()` scans per edge. Both now use a Map/Set.

The offscreen `IntersectionObserver` pause and unmount cleanup are preserved, and the queued tick render is cancelled on unmount so it can't fire against a dead tree.

## 4. Images and framer-motion

Intrinsic `width`/`height` on the **19** brand-icon sites (a fixed-size asset whose box was only pinned in inline CSS, so nothing reserved it before styles resolved), and `decoding="async"` on the lazily-loaded ones. `next/image` deliberately not adopted, per your call.

`HowItWorks` (~54 `motion.*`, the landing page's only framer-motion consumer) is now `next/dynamic`, so the library leaves the critical path for a section below the fold. **Not** `ssr: false` — the markup still server-renders.

**`Study.tsx` and `Calendar.tsx` are deliberately left alone.** The brief listed them, but both wrap their screen's *primary content* in the motion element — deferring it would defer the content itself, which is worse than the bundle cost it saves. `Calendar.test.tsx`'s framer-motion mock stays green either way.

## One suppression count moves

`KnowledgeGraph2D` `react-hooks/refs` 13 → 14. Building the id→node Map is one more use of a ref-derived value during render, in a file where that pattern is already baselined 13 times. I verified in isolation that the Map is the sole cause (restoring `.find()` returns it to 13), and tried three formulations — all cost the same. The trade is one more instance of an existing suppressed pattern in exchange for deleting a per-frame O(E·N) scan.

## Gates

- `tsc --noEmit` clean · `npm run lint` 0 errors (35 warnings, one fewer than before) · `npx vitest run` 59 files, 415 tests

part of #111

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved responsiveness of the interactive knowledge graph and animated background effects.
  * Optimized homepage canvas link rendering for smoother visual performance.
  * Improved image loading and decoding across chat, reports, settings, and other screens.
  * Added image dimensions to logos to reduce layout shifts during page loading.

* **Tests**
  * Added comprehensive coverage for visual link-generation behavior and edge cases.

### DIFF
--- a/frontend/eslint-suppressions.json
+++ b/frontend/eslint-suppressions.json
@@ -55,7 +55,7 @@
       "count": 1
     },
     "react-hooks/refs": {
-      "count": 13
+      "count": 14
     }
   },
   "src/components/KnowledgeGraph3D.test.tsx": {
--- a/frontend/src/app/(public)/about/page.tsx
+++ b/frontend/src/app/(public)/about/page.tsx
@@ -73,7 +73,7 @@ export default function AboutPage() {
           >
             <img
               src="/sapling-icon.svg"
-              alt="Sapling"
+              alt="Sapling" width={26} height={26}
               style={{ width: 26, height: 26, flexShrink: 0, position: "relative", top: -2 }}
             />
             <span
@@ -281,7 +281,7 @@ export default function AboutPage() {
           }}
         >
           <div style={{ display: "flex", alignItems: "center", gap: 8 }}>
-            <img src="/sapling-icon.svg" alt="Sapling" style={{ width: 20, height: 20 }} />
+            <img src="/sapling-icon.svg" alt="Sapling" width={20} height={20} style={{ width: 20, height: 20 }} />
             <span style={{ fontSize: 14, color: "var(--text-muted)" }}>Sapling · © 2026</span>
           </div>
           <div style={{ display: "flex", flexWrap: "wrap", gap: 24 }}>
--- a/frontend/src/app/(public)/careers/CareersList.tsx
+++ b/frontend/src/app/(public)/careers/CareersList.tsx
@@ -50,7 +50,7 @@ export default function CareersList() {
           >
             <img
               src="/sapling-icon.svg"
-              alt="Sapling"
+              alt="Sapling" width={26} height={26}
               style={{ width: 26, height: 26, flexShrink: 0, position: 'relative', top: -2 }}
             />
             <span
@@ -363,7 +363,7 @@ export default function CareersList() {
           }}
         >
           <div style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
-            <img src="/sapling-icon.svg" alt="Sapling" style={{ width: 20, height: 20 }} />
+            <img src="/sapling-icon.svg" alt="Sapling" width={20} height={20} style={{ width: 20, height: 20 }} />
             <span style={{ fontSize: 14, color: 'var(--text-muted)' }}>Sapling · © 2026</span>
           </div>
           <div style={{ display: 'flex', flexWrap: 'wrap', gap: 24 }}>
--- a/frontend/src/app/(public)/careers/[slug]/ApplyForm.tsx
+++ b/frontend/src/app/(public)/careers/[slug]/ApplyForm.tsx
@@ -144,7 +144,7 @@ export default function ApplyForm({ job }: { job: Job | null }) {
           >
             <img
               src="/sapling-icon.svg"
-              alt="Sapling"
+              alt="Sapling" width={24} height={24}
               style={{ width: 24, height: 24, position: 'relative', top: -1 }}
             />
             <span className="h-serif" style={{ fontSize: 20, color: 'var(--text)' }}>
@@ -533,7 +533,7 @@ export default function ApplyForm({ job }: { job: Job | null }) {
           }}
         >
           <div style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
-            <img src="/sapling-icon.svg" alt="Sapling" style={{ width: 20, height: 20 }} />
+            <img src="/sapling-icon.svg" alt="Sapling" width={20} height={20} style={{ width: 20, height: 20 }} />
             <span style={{ fontSize: 14, color: 'var(--text-muted)' }}>Sapling · © 2026</span>
           </div>
           <div style={{ display: 'flex', flexWrap: 'wrap', gap: 24 }}>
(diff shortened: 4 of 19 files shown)


AndresL230:
### ⚠️ This overlaps PR #490 almost entirely — do not merge either without a decision

I did not know #490 existed when I opened this; my 5-agent review found it. @Jose-Gael-Cruz-Lopez opened #490 at 02:38Z against issue #111, ~50 minutes before this one, from the same base (`432c207`). The two PRs touch **nine of the same files**, including all three hot spots. Whichever merges first leaves the other with a large, error-prone conflict across `page.tsx`, `AtmosphericBackdrop.tsx` and `KnowledgeGraph2D.tsx`.

Posting an honest comparison rather than arguing for mine.

| | #490 | #492 (this) |
|---|---|---|
| hero link pass | axis-reject + squared distance — still O(N²), cheaper constant | spatial binning, sub-quadratic, extracted + equivalence-tested |
| KnowledgeGraph2D ticks | **direct DOM writes**, zero React work per tick | rAF-coalesced `forceRerender` — React still reconciles once per frame |
| backdrop | sprites + 30fps + **pauses on `visibilitychange`** | sprites + 30fps, no visibility pause |
| HowItWorks | `next/dynamic` **`ssr: false`** + placeholder | `next/dynamic`, **SSR kept** |
| Study.tsx | motion subtree extracted to `StudyMotion.tsx` | left alone deliberately |
| suppressions baseline | pruned 13 → **12** | grown 13 → **14** |
| local E2E lane | **could not run** (machine unprovisioned) | ran green, repeatedly |

**Where #490 is better:** the direct-DOM-write graph is the deeper fix and is what the issue's own proposal recommended; the `visibilitychange` pause is a real win this PR lacks; extracting `StudyMotion` is more than I did; and pruning the suppressions baseline beats growing it.

**Where this PR is better:** the link pass is asymptotically better, not just cheaper per pair, and its equivalence to the naive version is *proved* by test rather than asserted. `ssr: false` on `HowItWorks` removes a landing-page section from the server-rendered HTML — a crawler regression on the one page that needs SEO. And this branch has actually been through the local E2E lane.

**Reading #490 also caught a real bug in this PR** that my own review missed: my 30fps cap advanced the orb physics once per *painted* frame, so drift ran at half speed — a visual change, in a PR claiming none. #490 handled that explicitly ("drift speed preserved via elapsed-frame scaling"). Fixed in `463facd`.

My suggestion, for whoever decides: take #490 as the base — it is the more thorough pass and the graph fix is better — and port two things onto it from here: the binned `lib/linkPairs.ts` with its equivalence suite, and dropping `ssr: false` from the `HowItWorks` dynamic import. But that is a call for @AndresL230, not for me.

AndresL230:
Superseded by #490, which is merged.

@AndresL230's call was to take Jose's PR as the base, port anything worthwhile from here, verify it, and merge — which is what happened. Recording what carried over and what didn't:

**Carried over:** the `ssr: false` → SSR-preserving `HowItWorks` import (Jose made that change himself in `23b773f`, crediting this PR's review), and the intermittent CI fix from this branch's investigation (`FlashcardImportModal.test.tsx` never unmounting — the reason this very PR failed CI with 416/416 tests passing).

**Deliberately not carried over:** the spatially-binned `lib/linkPairs.ts`. I benchmarked it against #490's axis-reject loop rather than assuming, and **#490's is faster at the real workload** — 0.068ms vs 0.113ms per frame at the actual 226-node cloud. Binning only wins past ~N=2000; below that, Map allocation and per-point sorting cost more than it saves. Identical pairs at every N. The asymptotically better algorithm was the slower one here, so it stayed out.

**Also not carried over:** the rAF-coalesced tick render. #490 writes simulation positions directly to the DOM, which is strictly better — and is what issue #111's own proposal recommended.

The one thing this PR did that #490 hadn't: run the full local E2E cycle. That has now been run against the merged head — 35/35 journeys, oracles clean.

Opening this without checking for an existing open PR on the same issue was my mistake.

PR #493, opened 2026-07-31
docs(frontend): correct the engine-strict guard — it blocks `npm ci` too

AndresL230:
`frontend/.npmrc` told contributors that `engine-strict=true` was a **write-path-only** guard, because "`npm ci` ignores engine-strict in npm >=7". That is false for modern npm.

Probed against a minimal package carrying the same `>=10.9.0 <11` pin, on **npm 12.0.1 / node v26.4.0**:

| command | `engine-strict` | exit |
|---|---|---|
| `npm ci` | `true` | **1** — `EBADENGINE` |
| `npm install` | `true` | **1** — `EBADENGINE` |
| `npm_config_engine_strict=false npm ci` | bypassed | 0 |

It also explains a detail the old story never fit: the #113 frontend audit (2026-07-03, npm 11.6.2) had to run `npm_config_engine_strict=false npm **ci**`. If the guard were write-path-only, the bypass would not have been needed there at all. That session lost time to an empty `node_modules` — the failure is quiet, so `npm run lint` / `npm run typecheck` stop being a real gate instead of erroring.

## Changes

- **`frontend/.npmrc`** — replace the false mechanism block with what `engine-strict` actually does (both paths), the probe result and date, and the two real unblocks.
- **`frontend/README.md`** — new "If your npm is 11 or newer" section under Run, so a contributor hits the answer before the empty `node_modules`.

Both keep the never-commit-an-off-pin-lockfile rule prominent — an off-pin lockfile is what breaks the Cloudflare Workers deploy, and the whole reason the pin exists.

`.github/workflows/ci.yml`'s comment (":84 — `.npmrc` engine-strict=true enforces it during `npm ci`") was already correct and is unchanged.

Comments and docs only — no code, no lockfile, no dependency changes.

## Follow-up

The live Canopy doc `sapling-frontend-local-dev` carries the same false claim; a corrected **v2 is staged** and needs promotion.

### DIFF
--- a/frontend/.npmrc
+++ b/frontend/.npmrc
@@ -7,12 +7,28 @@
 #
 # What `engine-strict=true` actually does:
 #
-# It makes LOCAL `npm install` fail fast against the package.json
-# `engines` constraint when the running npm is out of range, stopping
-# the bad lockfile commit at its source. It does NOT change `npm ci`'s
-# behavior — `npm ci` ignores engine-strict in npm >=7. So this is a
-# guard on the WRITE path (regeneration), not the READ path
-# (consumption).
+# It makes npm fail fast against the package.json `engines` constraint
+# whenever the running npm is out of range. It blocks BOTH paths:
+#
+#   npm install — the WRITE path this guard exists for: it stops a
+#                 lockfile regenerated on the wrong npm at its source.
+#   npm ci      — the READ path, too. Verified 2026-07-31 on npm 12.0.1
+#                 against this `<11` pin: exit 1, code EBADENGINE.
+#
+# An earlier version of this comment claimed `npm ci` ignores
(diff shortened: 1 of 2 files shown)


AndresL230:
Verified independently before merging:

- `frontend/package.json` does pin `"npm": ">=10.9.0 <11"`.
- This machine runs npm **12.0.1** — out of range, same class as the probe environment.
- `.github/workflows/ci.yml:84` already said *"`.npmrc` engine-strict=true enforces it during `npm ci`"*, and `:92` pins `npm@10.9.2` with `--engine-strict=false` before installing. So the old `.npmrc` comment contradicted the repo's own CI config — the correction resolves a real internal inconsistency, not just a doc nit.

CI green across all lanes including the Workers build. Comments and docs only — no code, no lockfile, no dependency changes. Merging.

Issue #293, opened 2026-06-30
frontend: flatten components/ directory + co-locate marketing components


AndresL230:
~50 files in one `components/` folder. Group by responsibility; co-locate the pre-auth components (`HowItWorks`, `SignInModal`, landing sections) under `components/marketing/`. Also consider renaming the `.landing-page` scope class now that it's layout-hosted. Deferred from token-unification Phase 1 (PR #286). Tracker: docs/superpowers/followups/2026-06-30-token-unification-followups.md · Audit: docs/frontend-rhythm-audit.md

AndresL230:
## Proposal — flatten `components/` + co-locate marketing (B11, verified @ d1d724f)

**Not implemented — proposal only.**

### Current state

107 files total: **54 at the top level** of `components/` (45 non-test + 9 tests) plus 6 subdirs (`Gradebook/`, `screens/`, `screens/Gradebook/`, `flashcards/`, `flashcards/tabs/`, `ui/`).

Two findings that change the shape of this work:

1. **`Gradebook` exists at two levels** — `components/Gradebook/` (parts: AssignmentList, CourseCard, modals, `curveUtils.ts`…) and `components/screens/Gradebook/` (routes: Landing, Course). Any flatten has to resolve that split, and #139/#468 just added files to *both*.
2. **The landing sections are not components.** They're inline JSX inside `(public)/page.tsx:469-981`. So "co-locate landing sections under `components/marketing/`" is blocked on **extracting them first** — that extraction naturally belongs to #344 phase 1, not here.

`.landing-page` is applied exactly once (`page.tsx:469`); `.public-surface` at `(public)/layout.tsx:9`. The landing carries both classes today.

### Proposal (three separable commits)

1. Group the top level by responsibility (`chat/`, `graph/`, `study/`, `social/`, `gradebook/` — merging the two Gradebook levels with parts under `gradebook/` and routes staying in `screens/` — `ui/` unchanged).
2. Move `HowItWorks.tsx` and `SignInModal.tsx` into `marketing/` now; **defer landing-section co-location to #344**.
3. Optional: merge/rename the two-class marketing scope (`.public-surface` + `.landing-page`) now that tokens are layout-hosted.

### Scope and risk

- Pure moves + import rewrites; gates are tsc + vitest + lint. Zero behavior change.
- **The real risk is timing, not correctness:** this touches ~50 files and will conflict with every open frontend branch. It should land when the frontend queue is empty — not alongside other work.

[...]

Issue #477, opened 2026-07-30
e2e flake: tutor persistence assert can read messages before the turn's rows land (tutor.spec.ts:83)


AndresL230:
Hit once during PR #475's pre-merge cycles (2026-07-30), on a run whose only delta from a fully-green prior cycle was a one-line selector comment in an unrelated spec — so not a product regression from that PR.

**Symptom:** `tutor.spec.ts:44` "tutor turn renders a reply and persists encrypted to messages" failed at line 83: `expect(rows).toHaveLength(SEEDED_MESSAGE_COUNT + 2)` — expected 6, received exactly the 4 seeded rows. The UI reply had rendered (earlier assertions passed), but the one-shot `queryRaw` read the `messages` table before the turn's user+assistant rows were committed. Passed on the immediately preceding and following full cycles (zero-flake otherwise).

**Rate:** 1 failure in ~6 full local cycles today (each cycle runs the spec once) plus every recent green main run — rare.

**Root-cause candidates:** the persistence contract writes both rows before the SSE `done` event; if the client renders the final bubble off the last token rather than `done`, a fast function-mode turn can expose a window where the spec's DB read beats the commit. Either await `done` explicitly before the DB assert, or make line 83 an `expect.poll` (the pattern gradebook.spec.ts uses for its mutation assert).

**Evidence preserved** (screenshot / video / trace.zip / error-context.md) at:
`/tmp/claude-1000/-home-andresl-Projects-sapling/428d9f6f-5046-455f-a4b7-b17d5bc2e619/scratchpad/tutor-flake-evidence/` — copy out before that tmpdir is cleaned if needed.

Per the #388 zero-flake policy this should be fixed at the root (poll or done-gate), not retried.

AndresL230:
Fixed in #500 (`749c76e`).

## Root cause

Not a product bug, and worth stating explicitly since the alternative reading pointed at one. The server contract is sound: `stream_agent_turn` calls `on_complete` — which is `_persist`, a synchronous `save_message` — at `services/chat_stream.py:372`, and only *then* yields `done` at `:393`. So the rows really are committed before the turn ends.

[...]

PR #494, opened 2026-07-31
refactor(frontend): group components/ by responsibility (#293)

AndresL230:
The top level of `components/` held **58 files** with no organising principle. Grouped per the issue's own proposal — pure moves + import rewrites, zero behaviour change.

| group | files |
|---|---|
| `marketing/` | HowItWorks, SignInModal, HeroCard (+test) |
| `graph/` | KnowledgeGraph, 2D, 3D (+3 tests) |
| `chat/` | ChatPanel, MarkdownChat, MermaidBlock, FunctionPlot, AIDisclaimerChip, ModelToggle, SharedContextToggle, SessionSummary |

**Top level: 58 → 38.** `ui/`, `flashcards/` and `screens/` unchanged. Moves used `git mv`, so history follows the files.

## Timing

The proposal flagged that the real risk here is timing, not correctness: *"this touches ~50 files and will conflict with every open frontend branch. It should land when the frontend queue is empty."* It is — 0 open PRs when this branched. That's why this went first.

## Two deviations from the proposal, both deliberate

**`study/` and `social/` were not created.** Neither had any top-level members — those screens already live in `screens/` — so the groups would have been empty directories.

**`Gradebook/` was not renamed to `gradebook/`.** The proposal wanted "the two Gradebook levels merged", but they're already split cleanly by responsibility: `components/Gradebook/` holds the parts, `components/screens/Gradebook/` holds the routes. The only remaining delta is letter case, and a case-only directory rename is a real cross-platform hazard for the macOS contributors on this repo. Not worth it for a cosmetic gain — happy to do it if you disagree.

## Four things that had to move with the files

Each of these would have failed silently or confusingly:

- **`frontend/eslint.config.mjs`** — the testid-enforcement `files` array is path-keyed. Stale entries would have silently stopped enforcing the rule on the exact files it exists to cover. That's a *quiet* failure, which is the dangerous kind.
- **`frontend/eslint-suppressions.json`** — also path-keyed. This one failed loudly: 22 lint errors, because the suppressions no longer matched any file.
- **`docs/frontend-testids.md`** — the surface table names owning files; this is the other half of the testid convention.
- **`HeroCard.test.tsx`** resolved `src/` as `__dirname/'..'`, which silently became `components/` once the file moved a level deeper. Now `'../..'`.

Also refreshed live path references in `SECURITY.md` and the graph journey's header comment.

Historical records (`docs/frontend-audit/`, `docs/decisions/`, `docs/superpowers/plans/`) are **deliberately left** pointing at the old paths — they're dated snapshots of what was true when written, not live indexes. Rewriting them would falsify the record.

## Gates

part of #293


AndresL230:
### Review found two silent bugs the green gates could not

Multi-agent review over the move. `tsc`, `lint`, `vitest` and the full e2e lane were **all green on the broken state** — the exact hazard of a move-only refactor.

Two `vi.mock()` specifiers still pointed at pre-move paths:

```
Dashboard.test.tsx       vi.mock("../KnowledgeGraph")  -> "../graph/KnowledgeGraph"
chat/ChatPanel.test.tsx  vi.mock("./Icon")             -> "../Icon"
```

vitest doesn't error on a mock specifier that resolves to nothing — it just declines to intercept. So the *real* components rendered instead of the stubs and the tests stayed green either way. `tsc` can't see these because a `vi.mock` path is a string literal, not an import.

Fixed in `b72640f`, and I swept every relative `vi.mock` specifier in `src/` against the filesystem: those two were the only unresolvable ones, and there are now zero.

Review also verified, all clean: every entry in `eslint.config.mjs`'s testid `files` array and all 28 `eslint-suppressions.json` keys resolve to existing files (a stale entry there stops enforcing silently); `docs/frontend-testids.md` surface table updated; internal relative imports in all moved files correctly re-depthed; no other test resolves paths via `__dirname`; `next.config.ts` / `tsconfig` / `vitest.config` / `playwright.config` / CI workflows carry no old paths.

### E2E

```
35 passed (1.3m)
```

Worth recording: the **first** run of this cycle failed with 34/35 — `study-room.spec.ts` read a different room's messages. My diff has zero overlap with the social/realtime path (verified: no social, realtime or room file appears in it), and the re-run at the same sha passed 35/35. That's a two-context realtime race, the same family as open issue #477. Flagging it there rather than papering over it here.

Gates: `tsc --noEmit` clean · `npm run lint` 0 errors · `npx vitest run` 58 files / 415 tests.

Issue #344, opened 2026-07-15
[p1] frontend PRIORITY - Landing page redesign


Darkest-Teddy:
The frontend design from Claude design for the landing page should be transferred to staging.

However, there are still a couple of notable tasks that need to be completed:

- The graph demo needs to be redesigned so it can be fully launched interactive with text fading out as the user explores how course and concept nodes work
- The overall of all the features need to be redesigned to reflect the actual pages on the Sapling platform, not be generic interfaces
- General cleaning of the UI, remove all unnecessary, redudant components such as the left side of the page indication, text explaining to drag, scroll, etc, plus buttons for each feature preview, and much more
- The site still needs a refreshed redesign to be more visually pleasing, 3d helix animations, and more to impress the audience
- The website still feels generic without unique assets to express Sapling as a whole

AndresL230:
## Proposal — landing redesign (B11, verified against `origin/main` @ d1d724f)

**Nothing here is implemented.** This is a scoped plan for you to green-light, amend, or reject. Current-state findings first, because two items in the issue body are already stale.

### What the code actually says today

- **The "graph demo" is not interactive at all.** `(public)/page.tsx:530` is a `<canvas>` driven by the RAF at `:130-268`: 220 procedural + 6 cluster nodes, fixed rotation (`rotAngle += 0.0008`), and the mouse only feeds a global parallax skew. No hit-testing, no labels, no click/hover handlers, no course/concept semantics. `pointer-events-auto` is set but nothing consumes pointer events. So "redesign it to be fully interactive with text fading as the user explores" is a **build**, not a tweak.
- **The feature previews are text-only.** Two distinct things read as generic: the hero floating glass cards (`:532-561`, fake mini-UI chips that match no real screen) and the 6-row editorial catalog (`:635-683`, lucide icons + prose, zero screenshots of real Sapling pages).
- **Already done, contrary to the issue body:** the per-feature preview buttons are gone (`.landing-spotlight-card` is orphaned CSS with zero markup consumers).
- **Stale detail:** the page indicator is on the **right**, not the left — `HowItWorks.tsx:497-526`, `right: clamp(1rem, 2vw, 2rem)`.
- **Still present:** the hint text at `:601` ("SEE WHAT'S INSIDE"), `:629` ("Scroll the catalog"), `:688` ("— end of catalog").
[...]

PR #496, opened 2026-07-31
refactor(landing): delete unreachable effects and orphaned CSS (#344 phase 1)

AndresL230:
Phase 1 of the landing redesign: **dead-code removal only, no visual change.** Phases 2–4 need your and Jack's taste input and aren't touched.

## Removed — each verified unreachable, not assumed

| what | why it was dead |
|---|---|
| `startCounters` (18 lines) | its only caller was the `landing-stat-fade-up` branch of the fade-up observer, and that class appears in **no markup anywhere** |
| the spotlight mouse-follow effect (42 lines) | opens with `if (cards.length === 0) return`, and `.landing-spotlight-card` is applied in no markup — it early-returned on every mount |
| 16 orphaned CSS rules | `.landing-btn-shimmer`, `.landing-spotlight-card`, `.landing-icon-container` — zero consumers outside `globals.css`, in **both** the unscoped and scoped copies |

The observer itself stays: `.landing-fade-up` is live at three call sites, so only the stat branch and the dead selector go.

**Net: 117 deleted, 1 added.**

## One thing in the diff that looks alarming and isn't

The suppressions baseline swaps `react-hooks/immutability` → `react-hooks/set-state-in-effect` on `(public)/page.tsx`. **That is not a new violation.**

The deleted code mutated DOM styles and drove `requestAnimationFrame` loops, which made React Compiler bail out on the whole component — masking a pre-existing `setState` inside the mount-time URL-error effect at `:110`. Removing the dead code made the file analyzable again, and the latent violation surfaced.

Verified by swapping main's `page.tsx` into this branch: with main's file the full lint is **0 errors**, with mine it's **1**. Same suppressions file both times.

The pattern itself predates this PR and can't move to lazy state init — the initializer would also run during SSR, where `window` is undefined. Suppression count for the file is unchanged: one out, one in.

## Deliberately not done

The wider de-dup of the unscoped vs `.landing-page`-scoped rule blocks. I verified it's safe *in principle*:

- all 21 unscoped `landing-*` selectors have scoped twins
- the two apparent "properties only the unscoped rule defines" are comment-parsing artifacts — the scoped rules do define them, identically
- no `.public-surface` content page (about/careers/privacy/terms) uses a `landing-*` class

But the two blocks are **interleaved with 11 unrelated rules** (`.glass-panel`, `.liquid-glass-*`, `.glass-input`, `:root`, toast keyframes…), so it can't be a range delete — and this phase's own gate is *"screenshot diff shows no change"*, which I can't run here: Turbopack refuses the symlinked `node_modules` in a worktree. Left for a follow-up that can actually be screenshotted, rather than shipping ~200 lines of CSS deletion on the acquisition surface unverified.

Also untouched: the hint text (`"SEE WHAT'S INSIDE"`, `"Scroll the catalog"`, `"— end of catalog"`) and the page indicator. Both are **visible** changes, so they belong with your phase-2 sign-off, not in a zero-visual-change cleanup.

## Gates

part of #344

## Summary by CodeRabbit

* **Refinements**
  * Simplified landing-page scroll animations to focus on supported fade-up elements.
  * Removed animated statistic counters from the landing page.
  * Streamlined landing-page visual effects, including button shimmer, card glow, and feature-icon hover styling.
* **Maintenance**
  * Updated linting configuration to reflect the current React hook usage.

### DIFF
--- a/frontend/eslint-suppressions.json
+++ b/frontend/eslint-suppressions.json
@@ -6,7 +6,7 @@
     "prefer-const": {
       "count": 2
     },
-    "react-hooks/immutability": {
+    "react-hooks/set-state-in-effect": {
       "count": 1
     }
   },
--- a/frontend/src/app/(public)/page.tsx
+++ b/frontend/src/app/(public)/page.tsx
@@ -404,77 +404,16 @@ export default function LandingPage() {
       entries.forEach(entry => {
         if (entry.isIntersecting) {
           entry.target.classList.remove('opacity-0', 'translate-y-[30px]');
-          if (entry.target.classList.contains('landing-stat-fade-up')) startCounters(entry.target as HTMLElement);
           obs.unobserve(entry.target);
         }
       });
     }, { threshold: 0.2 });
-    document.querySelectorAll('.landing-fade-up, .landing-stat-fade-up').forEach(el => {
+    document.querySelectorAll('.landing-fade-up').forEach(el => {
       el.classList.add('opacity-0', 'translate-y-[30px]', 'transition-all', 'duration-700', 'ease-out');
       obs.observe(el);
     });
     return () => obs.disconnect();
   }, []);
-
-  // Spotlight card mouse-follow
-  useEffect(() => {
-    const cards = Array.from(document.querySelectorAll<HTMLElement>('.landing-spotlight-card'));
-    if (cards.length === 0) return;
-
-    // `getBoundingClientRect` per mousemove forces a layout flush on every
-    // pointer sample. Measure on enter instead and hold the rect until
-    // something can actually have moved the card.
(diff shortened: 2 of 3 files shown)


AndresL230:
### Local e2e cycle — green

Run at `cbf88c8` inside one `flock`, `SAPLING_MODEL_MODE=function`:

```
35 passed (1.3m)
0 finding(s), 0 suppressed (allowlisted)
```

Disclosure: an earlier run of this cycle was **discarded, not reported**. I switched the primary checkout to `main` while that run held the stack lock, so its oracles executed newer code than the build under test. It passed, but a contaminated gate isn't a gate — this result is from a clean re-run with the tree untouched throughout.

Gates: `tsc --noEmit` clean · `npm run lint` 0 errors · `npx vitest run` 58 files / 415 tests.

Issue #289, opened 2026-06-30
frontend: re-home onboarding into app layout/spacing/type


AndresL230:
[P1-C] `components/screens/Onboarding.tsx` is color-correct but structurally orphaned (centered card in a radial void, hard-coded px, bespoke type scale). Adopt `--pad-*` + the shell type scale; render inside the app's left-aligned frame. Deferred from token-unification Phase 1 (PR #286). Tracker: docs/superpowers/followups/2026-06-30-token-unification-followups.md · Audit: docs/frontend-rhythm-audit.md

AndresL230:
## Proposal — re-home onboarding (B11, verified @ d1d724f)

**Not implemented — proposal only.**

### Current state

`components/screens/Onboarding.tsx` (548 lines) is **color-correct but structurally orphaned**, exactly as the issue says. Its route (`app/onboarding/page.tsx`) renders it bare, **outside `(shell)`** — no ShellFrame, no nav, no `<main>` padding. Specifically:

- centered card in a radial void: `:164-172` (`minHeight: 100vh`, centered flex, `radial-gradient(...)`, `padding: 40px 20px`)
- hard-coded px throughout, none density-aware: `:171`, `:179` (`40px 36px`), `:186`, `:191`, `:245`, `:289`, `:469`, `:527`
- a bespoke type ramp: `32px` on the welcome step, `26px` on every other (`:268, :307, :329, :351, :445, :513`) — matching neither TopBar's 30 nor Dashboard's 42
- a third bespoke step-dot implementation (`:191-204`), separate from HowItWorks' `StepIndicator`

It *is* token-disciplined on color (uses `.card`, `.btn`, `.chip`, `var(--accent)`, zero hard-coded hex). **The debt is layout, spacing, and type — not color.**

### Proposal

Keep the flow and steps exactly as they are. Change only the frame: adopt `--pad-*` (`globals.css:136-140`, density-aware) and the shell type convention (`.h-serif` + TopBar-scale sizes), render inside the app's left-aligned frame with shell-like chrome but no nav (the user is pre-approval), and replace the radial void with the app backdrop.

**One thing that needs your input:** the repo has no `--fs-*` type-scale tokens — the "shell type scale" is a convention (`.h-serif` + a per-site inline size), not a token set. So this either (a) follows the convention with inline sizes, or (b) introduces the scale tokens as part of the work. (b) is better long-term and bigger.

### Scope and risk

- ~1 day for (a); ~2 with (b).
[...]

PR #497, opened 2026-07-31
feat(ui): re-home onboarding onto app spacing/type tokens (#289)

AndresL230:
Onboarding was **color-correct but structurally orphaned**: 12 hardcoded px values, a bespoke type ramp (32 on welcome, 26 everywhere else — matching neither TopBar's 30 nor Dashboard's 42), and a centred card floating in its own radial void. The flow and steps are untouched; only the frame changed.

## The `--fs-*` scale — option (b), as you chose

**Derived, not invented.** Every step is a size the app already uses, counted across `screens/` and `components/`:

| | |
|---|---|
| `--fs-md: 13px` | the app's true body size — **107 call sites** |
| `--fs-sm/xs/2xs` | 12 / 11 / 10 — chrome and micro-labels (215 uses combined) |
| `--fs-base/lg` | 14 / 16 — body copy |
| `--fs-xl/2xl/3xl` | 18 / 22 / 26 — headings |
| `--fs-4xl: 30px` | screen title — matches TopBar |
| `--fs-5xl: 32px` | Dashboard display numeral |

**Deliberately not density-aware**, unlike `--pad-*`. Retuning the whole type ramp with the density preference changes how much text fits on every screen — that's a product decision, not a refactor. Additive to add later; doing it now would hide a behaviour change inside a token introduction.

**Onboarding is the scale's first and only consumer.** The rest of the app keeps its inline sizes until each screen is converted deliberately — a token set with zero consumers is exactly the state #288 just finished cleaning up, and I'd rather not recreate it.

## The rest of the frame

- Every hardcoded padding → `--pad-*`, so onboarding finally responds to the density preference at all.
- The bespoke radial void → the app's own `--bg`.
- `minHeight: 100vh` → `100dvh` — the same iOS Safari lesson ShellFrame learned in #331, which onboarding never got because it renders outside `(shell)`.

**One visible change worth calling out:** step headings move **26 → 30** to match TopBar. That's the alignment the issue asks for, not a side effect.

## The journey that didn't exist

There was **no e2e journey over `/onboarding` at all** — notable, since it's the first screen a newly-approved student sees and it renders bare, with no ShellFrame, nav or `<main>` padding to inherit. Nothing in the suite would have noticed it breaking.

Added one. It signs in as `USER_NEW` (the only seeded user with `onboarding_completed=False`, so the only one the funnel renders for) and asserts the card actually **occupies the screen with resolved padding** — a mistyped `var()` computes to `0px` and silently collapses the box, which no unit test would catch and which is the specific risk this PR's conversion introduces.

It's a smoke-plus-layout journey rather than a five-step walkthrough on purpose: the field validation is unit-tested, and a long scripted click-path over a funnel that's still being redesigned would break on every copy tweak.

## Testids — both halves

Driving onboarding made it an E2E surface, so `Onboarding.tsx` joins the enforcement list in `eslint.config.mjs`. That immediately flagged **7 untagged controls** (the rule doing its job), so all ten are now tagged and registered in `docs/frontend-testids.md` — surface table and inventory.

## Not done

The proposal also suggested rendering onboarding in the app's **left-aligned frame with shell-like chrome**. I've left the centred card: moving the first-run funnel from centred-card to left-aligned-shell changes its whole character, and that's a taste call rather than debt cleanup. Happy to do it if you want it.

## Gates

part of #289

### DIFF
--- a/docs/frontend-testids.md
+++ b/docs/frontend-testids.md
@@ -114,10 +114,15 @@ route:
 | `onboarding-continue` | primary advance button ("Continue" / "Enter Sapling →") |
 | `onboarding-back` | "Back" (absent on step 0) |
 | `onboarding-close` | "×" close/exit control |
-| `onboarding-input` | the shared `TextInput` used by every step |
-| `onboarding-chip-add` / `onboarding-chip-remove` | majors/minors chip editor |
-| `onboarding-course-result` / `onboarding-course-remove` | course search result and selected-course chip |
-| `onboarding-learning-style` | learning-style radio option |
+| `onboarding-input` | the shared `TextInput` default; every simultaneous instance overrides it (below) |
+| `onboarding-first-name` / `onboarding-last-name` | the two name fields — they render side by side |
+| `onboarding-school` | school field (disabled; BU-only today) |
+| `onboarding-chip-input-${field}` / `onboarding-chip-add-${field}` | majors/minors editor — `field` is `majors` or `minors`, since both render at once |
+| `onboarding-chip-remove-${field}-${value}` | one chip's remove control |
+| `onboarding-course-search` | course search field |
+| `onboarding-course-result-${course.id}` | one course search result |
+| `onboarding-course-remove-${course_code}` | one selected-course chip's remove control |
+| `onboarding-learning-style-${style.id}` | one learning-style radio option (5 render at once) |
 
 ### `upload-modal`
 
--- a/docs/frontend-testids.md
+++ b/docs/frontend-testids.md
@@ -68,6 +68,7 @@ renders the element.
 | --- | --- | --- |
 | Sign-in | `signin` | `frontend/src/components/marketing/SignInModal.tsx` (+ the trigger in `src/app/(public)/page.tsx`) |
 | Approval gate | `pending` | `frontend/src/app/pending/page.tsx` |
+| Onboarding | `onboarding` | `frontend/src/components/screens/Onboarding.tsx` (the first-run funnel, rendered bare outside `(shell)`) |
 | Upload modal | `upload-modal` | `frontend/src/components/DocumentUploadModal.tsx` |
 | Tutor | `tutor` | `frontend/src/components/chat/ChatPanel.tsx` (rendered by `screens/Learn.tsx`; + the session-resume rows in `src/components/screens/Learn.tsx` itself) |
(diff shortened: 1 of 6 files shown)


AndresL230:
### Review found four things — all fixed in `25ab839`

**Six of the ten testids were on elements that render simultaneously.** Two `TextInput`s side by side in `StepName`, two `TagInput`s (majors + minors) each with their own add button and chips, five learning-style radios, N course results — all sharing one testid, so `getByTestId` couldn't disambiguate them. `docs/frontend-testids.md` already has a "Repeated / list items" rule requiring a stable suffix, and the codebase already had the precedent (`upload-modal-course-result-${c.id}`). I missed both. Now suffixed by id/field/value, and `TagInput` takes a `field` prop for the same reason.

Worth noting the eslint rule only checks a `data-testid` is *present*, not that it's unique — so lint was green on ambiguous ids.

**Two paddings collapsed a two-value shorthand into one.** `"40px 36px"` → a single `--pad-xl` and `"14px 12px"` → a single `--pad-md` threw away deliberate vertical/horizontal asymmetry, on the card every step renders inside. Both are two-value again.

**A comment asserted something I hadn't verified.** The `USER_NEW` docstring claimed "approved so the middleware lets them past the pending gate" — but `/onboarding` isn't in `middleware.ts`'s `PROTECTED` list or its matcher, and `Onboarding.tsx` only redirects when unauthenticated. The route renders for any signed-in user. `USER_NEW` is right because it's the realistic first-run state, not because the route gates on it. Corrected in both the constant and the spec docstring.

### Value shifts from adopting a quantised scale

Listing these rather than leaving them to be found. The `--pad-*` scale has no 36/40 step, so:

| was | now | delta |
|---|---|---|
| card `40px 36px` | `--pad-xl --pad-lg` = `32px 22px` | −8 / −14 |
| page `40px 20px` | `--pad-xl --pad-lg` = `32px 22px` | −8 / +2 |
| inputs `10px 12px` | `--pad-sm --pad-md` = `10px 16px` | 0 / +4 |
| loading rows `14px 12px` | `--pad-md --pad-sm` = `16px 10px` | +2 / −2 |

Type is exact except the intentional 26 → 30 on step headings.

### E2E

```
36 passed (1.3m)     # 35 existing + the new onboarding journey
0 finding(s)
```

Gates: `tsc` clean · `lint` 0 errors · `vitest` 58 files / 415 tests.

Issue #315, opened 2026-07-03
Store image dimensions on room chat attachments to reserve space (lazy-load pop-in follow-up)


AndresL230:
Follow-up from #312 (part 1 of #111), which added `loading="lazy"` to room-chat image attachments.

## Problem

Chat attachments render with no reserved height (`Social.tsx` RoomChat: `maxWidth: 260`, no width/height/aspect-ratio). With lazy loading, scrolling **up** through history fetches each image as it approaches the viewport and it expands from ~0 height, shifting the transcript mid-scroll.

- Chrome/Firefox: mostly absorbed by native scroll anchoring.
- **Safari (desktop + iOS): no scroll anchoring — the viewport visibly jumps on every image load.**
- The `loadEarlier` scrollTop compensation (`Social.tsx:398-408`) measures `scrollHeight` before prepended images load, so late expansions also invalidate it (this part was already fragile pre-lazy).

## Why it can't be fixed in the frontend alone

`room_messages` stores only `image_url` (see `backend/routes/social.py:293` select list); there are no dimensions to reserve an aspect-ratio box with.

## Proposed fix

1. At upload time (Social.tsx upload path), read the image's natural width/height before/while uploading and send them with the message.
2. Add `image_width`/`image_height` columns to `room_messages` (append-only migration) and include them in the select.
3. Render attachments in a fixed `aspect-ratio` placeholder box so lazy loading causes zero layout shift.

Same mechanism exists in MarkdownChat's img override, but tutor output effectively never contains image URLs (prompt only allows embedding real URLs it's given), so no action needed there.

AndresL230:
Landed via #498 (`4236c32`). **Leaving open for your UI verification.**

Implemented exactly the three-part fix proposed here: measure at upload, store on `room_messages`, render into a reserved box.

- **`0040`** adds nullable `image_width`/`image_height`
- **`Social.tsx`** measures the picked `File` via an object URL **before** upload, so dimensions travel with the message that creates it — history renders reserved from the first paint rather than after a round trip
- attachments render into an `aspect-ratio` box

This also closes the gap **#111 deliberately left open**: that PR added `decoding="async"` here but declined to invent dimensions for natural-aspect user uploads, precisely because none were stored.

## One thing the review caught that's worth knowing

[...]

PR #498, opened 2026-07-31
feat(social): reserve space for chat image attachments (#315)

AndresL230:
Room-chat attachments rendered into a **zero-height box** until the image loaded. Combined with the `loading="lazy"` #312 added, scrolling *up* through history expanded each image as it neared the viewport and shifted the transcript mid-scroll.

- Chrome/Firefox: mostly absorbed by native scroll anchoring
- **Safari (desktop + iOS): no scroll anchoring — the viewport visibly jumps on every image load**

It also invalidated the `loadEarlier` scrollTop compensation, which measures `scrollHeight` before the prepended images have loaded.

This is also the gap **#111 deliberately left open** — that PR added `decoding="async"` to these images but declined to invent dimensions for natural-aspect user uploads, precisely because there were none stored. Now there are.

## Why it needed the backend

`room_messages` stored only `image_url`. Nothing to reserve a box with, so it couldn't be fixed in the frontend alone.

| layer | change |
|---|---|
| `0040` | `image_width` / `image_height` on `room_messages`, nullable |
| `models` | same two fields on `SendMessageBody`, optional |
| `social.py` | both columns in the select list **and** the insert |
| `api.ts` | `sendRoomMessage` takes an optional `imageSize` |
| `Social.tsx` | measures the file before upload; renders into an aspect-ratio box |

The measurement runs on the picked `File` via an object URL **before** upload, so the dimensions travel with the message that creates it — history then renders reserved from the first paint rather than after a round trip.

## Degrades quietly at every layer

That's deliberate, because this is a layout hint and must never be able to block a message:

- columns are **nullable and unbackfilled** — every pre-existing message keeps `NULL`
- `readImageSize` resolves `undefined` on a non-image or a decode failure rather than throwing
- the renderer applies `width`/`height`/`aspect-ratio` **only when both values are present**

Anything without dimensions renders exactly as it does today. Backfilling would mean fetching every historical image to measure it — not worth it for a layout hint.

## Gates

- backend `pytest` 1526 passed, 32 skipped · `ruff check` clean
- **From-empty replay** (not just a normal cycle — this adds a migration, and `e2e-up` runs against a DB that already has the schema): in a comment below

part of #315

### DIFF
--- a/backend/db/migrations/0040_room_message_image_dimensions.sql
+++ b/backend/db/migrations/0040_room_message_image_dimensions.sql
@@ -1,0 +1,22 @@
+-- 0040: store intrinsic dimensions for room-chat image attachments (#315).
+--
+-- `room_messages` carries only `image_url`, so the client has nothing to
+-- reserve space with and renders attachments into a zero-height box until the
+-- image loads. With `loading="lazy"` (added by #312) that means scrolling UP
+-- through history expands each image as it nears the viewport and shifts the
+-- transcript mid-scroll. Chrome and Firefox mostly absorb it via scroll
+-- anchoring; Safari has none, so the viewport visibly jumps on every load.
+--
+-- It also breaks the `loadEarlier` scrollTop compensation, which measures
+-- scrollHeight before the prepended images have loaded.
+--
+-- Nullable and unpopulated by design: every message written before this
+-- migration keeps NULL, and the client falls back to its current behaviour
+-- for those rows. Backfilling would mean fetching every historical image to
+-- measure it, which is not worth it for a layout hint.
+--
+-- IF NOT EXISTS so this is a no-op wherever the columns already exist.
+
+ALTER TABLE room_messages
+    ADD COLUMN IF NOT EXISTS image_width  INTEGER,
+    ADD COLUMN IF NOT EXISTS image_height INTEGER;
--- a/backend/db/migrations/0040_room_message_image_dimensions.sql
+++ b/backend/db/migrations/0040_room_message_image_dimensions.sql
@@ -20,3 +20,25 @@
 ALTER TABLE room_messages
     ADD COLUMN IF NOT EXISTS image_width  INTEGER,
     ADD COLUMN IF NOT EXISTS image_height INTEGER;
+
+-- Bounds at the database too, not only in the request model. These values are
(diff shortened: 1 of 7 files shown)


AndresL230:
### Review found a real escalation in blast radius — fixed in `7c960ed`

`image_width`/`image_height` arrive from the client and the transcript renders them **directly as an aspect-ratio**, with no bounds anywhere. Any room member could POST:

```json
{ "image_width": 1, "image_height": 2000000000 }
```

Comfortably inside Postgres `INTEGER` range, so it inserts cleanly — and every viewer of that room gets a ~2-billion-pixel-tall bubble, with no edit path to recover short of a DB fix. `width=1` never trips `maxWidth: 260`, so nothing capped it.

That's an escalation my own change introduced: before this PR the worst a bogus attachment could do was render a broken-image icon; it could not distort layout for other people.

Bounded at three layers, each catching a different failure:

| layer | fix | catches |
|---|---|---|
| model | `Field(gt=0, le=20000)` | the absurd pair, before it reaches the DB |
| migration | `CHECK` constraint, `NOT VALID` | anything that bypasses the API |
| render | `objectFit: contain` + `maxHeight` | dimensions that are merely *wrong* — they letterbox instead of stretching |

`le=20000` is comfortably past any real image (8K is 7680). The `Field(gt=0)` shape matches what `models/__init__.py` already uses for client-supplied numerics, and the `CHECK` matches the convention `0021` established. `NOT VALID` so it governs new writes without scanning existing rows — every pre-existing row is NULL in both columns, and NULL passes a CHECK anyway.

**Edited `0040` rather than adding `0041`** because it isn't applied anywhere: unmerged, and the only database that has seen it is the local throwaway the from-empty replay rebuilds each run.

Added three tests — nothing covered these fields at all before.

### From-empty replay (re-run after the hardening)

```
RESET_EXIT=0        # supabase db reset + every migration, incl. the new CHECK
JOURNEYS            36 passed
ORACLES             0 finding(s)
```

Backend `pytest` 1529 passed, 32 skipped · `ruff` clean · frontend `tsc`/`lint` clean · `vitest` 415.

Also verified clean by review: the realtime path re-fetches through the decrypting REST endpoint rather than trusting `payload.new`, and the optimistic append uses the POST response (`Prefer: return=representation`) — so both pick up the new columns without further change.

Issue #476, opened 2026-07-30
Study: opening a guide from the Recent guides rail clears the exam selection, leaving Regenerate permanently disabled


AndresL230:
Found while testing #475 (the #141 reframe), pre-existing on main.

**Repro:** on /study, open a study guide from the "Recent guides" rail (not via an exam selection). The exam selection state is cleared, and the Regenerate button stays permanently disabled on that path — there is no way to regenerate the guide you're looking at without re-selecting the exam first.

**Where:** `frontend/src/components/screens/Study.tsx` — the recent-guides open path clears the selected exam; Regenerate is gated on it.

Small UX fix; noted in PR #475's description. Not addressed there to keep that PR scoped to semester threading.

AndresL230:
Fixed in #499 (part of #476). Leaving this open for your UI verification.

## The issue body misattributes the cause

> "the recent-guides open path clears the selected exam"

It doesn't — `openRecent` sets **both** ids. The clearing came from the
**courseId-keyed exams effect**, whose first statement was an unconditional
`setExamId("")`. That reset can't tell *"the user switched course"* (selection
now invalid) from *"we just opened a specific guide"* (selection deliberate and
valid); it only sees that `courseId` differs from last render, which
`openRecent` also causes.

Both effects run in the same commit, so the loader still saw the intact pair
and the guide loaded — only the *next* render lost the exam. That's why the
symptom was "a guide on screen above a dead Regenerate button" rather than
"nothing opens".

**It needs a course change to reproduce.** A rail entry whose course was
already selected always worked. That's pinned as a control test — it passed
before the fix and still passes, which is what proves the mechanism is the
reset rather than the open path.

## What changed

The reset now happens at the two events that mean it: the course picker's
`onChange`, and a term switch. The term case adjusts state during render (the
`StudyModePanel` pattern already in the file) instead of in an effect, because
an effect-time reset lands a render late — the loader would commit one read of
the old exam under the *new* term first. That was the same defect's second
trigger; it reproduced on main and now has a test.

Making Regenerate reachable also exposed a term hazard: it sent the **active
selector's** term while a recent entry opens under its **own** term (#475 F1).
Latent before, because the button was dead on the one path where those differ.
Regenerate now replays the term the displayed guide was loaded with.

## What to check when you verify

1. `/study`, click any entry in "Recent guides" → the exam picker should show
   that exam and **Regenerate should be clickable**.
2. Switch course in the picker → exam selection clears (unchanged behavior).
[...]

PR #475, opened 2026-07-30
feat(study): semester-scope the study-tool reads (#141 reframe)

AndresL230:
## #141, reframed (approved by Andres on 2026-07-30)

Closes #141.

No Archive toggle. The existing semester selector (All-semesters default untouched — the #360 e2e-vetoed contract) now scopes the **study-tool read paths** the way it already scopes the graph. Full analysis + reframe rationale: the corrected-state comment on #141.

### Backend — optional `semester` (term label via `academics.term_id_for_label`), strict resolution
- **`services/academics.py`**: `resolve_offering(..., fallback: bool = True)` — new strict mode. The existing `create=False` path silently falls back to *any* offering of the course; with an explicit `semester` that would have silently served another term's content. `fallback=False` returns `None` on a term miss. Additive, default-preserving; resolver tests pin both modes.
- **`routes/study_guide.py`**: `GET /{user}/guide?semester=`, `POST /regenerate` (body), `GET /{user}/exams?semester=` — term miss → 404 (guide/regenerate) with the agent never invoked, exams filtered via `term_for_offering`.
- **`routes/flashcards.py`**: `GET /user/{user}?semester=` — cards filtered to the term's offerings; **term-less cards (offering_id NULL) stay visible under any selection**; unknown label degrades to term-less-only, never 500. `POST /generate` grounds its docs context in the selected term's offering. `import/commit` stays current-term (comment).
- **`routes/notes.py`**: course-filtered read takes `semester`; term miss → empty list. Create/re-home stay current-term by design (comments).
- **`routes/quiz.py`**: untouched — no term resolution exists; quiz scoping is already client-side via the graph picker.
- 17 new pytest cases (red-first): per-route current-term default / scoped / miss-degrades-not-500, plus the resolver modes.

### Frontend
- `lib/api.ts`: optional `semester` on `getStudyGuideExams` / `getStudyGuide` / `regenerateStudyGuide` / `getFlashcards` / `generateFlashcards`.
- `Study.tsx`: `useActiveSemester()` threaded into both modes, fetches gated on the hydrated flag (Dashboard pattern, call-count-pinned — no unscoped-then-scoped double fetch). Notetaker deliberately not wired (no semester context on that screen; the notes param is API-completeness).
- Journey `e2e/study-semester.spec.ts`: All-semesters default shows the fall + spring decks together; hub → "Fall 2025" → only the fall deck serves. No generation triggered (function-mode-seam safe); no new interactive elements → no surface registration needed.

### Gates
[...]

PR #499, opened 2026-07-31
fix(study): keep the exam selected when opening a recent guide (#476)

AndresL230:
Part of #476.

## The issue body misattributes the cause

> "the recent-guides open path clears the selected exam"

It doesn't. `openRecent` (`Study.tsx`) sets **both** ids:

```js
recentTermRef.current = entry.semester ?? "";
setCourseId(entry.course_id);
setExamId(entry.exam_id);
```

The clearing comes from the **courseId-keyed exams effect**, whose first
statement was an unconditional `setExamId("")`. That reset cannot tell
*"the user switched course"* (selection now invalid) from *"we just opened a
specific guide"* (selection deliberate and valid) — it only sees that
`courseId` differs from its last value, which `openRecent` also causes.

Both effects run in the **same commit**, so the loader still saw the intact
pair and the guide loaded; only the *next* render lost the exam. That is why
the symptom is "a guide on screen above a dead Regenerate button" rather than
"nothing opens" — and why `disabled={… || !examId}` pins the button off.

**Discriminator (pinned as a control test):** it needs a course *change* to
reproduce. A rail entry whose course is already selected never trips the
effect and worked fine on main. That test passed before the fix and still
passes after — it's what proves the mechanism is the reset, not the open path.

## The fix

The reset now happens at the two events that actually mean it:

1. **Course picker `onChange`** — the event that genuinely invalidates the exam.
2. **Term switch** — a re-scope of the exam list, so the exam picked under the
   old term need not exist in the new one.

The term case adjusts state **during render** (the `StudyModePanel` pattern
already in this file) rather than in an effect. That matters: an effect-time
reset lands a render late, so the loader commits one read of the old exam
under the *new* term first. That was the same defect's second trigger — it
reproduced on main and now has a test.

The exams effect is now purely a fetch.

## One consequence worth flagging

Making Regenerate reachable on the rail path **exposed a term hazard**.
`regenerate` sent the **active selector's** term, while a recent entry opens
under its **own** term (#475 F1). On main that was unreachable — the button
was dead on the one path where the two differ — so it was latent. Enabling
the button would have made it live: regenerating a Fall guide while the
selector says Spring rebuilds against an offering the displayed guide never
came from, which is exactly the mismatch #475 F1 fixed for reads.

Regenerate now replays the term the displayed guide was **loaded** with.

## Tests

- `Study.recentGuides.test.tsx` (new, 7 cases) — written red first:
  4 failed on the `Regenerate`/picker assertions, the term-switch case failed
  on the stale re-read, and the same-course control passed throughout.
- `Study.semester.test.tsx` — updated a comment that documented the bug as
  permanent ("Regenerate isn't drivable from the recent-guides rail").
- `e2e/study-recent-guides.spec.ts` (new) — promoted regression journey.
- `db/seed_local_rich.py` — seeds a **cached** guide so the journey can open
  the rail without generating; the `study_guide` agent has no function-mode
  handler, and the guide GET returns the cache hit before reaching it. The
  journey asserts Regenerate is *enabled* and deliberately never clicks it.

## Gates

- `tsc --noEmit` clean; `eslint src` 0 errors; `vitest run` 422/422.
- backend `pytest` 1529 passed / 32 skipped; `ruff check` clean.
- Full local e2e cycle (up → Playwright → oracles → down) in one flock.

## Summary by CodeRabbit

* **New Features**
  * Added a cached CS101 Fall 2025 midterm study guide for local study experiences.
  * Recent guides now preserve the associated course, exam, and semester when opened.
  * Regeneration uses the selected guide’s original term and keeps the Regenerate option available.

* **Bug Fixes**
  * Prevented outdated exam and guide selections from carrying over when switching courses or semesters.
  * Improved failed-guide retries by preserving the original study selections.

### DIFF
--- a/backend/db/seed_local_rich.py
+++ b/backend/db/seed_local_rich.py
@@ -575,6 +575,51 @@ def seed_flashcards() -> None:
         )
 
 
+# (guide_id, user_id, offering_id, exam_id, generated_at, content)
+#
+# A CACHED guide, so /study's "Recent guides" rail has an entry to open without
+# any generation: the guide GET returns a cache hit on (user, offering, exam)
+# before it ever reaches the study_guide agent, which has no function-mode
+# handler. `exam_id` points at the real fall-2025 CS101 exam assignment so the
+# exam picker can resolve the same row the rail entry opens.
+_STUDY_GUIDES = [
+    ("rich-guide-cs-f25-mid", USER_ACTIVE, OFF_CS_F25, "rich-asg-cs-f25-mid",
+     "2026-03-01T12:00:00Z",
+     {
+         "exam": "Midterm Exam",
+         "due_date": "2025-10-15",
+         "overview": "Covers variables, control flow, and functions.",
+         "topics": [
+             {
+                 "name": "Variables",
+                 "importance": "Every later topic builds on binding names to values.",
+                 "concepts": ["Assignment", "Scope"],
+             },
+             {
+                 "name": "Recursion",
+                 "importance": "The midterm's hardest questions are recursive traces.",
+                 "concepts": ["Base case", "Call stack"],
+             },
+         ],
+     }),
+]
+
+
+def seed_study_guides() -> None:
+    for guide_id, user_id, off_id, exam_id, generated_at, content in _STUDY_GUIDES:
+        h.insert_if_absent(
+            "study_guides",
+            guide_id,
+            {
+                "user_id": user_id,
+                "offering_id": off_id,
+                "exam_id": exam_id,
+                "generated_at": generated_at,
+                "content": content,
+            },
+        )
+
+
 # (qa_id, concept_node_id, difficulty, score, total, questions_json, answers_json, completed_at)
 _QUIZ_ATTEMPTS = [
     ("rich-qa-cs-variables-1", "rich-node-cs-variables", "easy", 9, 10,
@@ -670,7 +715,8 @@ _SUMMARY_ORDER = [
     "schools", "courses", "course_offerings", "users", "user_profiles", "user_roles",
     "enrollments", "graph_nodes", "graph_edges", "node_mastery_events",
     "gradebook_categories", "assignments", "rooms", "room_members", "room_messages",
-    "notes", "documents", "flashcards", "quiz_attempts", "sessions", "messages",
+    "notes", "documents", "flashcards", "study_guides", "quiz_attempts", "sessions",
+    "messages",
 ]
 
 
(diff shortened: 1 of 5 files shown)


AndresL230:
### Code review

Found 1 blocking issue, now fixed in 1778215, plus two notes recorded rather than fixed.

1. Re-picking the already-selected course wiped the loaded guide — a regression this PR introduced (bug due to `CustomSelect.commit()` calling `onChange(opt.value)` unconditionally, with no equality check against the current value).

Moving the reset out of the `courseId`-keyed effect onto the picker's `onChange` dropped a guard the effect had for free: `setCourseId(sameValue)` bails out, so the effect never re-ran. On a plain handler it has to be explicit. Reopening the course dropdown and clicking the course you were already on cleared `examId`/`guide` — the same class of defect as #476 itself.

https://github.com/SaplingLearn/Sapling/blob/065b3252f54fad1eb49fd59292fa75ba46b1e968/frontend/src/components/screens/Study.tsx#L252-L256

Fixed by guarding on an actual change, with a regression test (`re-picking the course already selected leaves the loaded guide alone`) that was watched fail first:

https://github.com/SaplingLearn/Sapling/blob/177821548b6856eae973a9472e16669b0be3b9c9/frontend/src/components/screens/Study.tsx#L252-L262

Notes, not blocking:

2. Cross-term rail open leaves the exam picker on its placeholder. The exam OPTIONS follow the active selector, so an entry opened under a different term isn't in that list and `CustomSelect` falls back to the placeholder — while the guide and Regenerate are live and correctly aimed at `loadedTerm`. This is not a regression (pre-fix that case was strictly worse: Regenerate was dead too), and listing the entry's term instead would offer exams the next picker-driven load would 404 on, since #475 pins only the recent-open's own load. Squaring it means tracking "the term I'm viewing" across list and loads — a #475 change. Recorded in a comment at the effect.

https://github.com/SaplingLearn/Sapling/blob/177821548b6856eae973a9472e16669b0be3b9c9/frontend/src/components/screens/Study.tsx#L275-L290

3. `loadGuide` has no stale-response guard, so two rail entries clicked in quick succession can resolve out of order. Pre-existing; this PR adds `loadedTerm` to the state that can go stale. Worth a follow-up, out of scope here.

Also checked and cleared: a review pass flagged a captured Playwright failure of this PR's own journey as evidence the fix doesn't work. That artifact is from a deliberate pre-fix run — `e2e/results/last-run.json` recorded `argv` of `playwright test e2e/study-recent-guides.spec.ts`, the single-spec invocation used to verify the journey goes red against pre-fix `Study.tsx`. Gitignored, so it was never part of this PR; deleted to stop it misleading future readers.

Issue #388, opened 2026-07-22
test(e2e): browser-lane stability gate — 20 consecutive green runs


AndresL230:
## Context
Chapter 2 and the CI gate both wait on this: proof the browser lane is *stable*, not merely passing. Any flake found here is a bug to fix, not a retry to add.

## Scope
- Run the browser lane 20× in a loop locally.
- Fix every flake at the root (no added retries, no timeout padding).

## Acceptance criteria
- [ ] 20/20 green, documented in this issue.

## Dependencies
Depends on: #386, #387
Blocks: #389, #399

---
Chapter: 1 (Regression Suite) | Lane: browser | Size: S | Suggested owner: Andres
Part of epic #402

AndresL230:
## Stability gate: PASSED — 20/20 consecutive green runs

Full browser suite (all 7 specs / 8 tests: smoke, dashboard, upload, tutor, quiz, study-room ×2-contexts, graph ×2+fixme) against a single `make e2e-up` boot (function mode, dummy GEMINI_API_KEY per #439), local retries=0 so every pass is a first-attempt pass:

| Run | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20 |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| ✅ 8 passed (s) | 28 | 25 | 25 | 26 | 25 | 25 | 25 | 24 | 25 | 25 | 24 | 24 | 28 | 29 | 28 | 29 | 28 | 29 | 29 | 28 |

Zero flakes found — nothing to fix at the root. Unblocks #389 (CI job, PR #440) and #399 (Chapter 2).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #500, opened 2026-07-31
fix(e2e): poll for the tutor turn's rows instead of reading once (#477)

AndresL230:
Part of #477.

## Cause

The persistence assert read `messages` once, immediately after the reply text rendered — treating "the reply is on screen" as "the rows are committed". Those are different signals.

The server side is correct: `stream_agent_turn` persists inside `on_complete` and only *then* yields `done` (`backend/services/chat_stream.py:372-393`), so the write does precede the end of the turn. But the composer renders off the streamed **tokens**, so both UI assertions can pass while `done` is still in flight — leaving a window where the one-shot read sees only the 4 seeded rows. That matches the reported failure exactly (expected 6, received 4, UI assertions green).

So this is a test-side assumption, not a product regression — worth stating, since the alternative reading (rows should already be there) would have pointed at a real persistence bug.

## Fix

Bounded `expect.poll` on the row count, the idiom `events.spec.ts` already uses for its fire-and-forget rollup. Root fix rather than a retry, per the #388 zero-flake policy.

## Honest limitation

A ~1-in-6 race is not reproducible on demand, so this is **not** verified by a watched-red test. It rests on the persist-before-`done` contract above plus green cycles. What the change does guarantee: the assert can no longer fail *because it read too early* — it now waits for the condition it depends on, or fails after 5s.

## Gates

tsc clean; full local cycle green (Playwright 37/37, oracles 0 findings).

## Summary by CodeRabbit

* **Tests**
  * Improved tutor end-to-end test reliability by waiting for expected messages to be persisted before validation.
  * Updated assertions to verify message roles and encrypted content more consistently.

### DIFF
--- a/frontend/e2e/tutor.spec.ts
+++ b/frontend/e2e/tutor.spec.ts
@@ -72,15 +72,26 @@ test("tutor turn renders a reply and persists encrypted to messages", async ({
   await expect(log).toContainText(TUTOR_REPLY);
 
   // ── Raw-SQL readback: the turn persisted, encrypted (#397 posture) ───────
-  const rows = (await queryRaw(
-    `SELECT role, content FROM messages
-      WHERE session_id = $1
-      ORDER BY created_at ASC`,
-    [SESSION_ID],
-  )) as { role: string; content: string }[];
+  // Polled, not read once (#477). The server persists inside on_complete and
+  // only THEN yields `done` (services/chat_stream.py) — the write really does
+  // precede the end of the turn. But the composer renders the reply off the
+  // streamed TOKENS, so the two assertions above can both pass while `done`
+  // is still in flight: a rendered reply is not a persistence signal, the
+  // rows are. Bounded poll rather than a retry or a bare sleep, per the #388
+  // zero-flake policy.
+  const readTurnRows = async () =>
+    (await queryRaw(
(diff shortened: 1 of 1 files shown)


AndresL230:
### Code review

Found 1 issue, fixed in 749c76e.

1. Polling the count and then re-querying for the content assertions decoupled two checks the original single read had joined (bug due to `const rows = await readTurnRows()` running as a separate query after the poll).

A duplicate-persistence regression could land a row in that gap; the poll's exact `toBe(N)` would already have passed against the earlier read, and the content assertions would still slice two valid-looking rows off the end and pass. That is precisely the invariant this journey exists to prove (#397 posture). `frontend/e2e/events.spec.ts` avoids this by capturing the payload as a side effect inside the poll predicate — same shape now used here.

https://github.com/SaplingLearn/Sapling/blob/749c76e/frontend/e2e/tutor.spec.ts#L88-L101

Checked and cleared: the poll is not weaker than the original assert (same exact `toBe`, just retried); the persist-before-`done` claim holds (`on_complete` at `chat_stream.py:372`, `done` yielded at `:393`, and `_persist` in `learn.py` is a synchronous `save_message`); no CLAUDE.md violation.

Issue #482, opened 2026-07-30
RAG hardening: silent-empty observability, poisoned embedding:None rows, durability inversion


AndresL230:
Split from the #323 audit (verified @ d1d724f). Three related operational hazards in the delivered document-RAG path:
1. **Silent-empty everywhere**: `retrieve_chunks` catches ALL exceptions → `[]` via a bare `print` (rag_service.py:137-139) — invisible to the app-logging path; an ungrounded tutor turn is indistinguishable from a grounded one at every layer. → logger + a metric/event.
2. **Poisoned rows**: a failed embed batch is logged and then upserted anyway with `embedding: None` (rag_service.py:192,213) — permanently unretrievable, and nothing ever re-indexes them.
3. **Durability inversion**: DBOS wraps `/upload/sync` (documents.py:615) which never indexes; the indexing route (streaming `/upload`) is non-durable by design (ADR 0011:179-181) — a crash between `_persist_document` and the post-roll task loses chunks permanently behind a healthy-looking documents row. At minimum document the gap; ideally give indexing a retry/backfill trigger (backfill_document_chunks.py exists but nothing invokes it).

AndresL230:
Items 1 and 3 addressed in #501 (`c03c983`). Item 2 was already fixed on `main`.

## Re-verified against `main` before starting

| # | Claim | State found |
|---|---|---|
| 1 | Silent-empty: bare `print` at `rag_service.py:137-139` | **live** → fixed |
| 2 | Failed batch upserted anyway with `embedding: None` | **already fixed** — `rag_service.py` filters `embedding is not None` before the upsert, with a comment citing this issue |
| 3 | Durability inversion | **live** → documented |

## Item 1

`rag_service` reported everything through bare `print`, so no failure reached the app-logging path and no rollup could count one. That matters most for `retrieve_chunks`, which degrades to `[]` — and `[]` is also "nothing relevant matched", so an ungrounded turn was indistinguishable from a grounded one at every layer above.

Module logger, all three `print` sites moved onto it, plus `rag.retrieval_failed` / `rag.chunks_dropped` events so the degrade rates are countable.

[...]

Issue #439, opened 2026-07-28
RAG embedding path sits below the SAPLING_MODEL_MODE seam — live embed calls fire even in function mode


AndresL230:
Found independently by three journey PRs (#432 upload, #434 tutor, #437 quiz) and verified by the #432 review panel.

`services/rag_service.py` (`_embed_query`, `_embed_documents_batch`) and `routes/documents.py::_index_document_chunks` (~:1088–1109, incl. the catalog-relevance gate) construct raw `google.genai.Client`s directly — they predate the #391 seam (`agents/_providers.py::model_for`) and never consult `SAPLING_MODEL_MODE`. Consequences in `function` mode:

- one live `gemini-embedding-001` call per document upload (post-roll, fire-and-forget, `except Exception` swallowed)
- one per quiz generate (RAG grounding)
- one per tutor turn when the course has a `course_code`

Determinism is unaffected (failures are swallowed → empty results), but the 'hermetic' E2E lane silently bills when a real `GEMINI_API_KEY` is present. The pytest hermetic guard (#379) can't help — it's an autouse fixture, absent in the live uvicorn.

Fix direction: route embeddings through a seam-aware provider (mode-gated no-op or scripted vectors in non-real modes), or gate `_index_document_chunks`/`_embed_query` on `_model_mode() == "real"`. Also relevant to the #354/#358 fresh-client sweep.

Interim mitigation used by the journeys: dummy `GEMINI_API_KEY` during verification runs.

Part of #402 follow-ups.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #501, opened 2026-07-31
fix(rag): put retrieval/indexing failures on the app-logging path (#482)

AndresL230:
Part of #482.

## Scope, after re-verifying the issue against `main`

The issue lists three hazards. Only one and a half were still live:

| # | Claim | State |
|---|---|---|
| 1 | Silent-empty: bare `print`, invisible to app logging | **live — fixed here** |
| 2 | Poisoned `embedding: None` rows persisted anyway | **already fixed on main** (`rag_service.py` filters them before upsert) |
| 3 | Durability inversion | **live — documented here**, code fix deliberately deferred |

## Item 1

`rag_service` reported every failure with a bare `print`, so failures never reached the app-logging path and no rollup could count them. That matters most for `retrieve_chunks`, which degrades to `[]` — and `[]` is also what "nothing relevant matched" returns, so an ungrounded tutor/quiz turn was indistinguishable from a grounded one at every layer above.

Adds a module logger, moves all three `print` sites onto it, and emits `rag.retrieval_failed` / `rag.chunks_dropped` (category `error`) so the degrade rates are countable.

## The part worth reviewing

The naive version of this is wrong, and the existing #439 tests caught it. `_require_real_mode()` raises `_EmbeddingDisabled` whenever `SAPLING_MODEL_MODE != "real"` — which is **every function-mode run**. So a blanket `except Exception` would emit an error event on every e2e tutor turn and every e2e upload, and the new signal would be pure noise the day it shipped.

`_EmbeddingDisabled` is now caught separately: INFO, no event. Only genuine failures warn and spend an event. Both #439 transport guards are updated to assert the new channel *and* to pin the no-event contract — the invariant they protect (no egress in function mode) is unchanged; only the output moved off stdout.

## Item 3 — documented, not fixed

Recorded in `docs/architecture.md` under Known sharp edges: DBOS wraps `/upload/sync`, which never indexes; the route that *does* index is the streaming `/upload`, which is non-durable by design (ADR 0011). A crash between `_persist_document` and the post-roll indexing task leaves a healthy-looking `documents` row whose content is absent from retrieval permanently, and `scripts/backfill_document_chunks.py` is invoked by nothing.

Wiring a retry trigger is a real design call (where it runs, what re-drives it, how it interacts with the streaming route's X-Request-ID replay semantic), so I documented the gap rather than guessing at it — the issue itself asks for documentation as the minimum bar.

## Gates

Backend pytest 1531 passed / 32 skipped (2 new), ruff clean, full local e2e cycle green (Playwright 37/37, oracles 0 findings).


AndresL230:
### Code review

Found 2 issues, both fixed in c03c983.

1. The two new event types were emitted **outside** the pinned #117 taxonomy. `log_event` deliberately does not enforce membership (it must never raise), so nothing failed — they were simply undocumented and missing from the frozenset whose whole job is to make a rename break loudly.

https://github.com/SaplingLearn/Sapling/blob/5d741c9/backend/services/rag_service.py#L148-L158

Registered in `EVENT_TAXONOMY`, the docstring table, and the exact-set test. Also recorded *why* they are not named `error.*`: `/api/admin/analytics/errors` filters `event_type like error.*` and projects an HTTP shape (path/method/status_code/duration_ms), so renaming would fill an HTTP-request table with null-path rows. These surface via `/usage/summary`'s `by_event_type` instead. Giving that feed a shape-agnostic projection is the real fix, and is out of scope here.

2. `test_retrieve_chunks_failure_is_logged_and_counted` depended on ambient `SAPLING_MODEL_MODE` (bug due to `_require_real_mode()` running before the mocked `_client` is reached). With function mode exported — which the E2E workflow instructs — it took the `_EmbeddingDisabled` branch and passed **vacuously** on an unrelated code path. Reproduced, then pinned to real mode and verified passing with the var both set and unset.

https://github.com/SaplingLearn/Sapling/blob/c03c983/backend/tests/test_rag_service.py#L389-L400

Noted, not fixed: six pre-existing tests in that file share the same latent dependence (`task_type`, `returns_count`, `chunk_ids`, `dedupe`). The hermetic lane runs with the var unset by contract, so they are green in CI, but the file deserves its own cleanup.

Checked and cleared: `_EmbeddingDisabled` is the only type `_require_real_mode` raises and is caught ahead of the generic handler in both call sites; the `embedding_disabled` batch flag is correct given `model_mode()` cannot change mid-call; normal empty results never emit an event; `log_event` is non-blocking and cannot raise into the request path; no import cycle (the keyless `import main` guard still passes).

Issue #316, opened 2026-07-04
staging: avatars storage bucket is private but code expects public read


Darkest-Teddy:
## Summary

  On the **staging** Supabase project (`ybgqdonkoqftwrmweuyv` / `sapling-staging`), the `avatars` storage bucket is `public = false`, but the codebase treats avatars as intended-public read (per the comment in
  `backend/db/migrations/0029_storage_lockdown_231.sql`: *"`avatars` stays public (intended public read)"*). This likely means avatar images fail to load on staging.

  ## Evidence

  Read-only query against staging `storage.buckets`:

  ```
   id      | public | file_size_limit | allowed_mime_types
  ---------+--------+-----------------+--------------------
   avatars | false  | null            | null
  ```

  (Note: `issues-media-files` and `application_resumes` buckets don't exist on staging at all.)

  ## Expected

  `avatars.public = true` on staging, matching prod and the documented intent.

  ## Suggested fix

  Flip the bucket to public on staging (via the sanctioned path — a migration or the storage admin API, not ad-hoc dashboard SQL, per `CLAUDE.md`). Confirm prod's `avatars` bucket is already public before/while doing so.

  ## Context

  Discovered incidentally while repairing staging's migration ledger during the RAG document-chunking work (2026-07-03). Not caused by that work.

AndresL230:
Migration landed in #502 (`ac5e21e`). **Leaving this open** — it does not fix staging by itself; that needs `python -m db.migrate` against the staging project, which is yours to run.

## Why the suggested fix wasn't quite the fix

The issue proposes flipping the bucket on staging. That works, but it doesn't explain why staging drifted, and the same drift stays possible on the next environment. Both mechanisms that "create" this bucket decline to correct an existing one:

- `0011_avatars_bucket.sql` inserts it with `public = true`, then ends in `ON CONFLICT (id) DO NOTHING` — a no-op once the row exists.
- `storage_service.ensure_bucket_exists` (lifespan, `public=True`) treats the Storage API's `409` as success and *deliberately* does not overwrite settings, "in case an admin has intentionally tuned them in the dashboard".

So a bucket that came into existence private stays private forever, through any number of deploys and migrations. An `UPDATE` is the only thing that corrects it — hence `0041` rather than a re-run of `0011`.

[...]

PR #502, opened 2026-07-31
fix(storage): converge the avatars bucket on public read (#316)

AndresL230:
Part of #316.

## The issue's suggested fix would have worked, but not lasted

The issue proposes flipping the staging bucket to public. That fixes staging. It does not explain why staging drifted, and it leaves the same drift possible on the next environment.

**Both mechanisms that "create" this bucket decline to correct an existing one:**

- `0011_avatars_bucket.sql` inserts it with `public = true` but ends in `ON CONFLICT (id) DO NOTHING` — a no-op once the row exists.
- `storage_service.ensure_bucket_exists` (called from `main.py`'s lifespan with `public=True`) treats the Storage API's `409` as success and *deliberately* does not overwrite settings, "in case an admin has intentionally tuned them in the dashboard".

So a bucket that came into existence private stays private forever, through any number of deploys and migrations. An `UPDATE` is the only thing that corrects it — hence a new file rather than a re-run of 0011.

## The one judgment call

This deliberately overrides `ensure_bucket_exists`'s "an admin may have tuned it" stance, for this bucket only. The read path is an unauthenticated `` against `/storage/v1/object/public/avatars/...`, so private isn't a valid tuning — it's broken avatars. `0029` already recorded the intent (*"`avatars` stays public (intended public read)"*); this asserts that state instead of assuming it.

## Verification

- **From-empty replay** (`scripts/local-db-reset.sh`, not a normal cycle — `e2e-up` runs against a DB that already has the schema): the whole chain applies clean with 0041 in it.
- **Effect on a drifted bucket**: forced `avatars` to `public=false` to reproduce the staging condition, then ran the statement — flips to `true`, idempotent on re-run, and a safe zero-row no-op against a non-existent bucket id.
- Full local e2e cycle green (Playwright 37/37, oracles 0 findings).

## This does not fix staging by itself

It converges on the next `python -m db.migrate` against that project. #316 should stay open until that's run — I don't have (and shouldn't use) staging credentials for it.

### DIFF
--- a/backend/db/migrations/0041_avatars_bucket_public.sql
+++ b/backend/db/migrations/0041_avatars_bucket_public.sql
@@ -1,0 +1,42 @@
+-- 0041: converge the `avatars` bucket on public read (#316).
+--
+-- Staging drift: on `sapling-staging` (ybgqdonkoqftwrmweuyv) the avatars bucket
+-- is public=false, while prod is public=true and 0029 states the intent
+-- explicitly ("`avatars` stays public (intended public read)"). Avatar <img>
+-- src values are plain public object URLs, so a private bucket serves 400s and
+-- every avatar on staging renders broken.
+--
+-- WHY THIS DRIFT NEVER SELF-HEALS (the part worth knowing):
+-- both mechanisms that "create" this bucket decline to correct an existing one.
+--   * 0011_avatars_bucket.sql INSERTs it with public=true but ends in
+--     ON CONFLICT (id) DO NOTHING — a no-op if the row already exists.
+--   * storage_service.ensure_bucket_exists (called from main.py's lifespan with
+--     public=True) treats the Storage API's 409 as success and deliberately
+--     does NOT overwrite settings, "in case an admin has intentionally tuned
+--     them in the dashboard".
+-- So a bucket that came into existence private stays private forever, no matter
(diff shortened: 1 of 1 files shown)


AndresL230:
### Code review

Independently verified beyond the diff:

- **Transactionality/idempotence** — `db/migrate.py` commits the migration SQL and the `schema_migrations` ledger insert together, so this runs exactly once per environment; the `UPDATE` is a no-op when already true and a zero-row no-op when the bucket is absent. From-empty replay is safe because `0011` (which INSERTs the row) sorts before `0041`.
- **Privilege claim is real, not assumed** — `0029` already performed an `UPDATE storage.buckets` through this same runner, and the prod snapshots in the working tree confirm those changes are live in prod today.
- **Security direction** — `upload_cosmetic_asset` writes into the same bucket under `cosmetics/`, so it's worth stating explicitly: those are admin-managed achievement/role reward assets, not user PII, and `docs/security/storage-hardening-plan.md` scoped avatars as intended-public while locking down `issues-media-files` and `application_resumes`. Prod has had avatars public all along — this converges the drifted staging copy onto already-live, already-reviewed state rather than opening new surface.
- **Numbering** — 0041 is the correct next number; no open PR touches `db/migrations/`.

Issue #265, opened 2026-06-24
Rewire Gradescope sync onto enrollment-keyed gradebook (post-redesign)


AndresL230:
## Context

The DB modular redesign (PR #264, branch `db/schema-foundation`) **absorbed the Gradescope DB tables** from `origin/Gradebook` (`0020_gradescope.sql`) and re-expressed them against the new enrollment-keyed schema (migrations `0021` + `0027`). **That migration file on `origin/Gradebook` is now superseded — drop it.** This issue tracks the **code** rewire so the Gradescope sync works against the new schema.

## What changed in the schema

| Original (`origin/Gradebook`) | Redesigned (PR #264) |
|---|---|
| `gradescope_credentials` (user_id PK, encrypted) | unchanged ✓ (`0027`) |
| `gradescope_course_links.sapling_course_id → courses(id)` | **`enrollment_id → enrollments(id)`**, `UNIQUE(enrollment_id, gradescope_course_id)` (`0027`) |
| `assignments.gradescope_assignment_id` + unique idx on `(course_id, …)` | kept; **idempotency index re-targeted to `(enrollment_id, gradescope_assignment_id)`** (`0021`) — assignments are now enrollment-scoped |

## Code to rewire

- The sync upsert must key on **`enrollment_id`**, not `course_id` (assignments no longer carry `course_id`).
- The course-picker links a Gradescope course to a user's **enrollment** (a specific class instance), not an abstract course. **Confirm this matches the intended UX** — if linking should be per-abstract-course, adjust the `0027` FK and the picker together.
- Credentials encrypt/decrypt is unchanged (`email`/`password`/`cookies` are 🔒 encrypted TEXT).

## Acceptance

- Gradescope sync upserts grades into the correct enrollment's assignments, idempotently, against the new schema; tests updated.

Refs: PR #264 · supersedes `origin/Gradebook:backend/db/migrations/0020_gradescope.sql`.

AndresL230:
Rewired in #504 (`b5555d5`). **Leaving open** for your verification — this touches the Gradescope UI's behaviour even though no frontend code changed.

## It was worse than this issue describes: five drift points, not one

`routes/gradescope.py` was never migrated past the DB redesign, so every link/sync/delete has been returning a PostgREST 400 **in production** — the same failure mode as #405:

| # | Code said | Schema says |
|---|---|---|
| 1 | table `user_courses` | renamed `enrollments` (0020) |
[...]

PR #504, opened 2026-07-31
fix(gradescope): rewire onto the enrollment-keyed schema (#265)

AndresL230:
Part of #265.

## Five drift points, not one

`routes/gradescope.py` was never migrated past the DB redesign. It named five things the schema does not have, so every link/sync/delete returned a PostgREST 400 in production — the same failure mode as #405:

| # | Code says | Schema says |
|---|---|---|
| 1 | table `user_courses` | renamed `enrollments` (0020) |
| 2 | `gradescope_course_links.user_id` / `.sapling_course_id` | `enrollment_id` (0027) |
| 3 | `assignments.user_id` / `.course_id` | enrollment-scoped (0021) |
| 4 | table `course_categories` | `gradebook_categories` (0021) |
| 5 | writes `source='gradescope'` | CHECK allows only `{manual, syllabus}` |

Point 5 matters: even a perfect column fix would still have failed every insert, because the sync had no legal value to write. Migration **0042** widens the CHECK — a strict superset, so it is added VALIDATED with nothing to grandfather.

## The UX question the issue asked to confirm

Answered **enrollment-keyed**. Grades land on enrollment-scoped `assignments`, so a link has to name the specific class instance — a retake gives two enrollments and an abstract-course link cannot say which term's assignments to write. The schema already said this in 0027; the code hadn't caught up.

Per the repo convention the HTTP boundary keeps the abstract `course_id` and resolves inward through the existing `academics.enrollment_id_for`, so **the frontend contract is unchanged** — `list_links` maps enrollment rows back out to course ids on the way through.

## The reason this survived the lane built to catch it

`RUN_INTEGRATION=1 pytest -m integration` — the documented invocation — ran **1 passed, 27 skipped** and exited **0**. It reported green having tested nothing.

Collecting the whole tree imports `tests/test_benchmark_quiz.py` and `tests/test_seed_quiz_fixture.py`, which import scripts that call `load_dotenv(".env.staging", override=True)` at module import. That clobbers `SUPABASE_URL` to staging for the rest of the session, so `_require_local_stack` saw a non-local URL and skipped all 27. `pytest tests/integration/` (directory form) ran fine on the same stack, which is presumably why it looked healthy.

The session fixture now re-asserts the local env at run time — not only at conftest import, which happens *before* those script imports — and **raises rather than skipping** when it is still wrong. That is the rule the same file already applied to `SUPABASE_DB_URL` (*"raises loudly rather than skipping... a silent skip would read as safe"*), now applied to `SUPABASE_URL` too.

Documented form: **1 passed / 27 skipped → 28 passed**.

I include this here rather than splitting it because the four new integration tests are #265's regression guard, and a guard that silently skips is not one.

## Verification

- **4 new integration tests** drive `POST /link`, `GET /links`, `DELETE /link/{id}` against real Postgres and read back with raw SQL — the assertions that would have caught the original drift. The mocked suite structurally cannot: it agrees with whatever column list the caller names.
- Unit tests pin the *shape* (enrollment_id written, never the course id) so a regression back to the abstract id fails fast.
- From-empty replay for 0042; hermetic 1534 passed; integration 28 passed; browser 37/37; oracles 0 findings.

**Not covered:** `POST /sync/{id}` drives a live Gradescope login and scrape, so its DB surface has no offline seam. Stated in the test file rather than left implicit.

### DIFF
--- a/backend/db/migrations/0042_assignments_source_gradescope.sql
+++ b/backend/db/migrations/0042_assignments_source_gradescope.sql
@@ -1,0 +1,20 @@
+-- 0042: allow 'gradescope' as an assignment source (#265).
+--
+-- 0021 recreated `assignments` with
+--   source TEXT NOT NULL DEFAULT 'manual' CHECK (source IN ('manual','syllabus'))
+-- but routes/gradescope.py::sync_course writes source='gradescope' on every
+-- synced row. So even once the column drift in that module is fixed, every
+-- insert would fail the CHECK — the sync has no valid value to write.
+--
+-- Widening only: {manual, syllabus} is a strict subset of the new set, so every
+-- existing row already satisfies it and the constraint is added VALIDATED
+-- rather than NOT VALID (nothing to grandfather).
+--
+-- The constraint name is Postgres's default for the inline CHECK in 0021
+-- (`<table>_<column>_check`); DROP ... IF EXISTS keeps this replay-safe on an
+-- environment where it was already renamed or absent.
+
+ALTER TABLE assignments DROP CONSTRAINT IF EXISTS assignments_source_check;
+
+ALTER TABLE assignments ADD CONSTRAINT assignments_source_check
+    CHECK (source IN ('manual', 'syllabus', 'gradescope'));
--- a/backend/db/migrations/0042_assignments_source_gradescope.sql
+++ b/backend/db/migrations/0042_assignments_source_gradescope.sql
@@ -7,8 +7,16 @@
 -- insert would fail the CHECK — the sync has no valid value to write.
 --
 -- Widening only: {manual, syllabus} is a strict subset of the new set, so every
--- existing row already satisfies it and the constraint is added VALIDATED
--- rather than NOT VALID (nothing to grandfather).
+-- existing row already satisfies it by construction.
+--
+-- Added NOT VALID anyway, and the reason is lock behaviour rather than data
+-- (an earlier draft of this comment conflated the two). A plain ADD CONSTRAINT
+-- ... CHECK takes ACCESS EXCLUSIVE on `assignments` for the whole validation
+-- scan, blocking every read and write to the gradebook's busiest table for its
+-- duration. NOT VALID skips that scan while still enforcing the constraint on
+-- every new and updated row — which is all this needs, since existing rows
+-- cannot violate a widened set. Matches 0021's convention for CHECKs on
+-- populated tables.
 --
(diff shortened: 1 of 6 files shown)


AndresL230:
### Code review

Found 3 issues, all fixed in b5555d5.

1. **The write routes re-derived the enrollment instead of following the existing link** — and it was the exact retake case this PR's rationale uses to justify enrollment-keying (bug due to `enrollment_id_for` resolving one current-term enrollment per call, while `list_links` enumerates all of them).

The seeded user holds CS101 in both fall-2025 and spring-2026. A link attached to the enrollment the heuristic doesn't pick was invisible to every write path: re-linking created a **second** row, `DELETE` removed nothing while still answering `ok: true`, and sync 400'd "No Gradescope course is linked" for a course `GET /links` was still listing. The GET/write asymmetry was the tell.

https://github.com/SaplingLearn/Sapling/blob/2c249c2/backend/routes/gradescope.py#L406-L419

Fixed with `_all_enrollments_for()`: `upsert_link` deletes across every enrollment in the course before inserting (so a rollover leaves exactly one link), `remove_link` deletes across all of them, and `sync_course` keys the run on the **link row's** `enrollment_id` rather than a fresh guess — the grades belong to the class instance the link was made against.

Two integration tests cover it, written against the enrollment the resolver does *not* prefer so the fixture is the hostile case. **Verified red against the pre-fix code** (2 failed / 4 passed) and green after.

2. `list_links` answered `sapling_course_id: null` for an enrollment whose course couldn't be resolved. Now dropped — a null there reads as a real link the UI has no id to act on.

3. Migration 0042's rationale answered the wrong question. It justified skipping `NOT VALID` on the data ("nothing to grandfather"), but the reason that matters is lock behaviour: a plain `ADD CONSTRAINT ... CHECK` takes ACCESS EXCLUSIVE on `assignments` for the full validation scan. Switched to `NOT VALID`, which still enforces every new and updated row — all a widened set needs.

Checked and cleared: the drift audit confirms exactly five points, no sixth (the remaining `user_id` filters target `gradescope_credentials`, which is genuinely still user-keyed per 0027); the `in.(...)` PostgREST syntax is correct; the conftest raise doesn't break any CI workflow (none set `RUN_INTEGRATION` without a stack).

Noted, not fixed: the conftest fix repairs `os.environ` but cannot un-freeze a module that already captured a staging value at import. Safe today only because `db.connection` is imported before those scripts during collection, and nothing asserts that ordering. The general hazard — arbitrary scripts calling `load_dotenv(..., override=True)` at import time — deserves its own fix in those scripts.

Issue #317, opened 2026-07-04
Prod migration ledger may have drifted from repo (staging did — 4 unrecorded migrations)


Darkest-Teddy:
## Summary

  Staging's migration ledger (`schema_migrations`) had drifted out of sync with the repo's `backend/db/migrations/` files — 4 files were unrecorded, causing `python -m db.migrate` to crash (it tried to apply obsolete migrations against
  tables the academics redesign had removed). This was repaired on staging (2026-07-03) by baselining the 4 stuck files. **Prod (`jxqcmjqtjlpuxfrxmrdv` / `sapling`) has not been checked and may have the same drift.**

  ## What happened on staging

  `python -m db.migrate` against staging crashed with `relation "course_categories" does not exist`. Investigation found `schema_migrations` was missing:

  - `0019_gradebook_drops.sql` — targets `course_categories` (designed out by the academics redesign → would crash)
  - `0020_gradescope.sql` — objects already present (already effectively applied)
  - `0021_gradebook_curve.sql` — targets `user_courses` (designed out → would crash)
  - `0029_storage_lockdown_231.sql` — target buckets absent on staging (no-op)

  Root cause: staging was created fresh in June 2026 *with the redesigned schema*, and its ledger was only partially baselined. Note the duplicate migration numbers (`0019_conventions_terms_schools` vs `0019_gradebook_drops`,
  `0020_academics_split` vs `0020_gradescope`, `0021_gradebook` vs `0021_gradebook_curve`) — one of each pair got recorded, the other didn't.

  Resolution on staging: `python -m db.migrate --baseline` (metadata-only; recorded the 4 as applied without running them). `db.migrate` now reports "No pending migrations."

  ## Action needed

  1. Check prod's `schema_migrations` vs the repo's migration files for the same gaps.
  2. For any gap, verify against prod's actual schema whether the migration is already reflected / obsolete (baseline) vs genuinely missing (apply). Do **not** blindly baseline prod — confirm each object's presence first.
  3. Consider a guard/CI check that fails when repo migrations and a target env's ledger diverge, to catch this earlier.

  ## Note on connectivity

  The direct DB host (`db.<ref>.supabase.co:5432`) is IPv6-only and unreachable from at least one dev machine; the IPv4 **session pooler** (`aws-1-us-west-2.pooler.supabase.com:5432`, user `postgres.<ref>`) works with `db.migrate`. Worth
  documenting for whoever runs migrations.

  ## Context

[...]

PR #506, opened 2026-08-01
ci: apply pending migrations to staging on merge to main

AndresL230:
Answers "shouldn't the Supabase connection handle this?" — it can't, and here's why.

## Why nothing was applying migrations

Checked all four places it could have happened:

| Candidate | Reality |
|---|---|
| `backend/Dockerfile` | CMD is a bare `uvicorn main:app` — no release step |
| Procfile / railway.json / nixpacks | none exist |
| `main.py` lifespan | does not migrate |
| **Supabase GitHub integration** | reads `supabase/migrations/` (CLI convention) — **this repo has no such directory** (`supabase/` holds only `config.toml` + `snippets`, and `schema_paths = []`) |

That last one is the answer to the question. The integration is connected but inert for migrations, which is exactly why its check reports **"skipping"** on every PR. This repo's migrations are raw DDL under `backend/db/migrations/`, applied by `db/migrate.py` against its own `schema_migrations` ledger — a deliberate choice per CLAUDE.md, not a misconfiguration.

So a merge shipped code whose schema hadn't moved. **#504 is live proof**: it merged code writing `source='gradescope'` while the CHECK still rejects that value until 0042 is applied.

## Scope: staging only

`main` deploys staging. Prod is a separate `production` branch promotion, and auto-applying **irreversible** DDL to prod on merge is a different risk decision — this runner has no down migrations. Not doing that here.

## Two safety properties, both exercised against a real database

**Inert until you opt in.** No `STAGING_SUPABASE_DB_URL` secret → notice + skip. Merging this file changes nothing until you add the secret.

**Refuses to apply onto a drifted ledger.** This is the #317 hazard: if `schema_migrations` is missing, a naive run treats all 45 files as pending and fails recreating existing objects. Preflight fails the job instead, and also fails on any *recorded-but-absent* filename — the repo and database disagreeing about history is not something to pour more DDL onto.

Tested three ways against the local Postgres:

```
healthy      → on disk: 45 | recorded: 45 | pending: 0   exit 0
injected drift → ::error:: recorded but not in repo: 9999_ghost.sql   exit 1
absent ledger  → ::error:: schema_migrations does not exist           exit 1
```

Worth noting the first draft queried a `version` column; the ledger's column is `filename`. Only running it against a real database caught that — it would have failed on the first real merge.

## To enable

Add repo secret **`STAGING_SUPABASE_DB_URL`** = staging's **direct** connection string (port 5432, *not* the pooler). Until then it no-ops.

## Summary by CodeRabbit

* **Chores**
  * Added automated staging database migration checks for migration changes.
  * Migrations now validate migration records before applying updates.
  * Added safeguards to detect missing or outdated migration records.

### DIFF
--- a/.github/workflows/migrate-staging.yml
+++ b/.github/workflows/migrate-staging.yml
@@ -29,6 +29,15 @@ concurrency:
 jobs:
   migrate:
     runs-on: ubuntu-latest
+    # workflow_dispatch lets you pick ANY branch containing this file, so
+    # without this a migration could be applied to shared staging straight from
+    # an unmerged branch — bypassing review. Worse than the bypass: the
+    # filename lands in the ledger, so if the file is then edited before merge
+    # (easy, since it was only "tested"), the merge never re-applies it and
+    # staging silently diverges from the canonical file with no pending/orphan
+    # signal to catch it. That is exactly the immutability rule CLAUDE.md
+    # states — migrations are immutable once applied.
+    if: github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@v4
 
@@ -37,7 +46,9 @@ jobs:
           python-version: "3.12"
 
       - name: Install the runner's only dependency
-        run: pip install "psycopg[binary]"
+        # Range matches backend/requirements.txt so this can't drift onto a
+        # psycopg major the app has never run against.
+        run: pip install "psycopg[binary]>=3.2,<4"
 
       - name: Preflight — report ledger drift instead of pushing through it
         id: preflight
--- a/.github/workflows/migrate-staging.yml
+++ b/.github/workflows/migrate-staging.yml
@@ -1,0 +1,86 @@
+# Apply pending migrations to STAGING when they land on main.
+#
+# Why this exists: nothing else applies them. The backend image's CMD is a bare
+# uvicorn, there is no Procfile/release step, main.py's lifespan does not
+# migrate, and the Supabase GitHub integration reads `supabase/migrations/`
+# (the CLI convention) which this repo does not use — its migrations are raw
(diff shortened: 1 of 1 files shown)


AndresL230:
### Code review

Found 1 real issue and 1 nit, both fixed in 2192652.

1. **`workflow_dispatch` had no branch restriction**, so DDL could be applied to shared staging from an unmerged branch — bypassing the push-to-main gate the whole design assumes.

The bypass isn't the dangerous part. The filename lands in `schema_migrations`, so if the file is then edited before merging (easy — it was only "tested"), **the merge never re-applies it**, and staging silently diverges from the canonical file with *no pending/orphan signal*, because the recorded filename still matches. That's exactly the immutability rule CLAUDE.md states ("migrations are immutable once applied"), violated without a trace — and invisible to the very preflight this PR adds.

Fixed with a job-level `if: github.ref == 'refs/heads/main'`.

2. `psycopg[binary]` was installed unpinned while the app pins `>=3.2,<4`. Matched the range so the runner can't drift onto a major the app has never run against.

Checked and cleared, several of which I'd flagged as suspect: the skip output gates Apply correctly (and `db/migrate.py` independently refuses an empty DSN); a failed preflight also blocks Apply via the implicit `success()`; `filename` is the right ledger column; the Python heredoc's indentation survives the YAML block scalar (verified by extracting and running `bash -n` + `py_compile`, not by eye); the DSN never reaches logs and is auto-masked regardless; `concurrency` serializes DDL runs, and because `db.migrate` re-scans the full directory against the ledger rather than diffing one push, a superseded queue entry can't silently drop a migration.

Noted as pre-existing, not introduced here: a multi-file run has no cross-file rollback — each migration commits individually, so a mid-run failure stops the loop and is resumable. Worth knowing operationally now that it runs automatically.

PR #508, opened 2026-08-01
ci: the migrate secret must be the session-mode pooler, not the direct URI

AndresL230:
Follow-up to #506, found by actually trying to run the migration.

## What happened

`venv/bin/dotenv -f .env.staging run -- venv/bin/python -m db.migrate` fails:

```
connection to server at "2600:1f14:131e:fd00:...", port 5432 failed:
server closed the connection unexpectedly
```

Diagnosis — not credentials, not the command:

```
db.<ref>.supabase.co  →  AAAA only, no A record
this machine          →  0 global IPv6 addresses (link-local only, despite an RA default route)
raw TCP connect       →  OSError 101, Network is unreachable
```

Canopy already recorded this from the #481 work — *"direct psycopg to staging is blocked — IPv6-only endpoint refuses"* — so it's a known wall, and I walked into it anyway.

## Why this is a defect in #506, not just an operator note

The workflow's skip notice said to use *"the DIRECT connection string, port 5432, not the pooler."* **GitHub-hosted runners have no outbound IPv6**, so its first real run would have failed identically — and the message I wrote would have sent whoever debugged it back to the same host.

## The fix

Pooler hosts publish A records (`aws-0-us-east-1.pooler.supabase.com → 44.208.221.186`), so:

- **Session mode, port 5432** — behaves like a direct connection.
- **Not 6543** (transaction mode), which drops the session-level behaviour psycopg and DDL rely on.
- Username changes to `postgres.<ref>`, easy to miss.

On `db/migrate.py`'s "SUPABASE_DB_URL, **NOT** the pooler" — I'm contradicting that deliberately rather than quietly. That warning is about *transaction* mode and predates the IPv6-only endpoint; session mode is the correct substitute, and with the direct host unreachable it's the only one. Said so in the file so the next person doesn't have to re-derive it.

Docs only — header rationale and the skip notice. No behaviour change.

## Summary by CodeRabbit

- **New Features**
  - Added a read-only migration drift report identifying pending, orphaned, conflicting, and potentially blocking database changes without modifying the database.
  - Added a command-line utility to generate Supabase session-mode connection URLs on port 5432, with safe password masking and optional raw output.

- **Documentation**
  - Clarified migration setup requirements, including the session pooler, required username and cluster format, and unsupported direct or transaction-mode connections.
  - Updated staging setup guidance, migration commands, and missing-configuration messages.

### DIFF
--- a/.github/workflows/migrate-staging.yml
+++ b/.github/workflows/migrate-staging.yml
@@ -12,6 +12,19 @@
 # separate `production` branch promotion, and auto-applying irreversible DDL to
 # prod on merge is a different risk decision. This runner has no down
 # migrations.
+#
+# THE SECRET MUST BE THE SESSION-MODE POOLER URI, NOT THE DIRECT ONE.
+# `db.<ref>.supabase.co` publishes only an AAAA record, and GitHub-hosted
+# runners have no outbound IPv6 — a direct string fails with "Network is
+# unreachable" / "server closed the connection unexpectedly" before it ever
+# authenticates. (Same wall on a home network without a global IPv6 address;
+# it is why staging migrations had to be applied by hand.)
+#
+# Use the pooler host on port 5432 — SESSION mode. Not 6543 (transaction mode),
+# which drops the session-level behaviour psycopg and DDL rely on. db/migrate.py
+# still says "NOT the pooler"; that warning is about transaction mode and
+# predates the IPv6-only endpoint. Session mode behaves like a direct
+# connection. Note the pooler also changes the username to `postgres.<ref>`.
 name: Migrate (staging)
 
 on:
@@ -56,7 +69,7 @@ jobs:
           SUPABASE_DB_URL: ${{ secrets.STAGING_SUPABASE_DB_URL }}
         run: |
           if [ -z "${SUPABASE_DB_URL}" ]; then
-            echo "::notice::STAGING_SUPABASE_DB_URL is not set — skipping. Add the secret (the DIRECT connection string, port 5432, not the pooler) to enable."
+            echo "::notice::STAGING_SUPABASE_DB_URL is not set — skipping. Add the secret to enable: the SESSION-mode pooler URI (aws-0-<region>.pooler.supabase.com:5432, user postgres.<ref>). Not the direct db.<ref>.supabase.co string — that is IPv6-only and unreachable from GitHub runners — and not port 6543, which is transaction mode."
             echo "skip=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
--- a/.github/workflows/migrate-staging.yml
+++ b/.github/workflows/migrate-staging.yml
@@ -21,10 +21,17 @@
 # it is why staging migrations had to be applied by hand.)
 #
 # Use the pooler host on port 5432 — SESSION mode. Not 6543 (transaction mode),
-# which drops the session-level behaviour psycopg and DDL rely on. db/migrate.py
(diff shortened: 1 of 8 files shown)


AndresL230:
### Code review

Found 4 issues:

1. `pooler_url.py` double-encodes the password. `urlparse().password` returns the value still percent-encoded, and `quote(..., safe="")` encodes it again — a password of `p%40ss%2Fword` becomes `p%2540ss%252Fword`, producing a URI that fails authentication with no useful error. This is the one code path in the PR, and it breaks on exactly the Supabase-generated passwords that contain reserved characters. (Raised by CodeRabbit in [discussion_r3694418319](https://github.com/SaplingLearn/Sapling/pull/508#discussion_r3694418319), still open at head.)

https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/backend/scripts/pooler_url.py#L60-L64

2. The workflow's skip-notice hardcodes `aws-0-<region>`, which `pooler_url.py` — added in this same PR — explicitly calls a guess, noting "the number is not derivable from the region — staging is `aws-1-us-west-2`". Verified against the live projects: staging resolves only on `aws-1-us-west-2`, production only on `aws-0-us-west-2`. An operator who copies this notice literally to set `STAGING_SUPABASE_DB_URL` gets `Tenant or user not found` — the failure class this PR exists to eliminate.

https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/.github/workflows/migrate-staging.yml#L71-L73

3. `db/migrate.py` still tells operators to do the thing this PR is correcting. The PR description says "Said so in the file so the next person doesn't have to re-derive it" — but the file it said so in is the workflow YAML, not the runner. The docstring still reads "the Supabase *direct* connection string (SUPABASE_DB_URL, NOT the pooler)", and `main()`'s unset-variable error routes the reader to "Connection string → Direct". The PR's own repro was running `python -m db.migrate` directly, so this is the exact path still misdocumented. Root [CLAUDE.md](https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/CLAUDE.md#L48) line 48 repeats it: "needs SUPABASE_DB_URL = direct conn string".

https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/backend/db/migrate.py#L4-L7

https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/backend/db/migrate.py#L86-L91

4. The two operator-facing docs still specify the direct connection. `docs/staging/setup-checklist.md` is precisely the document someone follows to set staging up — the task this PR unblocks — and it says "Direct connection (port 5432, not the pooler)". `README.md:236` says the same. (Also flagged by CodeRabbit in [discussion_r3694418307](https://github.com/SaplingLearn/Sapling/pull/508#discussion_r3694418307).)

https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/docs/staging/setup-checklist.md#L18-L24

https://github.com/SaplingLearn/Sapling/blob/28103e294e819110955149d63dba34eaf493894a/README.md#L235-L237

Below the reporting bar but worth noting while the file is open: the module docstring says "Three sections" while the script now prints four (`DATA BLOCKERS` was added in 28103e2), and `main()` returns 0 even when it reports orphans or data blockers — which matters because the obvious next step is having the workflow call this script instead of duplicating its logic.

PR #512, opened 2026-08-02
feat(academics): ingest real per-section Fall 2026 BU offerings (#280)

### LINKED ISSUE #280: Ingest real Fall 2026 BU course offerings (sections, instructors, meeting times, locations)
## Summary

Staging currently has a **catalog-only** academics dataset: ~8,195 real BU courses (code / name / department / description) each paired with a single **hollow** `course_offerings` row. The operational layer — **sections, instructors, meeting times, locations, syllabi, and correct term** — exists only for the 4 hand-written `seed_staging` demo classes. This issue tracks ingesting **real Fall 2026 (`fall-2026`) BU offering data** so courses behave like a real registrar.

The scrape itself will be run separately; this issue covers defining the scrape contract and building the import + supporting model/API changes.

## Current state (measured against staging)

| Table | Rows | Notes |
|---|---|---|
| `courses` | 8,195 | full BU catalog; `school_id` null on ~8,192, `credits` null on all but 3 |
| `course_offerings` | 8,196 | every course has exactly 1 offering; CS101 has 2 (the +1) |
| `terms` | 4 | `fall-2025`, `spring-2026`, `summer-2026`, **`fall-2026`** all seeded (0019) |
| `schools` | 1 | only "Sapling Demo University" (`seed-school-demo`) |
| `enrollments` | 4 | demo only |

Offering field population across all 8,196 rows:

| Field | Populated |
|---|---|
| `section` | **0** (null on every row) |
| `instructor_name` | 4 (demo only) |
| `meeting_times` | 4 (demo only) |
| `location` | 4 (demo only) |
| `syllabus_url` | **0** |

Term spread of existing offerings: `spring-2026` → 8,194, `fall-2025` → 2. **`fall-2026` has 0 offerings today.**

Relevant code:
- `backend/db/seed_staging.py` — demo seed only; never touches the real catalog.
- `backend/services/academics.py` — `resolve_offering` / `current_term` / `term_for_offering`.
- `backend/db/migrations/0020_academics_split.sql` — the catalog/offering split + `UNIQUE (course_id, term_id, section)`.
- `backend/routes/academics.py` — currently only exposes `GET /semesters`.
- All DB access goes through `backend/db/connection.py::table()` (PostgREST, no DDL).

---

### 1. Scraper / data spec (contract only — scrape run separately)
- [ ] Define the structured output schema the Fall 2026 scrape must produce (the contract the importer consumes). Proposed per-section record:
  - `course_code` (must match catalog format, e.g. `CAS CS 111` — space-separated, uppercase)
  - `section` (e.g. `A1`, `B2`)
  - `instructor_name`
  - `meeting_times` (e.g. `MWF 09:00`)
  - `location`
  - `course_name`, `credits`, `description`, `syllabus_url` (optional; used to enrich/create the catalog course if missing)
- [ ] Decide format (JSON lines / CSV) and where the scraped file lives (not committed; gitignored ops input).
- [ ] Document `course_code` normalization rules so scrape output joins cleanly to `courses.course_code`.

### 2. Importer + section model (core)
- [ ] New idempotent ops script (e.g. `backend/db/import_offerings.py`, run like `seed_staging` under the target env) that:
  - resolves each record's `course_code` → existing `courses.id` (create the catalog course if absent, enriching name/credits/description),
  - upserts one `course_offerings` row **per section** for `term_id = 'fall-2026'`,
  - writes `section`, `instructor_name`, `meeting_times`, `location`, `syllabus_url`,
  - is idempotent (deterministic id or upsert-on-UNIQUE) so re-runs add nothing.
- [ ] All access via `db/connection.py::table()`; env-agnostic (run against staging first).
- [ ] **Fix the section dedup semantics.** `UNIQUE (course_id, term_id, section)` is NULL-distinct today, so real multi-section data won't dedup. Add an append-only migration (next number, currently at 0028 → `0029`) to either set a non-null default `section` (e.g. `''`) or switch to `UNIQUE ... NULLS NOT DISTINCT` (PG 15+). Pick one and make the importer consistent with it.
- [ ] Tests in `backend/tests/` (mirror `test_seed_staging.py`): idempotency, code→course resolution, per-section row creation, dedup on re-run.

### 3. BU school + catalog linkage
- [ ] Add a real `schools` row for **Boston University** (proper `name`/`slug`).
- [ ] Link catalog `courses` to it (`school_id` is null on ~8,192 rows today).
- [ ] **Pre-check for duplicate `course_code`s** before linking — `UNIQUE (school_id, course_code)` is currently NULL-distinct, so dup codes may exist that would collide once `school_id` is set. Resolve/merge duplicates as part of this task.

### 4. API + frontend display
- [ ] Audit the offering read path and surface `section` / `instructor_name` / `meeting_times` / `location` / `syllabus_url` through the course/offering API (today `routes/academics.py` only returns `/semesters`).
- [ ] Update the relevant frontend components so a course with a Fall 2026 offering shows its section, instructor, meeting times, and location.

---

## Notes / gotchas
- `fall-2026` runs 2026-08-24 → 2027-01-03; the term row already exists, so the importer just references it by id.
- `current_term()` is date-derived — today (2026-06-26) resolves to **Summer 2026** (0 offerings). The importer should target `fall-2026` **explicitly**, not "current term".
- The knowledge graph keys on the abstract `course_id` (cumulative across terms); gradebook on `enrollment_id`; study/analytics on `offering_id`. Adding Fall 2026 offerings must not fork a course's graph identity — resolve to the existing catalog `course_id`.
- Staging and prod both currently hold only catalog data (no real user data in prod per the redesign). Run/verify on staging first.

Darkest-Teddy:
Refs #280 (tasks 1–3). **Does not close #280** — task 4 (API + frontend read path, letting a student pick the section they actually registered for) is deliberately deferred, so the issue must stay open after this merges.

## Why

`course_offerings` has been hollow since the 0020 academics split. Staging held 12,318 rows, of which 4 (the demo seed) carried a section and **none** carried a meeting time or location. Courses could not behave like a registrar.

This ingests the operational layer for `fall-2026` from bu.edu: one offering per published section, carrying `section` / `instructor_name` / `meeting_times` / `location`.

## What changed

**Scraper (task 1)** — `_extract_schedule` flattened each course's schedule tables down to a semester list, discarding exactly the fields #280 needs. Replaced with `_extract_sections` + `_derive_schedule`:

- Columns are looked up **by table header name, not position**, so a column BU inserts doesn't silently shift the data.
- `Staff`/`TBA` instructors and `NO ROOM` locations become `NULL` rather than being stored literally and shown to students.
- New `--rescan` mode refetches only the URLs already in the JSON. Sections were never stored, so recovering them required a re-fetch; skipping the listing walk preserves the index-orphan courses that only `probe_unscraped_courses.py` finds.

**Importer (task 2)** — `backend/db/import_offerings.py`, dry-run by default.

It **adopts rather than re-creates**. Staging already had 4,122 hollow fall-2026 offerings with enrollments, sessions, documents and notes pointing at them, and `course_offerings` is referenced by 8 tables with mixed `CASCADE` / `SET NULL` semantics — a delete-and-reinsert import would have destroyed or silently orphaned real rows. So for each course the first published section takes over the existing section-less row **in place, id preserved**, and only the remaining sections are inserted. Re-running matches every section by code and writes nothing.

Offerings the scrape no longer lists are **reported, never deleted**. A partial or failed scrape must not be able to wipe a term.

BU publishes **multiple meeting patterns per section** — CAS MA 123 A1 is an MWF 9:05 lecture *and* a Thursday 6:30 exam block. A keep-first dedup would have dropped a real meeting from 308 sections, so `_merge_meetings` joins them (`"R 6:30 pm-8:30 pm; MWF 9:05 am-9:55 am"`). Verified across the whole scrape that no multi-meeting section has two different instructors.

**School linkage (task 3)** — `--link-school` creates the Boston University `schools` row and links courses to it, after a duplicate-`course_code` precheck that **refuses outright rather than half-linking**. All 8,510 courses turned out to be unique, so the merge work #280 anticipated wasn't needed.

**Determinism fix** — with one offering per section, `resolve_offering`'s `order="created_at.asc", limit=1` was non-deterministic: sections land in a single batch insert and share a `created_at`, so the winner was the planner's choice. Two calls could split one user's documents and notes across sections. Now ordered by `section.asc` first; `''` sorts before `A1`, preserving pre-#280 behaviour wherever a hollow offering remains.

## No migration in this PR

The `section NOT NULL DEFAULT ''` change this depends on is **already on main** as `0033_offering_section_not_null.sql` (recovered by #510 — it reached staging out-of-band during this work), and 0036's now-unreachable NULL-section partial index was dropped by `20260801062439_drop_dead_null_section_index.sql`. Reasoning for choosing the non-null default over `NULLS NOT DISTINCT` is recorded in the ADR.

## Separable

`backend/.env.example` and `backend/.env.staging.example` still handed out the direct `db.<ref>.supabase.co` host, which is IPv6-only and is the reason applying migrations failed locally. `migrate.py` and `CLAUDE.md` were already fixed upstream; these two templates were missed. Happy to split this into its own PR if preferred.

## Out of scope

Task 4 (API + frontend read path, letting a student pick the section they actually registered for) is not here. Today a student enrolling in CAS CS 330 lands in A1 regardless — this PR makes that choice *stable*, not *correct*.

## Verification

Staging, after applying:

- 11,079 fall-2026 offerings — 10,958 sectioned, 121 left hollow for courses BU dropped from the schedule
- coverage: `meeting_times` 100%, `instructor_name` 92%, `location` 59%
- 0 duplicate `(course, term, section)`; 0 orphaned enrollments / sessions / documents / notes
- BU school linked to 8,507 courses, demo's 3 untouched
- re-run reports "Nothing to do — already in sync"
- spot-checked against bu.edu: CAS CS 330 = 7 sections; CAS MA 123 A1 in `STO B50`

Tests: 53 targeted tests pass (`test_scrape_bu_catalog.py` 12, `test_import_offerings.py` 20 against a hermetic PostgREST fake, plus academics). Full backend suite: **1,585 passed, 41 skipped**. Ruff clean on every touched file.

## Unaffected by design

- **Knowledge graph** keys on the abstract `courses.id`, and the import resolves each `course_code` to the existing catalog row rather than creating a parallel course — sections cannot fork a course's graph identity.
- **RAG** keys `course_chunks` on `course_code`, not `offering_id`, so the shared-course-corpus property survives sections.
- **Gradebook** keys on `enrollment_id` and study analytics on `offering_id`; both keep working because adoption preserves offering ids.

## Summary by CodeRabbit

* **New Features**
  * Course catalogs now include detailed section schedules, instructors, meeting times, locations, and notes.
  * Added catalog rescan and refresh options to update course information while preserving failed records and backups.
  * Added tools to identify and investigate courses missing from catalog data.

* **Bug Fixes**
  * Improved offering selection for consistent results, prioritizing section information.

* **Documentation**
  * Added guidance for importing, validating, and verifying catalog data.


Darkest-Teddy:
@coderabbitai review

Darkest-Teddy:
@coderabbitai full review

Jose-Gael-Cruz-Lopez:
## Code review — per-section Fall 2026 BU offerings

This lands the operational layer for `course_offerings`: a header-driven section parser in `scrape_bu_catalog.py`, a `--rescan` mode to re-derive `sections` from already-crawled URLs, a dry-run-by-default importer (`db/import_offerings.py`) that **adopts** hollow section-less rows in place instead of re-creating them, a structural + live verifier, an index-orphan prober, and a determinism fix in `resolve_offering`. The shape is right and the risky decisions are the correct ones — adoption over delete-and-reinsert, report-never-delete for stale offerings, and no migration of its own (it rides `0033_offering_section_not_null.sql` + `20260801062439_drop_dead_null_section_index.sql`, both confirmed already on `main`). I verified the two things that would hurt most and both hold:

- **Idempotency: correct.** The conflict key the importer keys on is `(course_id, term_id, section)` — the same tuple as `course_offerings_unique` from `0020_academics_split.sql` — resolved client-side via `by_section = {(r.get("section") or ""): r for r in rows}` (`import_offerings.py:352`). A second run matches every section by code, `patch` comes out empty, and nothing is written. `None` and `''` both normalise to `''`, so pre- and post-0033 rows behave identically, and `hollow` can only ever hold one row per (course, term) once 0033 is applied. A crash mid-write is retryable: adoption is matched by section code, not by position.
- **Migration back-compat: not applicable, and correctly so.** No new migration; no NOT NULL add, no unique index over live data, no destructive drop. The dependencies are already on `main` and both carry guards.

Findings below are things I could not talk myself out of. Nothing rises to a data-loss or security issue.

### Findings

**[P1] `--link-school` is silently dropped when offerings are already in sync** — `backend/db/import_offerings.py:408-410`

```python
    if not to_insert and not to_update:
        print("\nNothing to do — already in sync.")
        return
```

This `return` fires before `link_school()` is ever reached (`import_offerings.py:428-429` for apply, `413-414` for dry run). The documented runbook command in this PR — `python -m db.import_offerings --apply --create-missing --link-school` (`CLAUDE.md:54`) — therefore does nothing for #280 task 3 on any re-run, and prints "Nothing to do" without mentioning the flag it skipped. This is exactly the retry path that matters: `link_school` deliberately bails out with `[FAIL]` on duplicate `course_code`s (`import_offerings.py:260-268`), so the intended recovery is "merge the dupes, re-run" — and after the offering sync has landed, that re-run silently links nothing. There is no `--link-school`-only entry point either. The early return should be gated on `not do_link_school`, or the linking step hoisted above it.

**[P2] Registrar placeholder schedules are stored verbatim in `meeting_times`** — `backend/db/import_offerings.py:130-132,157-163`

```python
_NO_LOCATION = {"no room", "tba", "tbd", "arr", ""}
...
        t = (s.get("meeting_times") or "").strip()
        if t and t not in times:
            times.append(t)
        loc = (s.get("location") or "").strip()
        if loc and loc.lower() not in _NO_LOCATION and loc not in locations:
            locations.append(loc)
```

`instructor_name` is normalised at scrape time (`_NO_INSTRUCTOR`, `scrape_bu_catalog.py:202-204`) and `location` at import time (`_NO_LOCATION`), but `meeting_times` gets no placeholder handling on either side — `scrape_bu_catalog.py:264` is a bare `"meeting_times": cell("schedule") or None`. This PR's own fixtures show the consequence: `test_scrape_bu_catalog.py:116-127` models a real arranged section as `<td>ARR 12:00 am-12:00 am</td>` in the Schedule column, and `test_import_offerings.py::test_placeholder_locations_become_null` feeds `("A1", "Erdos", "ARR", "NO ROOM")` and asserts only `location is None` — the resulting row keeps `meeting_times == "ARR"`. So an ARR section lands with `location = NULL` (honest) and `meeting_times = "ARR 12:00 am-12:00 am"` (a placeholder shown to students), which is precisely the `Staff`/`NO ROOM` failure mode the rest of the PR is careful to avoid. It also inflates the PR's headline metric: `verify_catalog_scrape.py:163-165` counts any truthy `meeting_times` as coverage and its `[FAIL]` placeholder check at `:168-174` covers instructors only, so "meeting_times 100%" cannot distinguish a real schedule from `ARR`. Applying the same placeholder set (or an `ARR`-aware one) to `meeting_times` in `_merge_meetings`, plus a matching `[FAIL]` in `_sections`, closes both halves.

**[P2] Updates and adoptions are one HTTP PATCH per row while inserts are batched** — `backend/db/import_offerings.py:417-424`

```python
    for i in range(0, len(to_insert), BATCH):
        table("course_offerings").insert(to_insert[i : i + BATCH])
        print(f"  inserted {min(i + BATCH, len(to_insert)):,}/{len(to_insert):,}", flush=True)

    for n, (offering_id, patch) in enumerate(to_update, 1):
        table("course_offerings").update(patch, filters={"id": f"eq.{offering_id}"})
```

`BATCH = 500` is honoured for `to_insert` and for `link_school`'s `in.(...)` windows, but the update path is a per-row round trip. The PR reports 4,122 hollow fall-2026 offerings on staging, every one of which goes through this loop as an adoption, plus one PATCH per section whose instructor/room/time the registrar changed on any later re-import — so a routine refresh after BU republishes is O(sections) sequential requests, each its own transaction and each firing `trg_course_offerings_updated_at`. `SupabaseTable.upsert(..., on_conflict="id")` already exists in `db/connection.py` and merges duplicates, which would let these be written in the same 500-row batches as the inserts.

**[P2] `resolve_offering`'s cross-term fallback is still unordered, and this PR is what makes that bite** — `backend/services/academics.py:169-174`

```python
    # No offering in the target term and not creating — fall back to any offering
    # of this course so reads still resolve to something sensible.
    any_off = table("course_offerings").select(
        "id", filters={"course_id": f"eq.{course_id}"}, limit=1
    )
    return any_off[0]["id"] if any_off else None
```

ADR 0026 Decision 3 and the fix at `:133-138` are right, but they only cover the term-filtered query. This `limit=1` with no `order` picks an arbitrary row, and the PR is what turns "arbitrary among 1" into "arbitrary among 7" for every scraped course. It is not reachable today for those courses because `current_term()` resolves to `fall-2026`, but it becomes live the moment a later term row is seeded (as `0032_retire_summer_2026.sql` shows terms do get maintained): every `resolve_offering(course_id)` fallback caller — `routes/study_guide.py:52`, `routes/notes.py:156`, `routes/flashcards.py:144,408` — can then hand the same user a different section on consecutive calls, which is the exact documents/notes split the ADR argues against. Same `order="section.asc,created_at.asc"` applies.

**[P3] The 409 race re-select still orders by `created_at.asc` alone** — `backend/services/academics.py:156-161`

```python
            rows = table("course_offerings").select(
                "id",
                filters={"course_id": f"eq.{course_id}", "term_id": f"eq.{term_id}"},
                order="created_at.asc",
                limit=1,
            )
```

Sibling of the query fixed 20 lines above, same function, same tuple, left on the ordering the comment at `:126-132` explains is planner-dependent. Benign in the common race (only the `''` row exists at that point), but it is the one path that can return a row the steady-state reader would not, and it now contradicts the new test's assertion that section is the primary sort.

**[P3] `_all_rows` silently returns one page if the row count can't be read** — `backend/db/import_offerings.py:92-99`

```python
    while True:
        rows, total = table(name).select_with_count(
            columns, filters=filters, order="id.asc", limit=PAGE, offset=offset
        )
        out.extend(rows)
        offset += len(rows)
        if not rows or offset >= total:
            return out
```

`SupabaseTable.select_with_count` returns `total = 0` whenever `Content-Range` is missing or unparseable (`db/connection.py`), so the guard becomes `1000 >= 0` and the helper returns a truncated view of `courses` / `course_offerings` without a word. Truncating `existing` is the dangerous direction: the importer would queue inserts for sections that already exist and take a duplicate-key 409 mid-batch. Terminating on a short page (`len(rows) < PAGE`) and treating `total` as advisory removes the dependency. `backend/scripts/probe_unscraped_courses.py:70-80` carries the same copy.

### What's good

- Adoption-in-place with the id preserved, and "stale offerings are reported, never deleted", are the two calls that decide whether this is safe to run against a database with live enrollments — both are argued in ADR 0026 and pinned by tests (`test_adopts_hollow_offering_in_place`, `test_stale_offerings_are_left_alone`, `test_other_terms_are_untouched`).
- Header-driven column lookup with the first-`<th>`-row rule, and the `FETCH_FAILED` sentinel that stops a transient listing timeout from reading as end-of-pagination, both come with regression tests that name the exact bug they prevent.
- Trust boundary is clean: `import_offerings` / `probe_unscraped_courses` / `verify_catalog_scrape` are CLI-only (`git grep` at head finds no import outside `CLAUDE.md`), all DB access goes through `db/connection.py::table()` per the Engineering Style Guide, and the `.env` templates now match the session-pooler guidance in the Infrastructure doc (`postgres.<ref>@...pooler...:5432`).

**Verdict:** request changes — the `--link-school` early return (P1) makes the documented command silently incomplete on any re-run, and the `meeting_times` placeholder gap (P2) puts registrar filler in a student-facing column. Everything else is small. Idempotency and migration back-compatibility both check out.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Major
- **`--link-school` was silently dropped on every re-run.** The `Nothing to do — already in sync` early return fired before `link_school()`, so the runbook command documented in `CLAUDE.md` did nothing for task 3 once offerings were synced — and that is precisely the retry path that matters, since `link_school` bails on duplicate course codes expecting a merge-and-re-run. Now gated on the flag, with an honest message. Three tests, including that a dry run still does not create the school row.

### Minor
- **Registrar placeholder schedules were stored verbatim** in a student-facing column while instructor and location were both normalised — an ARR section landed with an honest `NULL` location and `"ARR 12:00 am-12:00 am"` as its meeting time. Nulled on both sides with a predicate that keeps a real pattern following ARR, and the verifier now excludes placeholders from coverage instead of counting them (which had inflated the "100%" figure) and flags them.
- **Updates were one PATCH per row while inserts batched at 500** — 4,122 sequential requests plus trigger fires for the staging adoption pass. Now `upsert(on_conflict="id")` in the same 500-row windows; rows are sent whole because Postgres checks NOT NULL before detecting the conflict.
- `resolve_offering`'s cross-term fallback is ordered (this PR is what turned "arbitrary among 1" into "arbitrary among 7").

### Nits
409-race re-select ordered consistently · `_all_rows` terminates on a short page instead of silently truncating when `Content-Range` is unreadable (both copies) · multi-instructor warning · term-scoped fetch · shared PostgREST fake centralised in `conftest.py` · markdownlint fixes · `main()` return value and dual-import style.

### Also
The PR body said `Closes #280`; task 4 is deferred, so it now reads `Refs #280` and will no longer auto-close the issue.

**Verification** — `ruff check .` clean · **1569 passed, 38 skipped**

Jose-Gael-Cruz-Lopez:

### This is the PR where the database check matters most

It writes ~11k rows and **adopts 4,122 existing hollow offerings in place**, keeping their ids so enrollments, sessions, documents and notes stay attached. Nothing here can be judged from the repo alone.

### 1. Always dry-run first — it is the default

```
cd backend && python -m db.import_offerings
```

Compare against the numbers in the PR body: 11,079 fall-2026 offerings, 10,958 sectioned, 121 left hollow. A materially different count means the scrape or the term changed underneath it.

### 2. The placeholder change alters rows that are ALREADY stored

The fix nulls registrar filler (`ARR`, `TBA`, `TBD`) in `meeting_times`, matching how `instructor_name` and `location` were already handled. Staging rows imported before this change still hold values like `'ARR 12:00 am-12:00 am'`.

```sql
SELECT count(*) FROM course_offerings
WHERE term_id = 'fall-2026' AND meeting_times ~* '^(ARR|TBA|TBD)';
```

**Confirm a re-run actually queues these as updates rather than reporting "already in sync".** If the count is non-zero before the run and unchanged after, the reconciliation is comparing post-normalisation values on both sides and silently skipping them — please check before trusting the coverage figure, since `verify_catalog_scrape` now excludes placeholders from `meeting_times` coverage and the previously reported "100%" included them.

### 3. `--link-school` on a re-run — the P1 that was fixed

The early return used to skip linking entirely once offerings were in sync, so the documented command was a no-op on every retry.

```sql
SELECT count(*) FROM courses WHERE school_id IS NOT NULL;   -- expect ~8,510
SELECT count(*) FROM schools WHERE slug = 'boston-university';
```

### 4. Batched upserts must not have duplicated anything

Per-row PATCHes became `upsert(on_conflict="id")` in 500-row windows, sending whole rows.

```sql
SELECT course_id, term_id, section, count(*)
FROM course_offerings GROUP BY 1,2,3 HAVING count(*) > 1;   -- expect 0 rows

SELECT count(*) FROM enrollments e
WHERE NOT EXISTS (SELECT 1 FROM course_offerings o WHERE o.id = e.offering_id);  -- expect 0
```

### 5. Re-run idempotency

A second `--apply` must report nothing to do and write nothing. That is the property that makes this safe to retry.

PR #530, opened 2026-08-06
feat(graph-3d): Focused Minimal visual upgrade — labels, hover focus, recenter

AndresL230:
## Summary
- 3D graph mode grows from bare spheres into the approved Focused Minimal design (spec: docs/superpowers/specs/2026-08-05-3d-graph-focused-minimal-design.md)
- matte spheres + always-visible SpriteText labels (roots bold), unexplored tier washed toward warm gray
- hover-focus: 1-hop neighborhood lit, rest dimmed; sage halo on hovered + tutor-highlighted nodes
- ⌖ recenter (zoomToFit) sharing the 2D control's title/testid (auto-hidden on tutor rail)
- new dep three-spritetext, client-chunk only; 3D stays opt-in behind the existing toggle
- camera auto-fit on engine settle (bbox-stabilization poll + warmupTicks under test/reduced-motion) — 3 review-gated fix rounds, root-caused in a live browser

## Test plan
- [x] graph3dHelpers unit tests + rewritten KnowledgeGraph3D component tests
- [x] full frontend suite, lint, next build
- [x] flock'd E2E cycle: Chapter 1 journeys + oracles green
- [x] manual visual pass on /tree, dashboard, tutor sidebar (screenshots attached)
- [x] fresh 3D screenshots reviewed on /tree + dashboard (visual-pass artifacts in .superpowers/sdd/, gitignored)

## Summary by CodeRabbit

* **New Features**
  * Added a focused visual treatment for the opt-in 3D knowledge graph, including matte nodes, halos, and always-visible labels.
  * Hovering over a node now highlights its immediate connections while dimming unrelated nodes and links.
  * Added a “Reset view” control to recenter and fit the graph.
  * Improved automatic camera fitting when graph data changes, with reduced-motion support.

* **Tests**
  * Expanded coverage for rendering, highlighting, camera fitting, accessibility behavior, and visual styling.


AndresL230:
### Code review

Jose-Gael-Cruz-Lopez:
## Code review — graph-3d Focused Minimal upgrade

This turns the 3D mode from bare library spheres into the Focused Minimal design: `nodeThreeObject` groups (matte `MeshLambertMaterial` sphere + hidden accent halo + always-visible `three-spritetext` label), a `visualsRef` registry so hover-focus mutates materials instead of rebuilding geometry, `linkColor`/`linkWidth` re-keyed on hover, a `⟲` recenter button reusing the 2D control's `title="Reset view"` (which is what `globals.css:609` keys on to hide it on the tutor rail), and a bbox-stabilisation rAF poll that drives a one-shot `zoomToFit` on the first engine settle. The pure-helper extraction into `graph3dHelpers.ts` is clean, the auto-fit poll's epoch guard and rAF-cancel-on-unmount are correct, and the SSR/testid/click-whitelist contracts are all preserved. Two things I checked and want to save you the time on: the library **does** own GPU disposal (`react-kapsule` calls `_destructor` on unmount → `graphData({nodes:[],links:[]})` → `_deallocate` recursively disposes each group's geometries, materials and the sprite's `material.map`), and `zoomToFit` is null-safe on an empty graph — so CodeRabbit's disposal comment is not a real leak. What does need fixing is the hover path's interaction with `three-forcegraph`'s prop-change semantics, plus the new lint suppressions and the merge state.

### Findings

#### P1

**[P1] Every hover transition destroys and rebuilds every link mesh** — `frontend/src/components/graph/KnowledgeGraph3D.tsx:326-335`

```tsx
  const linkWidth = React.useCallback(
    (l: object) => {
      const link = l as FG3DLink;
      const base = 0.4 + (link.strength || 0) * 0.6;
      if (!hoverId) return base;
      const lit = linkEndId(link.source) === hoverId || linkEndId(link.target) === hoverId;
      return lit ? base + 0.6 : base;
    },
    [hoverId],
  );
```

`hoverId` is React state, so each hover enter/leave re-renders and hands `react-force-graph-3d` a new `linkWidth` function identity. `react-kapsule` pushes any identity-changed prop straight through (`react-kapsule.mjs:30-33`; `linkWidth` is a plain prop per `react-force-graph-3d.mjs:126-129`), and `three-forcegraph` treats `linkWidth` as an object-invalidating prop: `three-forcegraph.mjs:1199-1201` — `if (state._flushObjects || hasAnyPropChanged(['linkThreeObject','linkThreeObjectExtend','linkWidth'])) state.linkDataMapper.clear();`. `clear()` is `digest([])`, which `scene.remove()`s and `_deallocate()`s (geometry.dispose + material.dispose) **every** link object and then recreates all of them. These aren't cheap lines either — `useCylinder = !!widthAccessor(link)` (`:1220`) is always true here, so each link is a `CylinderGeometry` mesh. On a few-hundred-edge graph that's a full teardown/rebuild of the link layer on every mouse enter and every mouse leave. This is exactly what `docs/superpowers/specs/2026-08-05-3d-graph-focused-minimal-design.md:138-142` promised not to do: *"Hover mechanics — no per-hover geometry rebuilds … (links are cheap line materials)."* Direction: give `linkWidth` a stable identity that reads `hoverRef.current`; `linkColor`'s own change already triggers the digest, and the link `onUpdateObj` re-reads `widthAccessor(link)` each digest, so widths still update — just without the `clear()`. (Nothing visually breaks — `three-forcegraph.mjs:1483` sets `engineRunning = true` at the end of every `update()`, so the next `layoutTick` repositions the new meshes — but it does mean `onEngineStop` also re-fires on every hover; the `didFitRef` guard absorbs that correctly.)

**[P1] Two new `react-hooks/refs` suppressions added for new production code** — `frontend/eslint-suppressions.json:99-103`

```json
  "src/components/graph/KnowledgeGraph3D.tsx": {
    "react-hooks/refs": {
      "count": 2
    }
  },
```

The Canopy **Engineering Style Guide §7** states the ratchet explicitly — *"don't fix the whole backlog to land a change, but never add to it"* — and this PR's own plan repeats it at `docs/superpowers/plans/2026-08-05-3d-graph-focused-minimal.md:20`: *"eslint suppressions are a ratcheted baseline — new code must be clean."* The two suppressed violations are real render-phase ref mutations: `KnowledgeGraph3D.tsx:173` `highlightRef.current = highlightId;` in the render body, and `:179` `visualsRef.current = new Map();` inside the `graphData` `useMemo`. The `useMemo` one is the one with teeth — under React 19 concurrent rendering a render that starts and is then discarded still clears the registry, leaving the committed tree with an empty `visualsRef` and hover-focus a silent no-op. (The test-file suppressions — `react-hooks/globals`, `react/display-name` — are less load-bearing but also new.)

**[P1] Branch is `CONFLICTING` with `main`, and `main` landed the #538 WebGL2 gate on the same three files**

`gh pr view 530 --json mergeable` returns `"mergeable":"CONFLICTING","mergeStateStatus":"DIRTY"`. Since the merge-base (`ec34bf17`), `main` has changed `KnowledgeGraph.tsx` (+201: WebGL2 capability probe, `ErrorBoundary`, `graph-crash-fallback`), added this header to `KnowledgeGraph3D.tsx` — *"#538: NEVER mount this component outside the `KnowledgeGraph` wrapper — three r163+ throws from the WebGLRenderer constructor when WebGL2 is unavailable"* — and replaced the local `next/dynamic` mock in **both** `KnowledgeGraph3D.test.tsx` and `KnowledgeGraph3D.testmode.test.tsx` with the shared `@/test-utils/mockNextDynamic` helper. This PR rewrites all three of those files from the pre-#538 base and still carries the old inline mock (`KnowledgeGraph3D.test.tsx:98-114`, `KnowledgeGraph3D.testmode.test.tsx:36-52`), so the conflict resolution has to re-apply main's work by hand rather than take either side wholesale. Worth rebasing before the draft flips to ready — main's shared helper does handle this PR's loader (which resolves to a bare function component, not a module) via `typeof mod === "function" ? mod : (mod.default ?? null)`.

#### P2

**[P2] `nodeThreeObject` doesn't apply the active focus state to the nodes it (re)builds** — `frontend/src/components/graph/KnowledgeGraph3D.tsx:251-286`

```tsx
      const sphereMat = new THREE.MeshLambertMaterial({
        color,
        transparent: true,
        opacity: NODE_OPACITY,
      });
```

The halo correctly reads the live refs (`halo.visible = n.id === hoverRef.current || n.id === highlightRef.current`), but the sphere is always built at full `baseColor`/`NODE_OPACITY` and the label at full opacity. Any `nodes`/`edges` identity change re-runs the `graphData` memo (clearing `visualsRef` at `:179`) and makes `three-forcegraph` rebuild every node object — node objects are keyed by identity and the memo mints fresh `{...n}` clones. If the pointer is still over a node at that moment (a tutor `graph_update` refreshing the Learn rail graph, a filter change on `/tree`), the hovered node keeps its halo but nothing dims: the focus effect renders half-applied until the pointer moves. The `[applyFocus, highlightId]` re-assert effect can't cover it because it runs at commit, before the library rebuilds. The component test doesn't catch this because it calls `nodeThreeObject` manually *before* driving `onNodeHover` (`KnowledgeGraph3D.test.tsx:451-456`), which is the opposite order.

**[P2] A halo geometry + material is allocated for every node although at most two are ever visible** — `frontend/src/components/graph/KnowledgeGraph3D.tsx:256-270`

```tsx
      group.add(new THREE.Mesh(new THREE.SphereGeometry(r, 24, 24), sphereMat));

      // Focus halo: slightly larger translucent accent sphere, hidden
      // until this node is hovered or is the persistent highlightId.
      const halo = new THREE.Mesh(
        new THREE.SphereGeometry(r * 1.4, 16, 16),
        new THREE.MeshBasicMaterial({
```

`halo.visible` is only ever true for the hovered node and `highlightId`, so an N-node graph carries N−2 unused `SphereGeometry` + `MeshBasicMaterial` pairs for the lifetime of the dataset, re-allocated in full on every refresh. A single shared unit-sphere geometry scaled per node, or lazy creation on first focus, avoids it. Separately, the main sphere went from the library's `nodeResolution={16}` (pre-patch) to `SphereGeometry(r, 24, 24)` — 2.25× the triangles per node — which the spec doesn't ask for.

**[P2] Each label is rasterised five times at construction** — `frontend/src/components/graph/KnowledgeGraph3D.tsx:273-279`

```tsx
      const label = new SpriteText(n.name);
      label.textHeight = spec.textHeight;
      label.fontWeight = spec.fontWeight;
      label.color = theme.ink;
      label.fontFace = '"JetBrains Mono", monospace';
```

Every `three-spritetext` setter re-runs `_genCanvas()` (measure text → resize canvas → repaint → `new CanvasTexture`). That's the constructor plus four setters = five full rasterisations and five texture allocations per node, paid again for the whole graph on every dataset refresh. `SpriteText`'s constructor takes `(text, textHeight, color)`, which folds three of the five. Related sizing note: `_fontSize` is fixed at 90 regardless of `textHeight`, so each label canvas is roughly 600×100 RGBA — at the spec's stated "few hundred nodes" ceiling that is tens of MB of texture memory. (No leak — `_genCanvas` disposes the previous map — just wasted work and VRAM.)

#### P3

**[P3] The base and dim link colors are still hardcoded hex while the lit one derives from the theme** — `frontend/src/components/graph/KnowledgeGraph3D.tsx:315,321`

```tsx
      if (!hoverId) return `rgba(138, 131, 114, ${BASE_LINK_ALPHA})`;
      ...
        : `rgba(138, 131, 114, ${DIM_LINK_ALPHA})`;
```

`138, 131, 114` is `#8a8372` = `--ink-400` (`globals.css:47`) — an existing token, not an unnamed value. `graph3dHelpers.ts:33-36` says `FALLBACK_THEME` is *"the ONE place that hex is allowed to be hardcoded"*, and the c65c075 pass fixed only the lit half. Adding a fourth `GraphTheme` field resolved from `--ink-400` would close the loop and match the Canopy design-token rule ("Use the tokens; don't hardcode values"). Low stakes — the spec does sanction "today's `rgba(138,131,114,…)` family" — but the file's own comment now contradicts itself.

### Test-plan checklist — per-item verdict

1. **graph3dHelpers unit tests + rewritten component tests** — supported. `graph3dHelpers.test.ts` covers `buildAdjacency`, `nodeVal`/`nodeRadius`, `mixHex`, `baseNodeColor`, `labelSpec`, `resolveGraphTheme`; the component test covers node composition, hover dim/restore, halo persistence, `linkColor`, recenter (both durations), the bbox poll, the frame cap and the epoch guard. Two gaps: the newly exported `hexToRgbTriplet` has no unit test (only indirect coverage via the component test), and nothing asserts the unmount `cancelAnimationFrame` even though `captureAnimationFrames` already spies on it.
2. **full frontend suite, lint, next build** — supported by CI (`Frontend (lint + tsc + vitest)` pass, `Workers Builds: frontend-staging` pass), with the caveat that lint only passes because of the new suppressions in the P1 finding above.
3. **flock'd E2E cycle: Chapter 1 journeys + oracles green** — not verifiable from the PR; there is no E2E job among the checks on #530. Taking it on trust.
4. **manual visual pass on /tree, dashboard, tutor sidebar (screenshots attached)** — the code claims hold up (the `title="Reset view"` ↔ `globals.css:609` rail-hiding seam is real), but no screenshots are attached to the PR body or to any comment, so "attached" doesn't match the PR as it stands.
5. **fresh 3D screenshots reviewed on /tree + dashboard** — artifacts are gitignored, so unverifiable from here. Note that the P1 hover finding wouldn't have shown up as a visual defect (it's allocation churn, not a broken frame), so this doesn't invalidate the pass.

### What's good

- The auto-fit story is genuinely well-engineered: the bbox-stabilisation poll, the `MAX_FRAMES` safety net, the `pollEpochRef` invalidation on mid-poll dataset change and the rAF cancel on unmount are all correct, and each one is pinned by a test that would fail if it regressed.
- Reduced-motion is handled properly rather than performatively — zeroing the `zoomToFit` duration on both call sites because a system-initiated camera fly is itself motion, and the `warmupTicks={200}` reasoning mirrors `KnowledgeGraph2D`'s `sim.tick(200)` precedent instead of inventing a new one.
- `graph3dHelpers.ts` is the right shape: pure, three.js-free, unit-testable, and the "always return `#rrggbb`" color contract is documented where it will actually be read.
- Every preserved contract really is preserved — sr-only list, all three testids, the `onNodeClick` id whitelist, and the `dynamic(ssr:false)` boundary all still hold, and the click-whitelist test now enumerates the full set of library-injected fields.

Verdict: solid design work with one real hover-path performance bug (P1), a ratchet violation (P1) and a rebase (P1) to clear before this leaves draft.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Major
- **Every hover destroyed and rebuilt every link mesh.** `linkWidth` was keyed on `hoverId` state, so each enter/leave handed the library a new function identity, and `three-forcegraph` treats `linkWidth` as object-invalidating → `linkDataMapper.clear()` → full teardown/rebuild of every `CylinderGeometry` link. That is exactly what the spec promised not to do. `linkWidth` now has a stable identity reading `hoverRef.current`; widths still track hover because `linkColor` already triggers the digest and `onUpdateObj` re-reads the accessor. Mutation-verified: restoring `[hoverId]` fails exactly one test.
- **Two new `react-hooks/refs` suppressions were added for new code**, against the ratchet rule this PR's own plan restates. Both fixed at the source — the render-body ref write moved into an effect, and `visualsRef` replaced by a dataset-scoped epoch object so a discarded concurrent render cannot clobber the committed registry. The test-file suppressions were fixed properly too, not baselined. Suppressions went **29 files/193 → 28 files/187, deletion-only**.
- **The branch was CONFLICTING with `main`** (#538 landed the WebGL2 gate on the same three files). Ported forward by hand — no rebase: adopted `@/test-utils/mockNextDynamic` byte-identical to main, re-applied the #538 header, and shrank two more noise conflicts. `git merge-tree --write-tree HEAD main` now reports **0 conflicts in all three source files**; the only residual is the generated `eslint-suppressions.json` baseline.

### Minor / nits
`nodeThreeObject` applies the live focus state to nodes it rebuilds, so a mid-hover dataset refresh no longer renders focus half-applied · shared halo geometry instead of one per node, sphere segments back to 16 · `SpriteText` built via its constructor args, cutting five rasterisations per label · base/dim link colours derived from the theme rather than a hardcoded `--ink-400` · e2e teardown `trap` in the plan doc · deterministic frame-cap test that pins `MAX_FRAMES` · unmount `cancelAnimationFrame` test · `hexToRgbTriplet` unit test.

### Not done — CodeRabbit's disposal finding is not valid
`react-kapsule` calls `_destructor` on unmount → `graphData({nodes:[],links:[]})` → `_deallocate`, which already disposes each group's geometries, materials and the sprite's `material.map`. Adding manual disposal would be redundant.

**Verification** — `tsc` clean · eslint **0 errors** · **604 frontend tests pass** · merge conflicts vs `main`: 3 source files → 0

Issue #125, opened 2026-06-08
[P1] search_course_materials leaks other users' documents into the tutor/note-chat LLM


AndresL230:
**Severity: HIGH** · Source: `docs/backend-contract-bug-audit.md` (finding #3)

`backend/agents/tools/chat_context.py:165-181,234-245` · `chat_tutor.py:102-107` · `note_chat.py:48-52`

`search_course_materials` filters the `documents` table on `course_id` only, but documents are user-scoped within a shared course. It returns and decrypts `summary`/`concept_notes` for **every** enrolled user, feeding another student's private content into the requester's LLM context and answers. `ctx.deps.user_id` is available but unused.

### Fix
Thread `user_id` into the tool and add `"user_id": f"eq.{user_id}"` to the filters, matching the scoping used everywhere else.

PR #534, opened 2026-08-11
fix(tutor): repair course-material retrieval, silence course-scope commentary, make quizzes practical

Darkest-Teddy:
Stacked on #533 — review that first; this PR's base is its branch, so the diff here is only the new work.

## 1. `search_course_materials` had never worked

`agents/tools/chat_context.py` filtered `documents.course_id`. **That column does not exist on any environment** — `documents` keys on `offering_id`. Every call returned `400 Bad Request`, and because the tool is written to "degrade silently to `[]`", the model saw an empty result and concluded the course had no such material.

That is what produced this, on the tutor with #533's prompt fix **already applied**:

> I'm sorry, but I couldn't find any information about Markov chains in the course materials. Let's focus on the main topics of this course.

So the two triggers are independent: #533 fixed the catalog block; this fixes the tool.

Two things kept it hidden: the silent degradation swallowed the 400, and the evals inject a fixture retrieval seam (ADR 0023) that never issues the real query — the suite could not have caught it.

Fixed by resolving offerings through `services/academics.user_offering_ids_for_course` (the idiom `routes/flashcards.py:141` already used), plus the `deleted_at` filter the old query omitted. The regression test pins the query **by column name**, so a schema rename breaks a test instead of silently disabling the tool again.

## 2. An empty lookup is not information about the course

`search_course_materials_tool` now returns `CourseMaterialsResult` (`materials` + `guidance`) instead of a bare list, so an empty lookup arrives carrying an explicit instruction not to mention it. A rule at the point of the empty result lands where one thousands of characters earlier in the preamble does not.

The preamble gains the matching rule: course *information* (instructor, prerequisites, credits, coverage) is surfaced ONLY when the student asks about the course itself — never an opener, never a qualifier. A test pins the other side too, so this can't over-correct into a tutor that refuses to discuss its own course.

## 3. Quizzes are now practical

For quantitative concepts, at least `ceil(2N/3)` questions must pose concrete values and require computation; the rest stay conceptual. Distractors must be answers a student actually reaches by making a specific mistake — a sign slip, a transposed matrix, an unnormalised vector — never arbitrary padding.

Prompt-only: `QuizQuestion`'s comments record that Gemini's constrained decoding hit "too many states for serving" on the Lite tier, so the schema stays MCQ-only. A worked problem is still four candidate results.

**Placement turned out to be the whole game.** Three live 6-question runs on Eigenvalues + Markov Chains, bar of 4:

| prompt | worked problems |
|---|---|
| rule stated late | 2 / 6 |
| hoisted before the tool workflow | 3 / 6 |
| + restated as a FINAL CHECK | **5 / 6** |

Same failure mode as #533's preamble — a correct instruction buried mid-prompt loses to the ones around it.

## Verification

Deterministic tests prove the text changed, not that the model complies, so each change was also checked live:

- **Tutor**, socratic and expository, on a geometric-algorithms course: teaches Markov chains, zero course commentary.
- **Quiz**: 5/6 worked problems, arithmetic verified by hand (trace 5 / det 6 → 2,3; det = 1·(−2)·3 = −6; `[0.5,0.5]P = [0.55,0.45]`; `πP = π → [1/3,2/3]`), distractors are real error-results.

## Known limits

- Quiz compliance is prompt-enforced, not schema-enforced. If it regresses, the next step is a deterministic post-generation count in `routes/quiz.py` with one revision pass — latency for a hard guarantee.
- Nothing gates `search_course_materials` *usage* in CI, so a future drop to zero calls stays invisible.

Spec: `docs/superpowers/specs/2026-08-11-tutor-grounding-and-practical-quizzes-design.md`


Jose-Gael-Cruz-Lopez:
## Code review — tutor retrieval + practical quizzes (stacked on #533)

PR #534 does three things: repairs `search_course_materials` (the `documents.course_id` filter was querying a column that has not existed since migration `0025_study_integrity.sql` recreated the table on `offering_id`), replaces the tool's bare-list return with a `CourseMaterialsResult` carrying model-facing `guidance` plus a matching preamble rule, and rewrites quiz generation to over-generate-and-select rather than retry. The retrieval half is the strong half — I verified `documents` really does carry `offering_id` and `deleted_at` and no `course_id`, and the query stays user-scoped so the #125 IDOR fix is preserved. The quiz half is where the problems are: the last three commits are a revert cycle (`26f1449` → `c84a5ad` → `1f83afa`) and the tail of that churn is still in the tree. CI is red, an added "retry on a different model" retries on the same model, an agent-level `thinking_budget=0` now covers the Pro path, and the design doc committed with the PR describes the design that was reverted.

### Findings

#### P0

**[P0] CI is red: 3 new ruff F401 violations, and they block the job before pytest runs** — `backend/tests/test_quiz_agent_imports.py:155,179,188`

```
F401 [*] `agents.quiz.is_worked_problem` imported but unused
   --> tests/test_quiz_agent_imports.py:155:33
    |
154 |     def test_drops_unanswerable_questions_using_the_surplus(self):
155 |         from agents.quiz import is_worked_problem, select_quiz_questions
    |                                 ^^^^^^^^^^^^^^^^^
help: Remove unused import: `agents.quiz.is_worked_problem`

Found 3 errors.
##[error]Process completed with exit code 1.
```

(run `31620819267`, head `1f83afa`.) Three of the new `TestOverGenerateAndSelect` methods import `is_worked_problem` and never call it. This originates entirely in #534's delta — #533's branch reports `mergeStateStatus: CLEAN` — and it is more than a lint nit: `.github/workflows/ci.yml` runs the `Lint (ruff — baselined ratchet)` step *before* `python -m pytest tests/`, so **pytest never executed for this head**. Every one of the ~470 new backend test lines in this PR is currently unverified. The last green run (`31608516686`) predates the final two commits.

#### P1

**[P1] The "retry on a different model" retries on the same model** — `backend/routes/quiz.py:295-324`

```python
    # The retry therefore has to CHANGE something. Re-running the same
    # payload on the same model reproduces the failure exactly — measured:
    # a plain fresh re-run failed both times, at 137s and 144s. Escalating
    # to gemini-2.5-flash is the smallest change that leaves the failing
    # input behind, ...
    fallback = _resolve_model_pref("fast")
```

`_PREF_MODEL_NAMES` in the same file (lines 122-125) maps `"fast" → "gemini-2.5-flash-lite"`, and `agents/_providers.py:60` sets `_DEFAULTS["quiz"] = "gemini-2.5-flash-lite"`. The fallback is the identical model that just failed. By the comment's own measurement that turns the empty-`finish_reason=error` case from ~140 s to ~280 s of student wait before the same 502. The log line at 322 also reports `"retrying on gemini-2.5-flash"`, which will send whoever reads it looking for a model that was never used. `"smart"` is the only entry in that map that is actually a different model.

**[P1] Pinning `thinking_budget=0` on the agent breaks the `model_pref="smart"` (Pro) path** — `backend/agents/quiz.py:288-295`

```python
_QUIZ_SETTINGS = GoogleModelSettings(
    max_tokens=8192,
    google_thinking_config=ThinkingConfig(thinking_budget=0),
)

quiz_agent = Agent[SaplingDeps, Quiz](
    model=model_for("quiz"),
    model_settings=_QUIZ_SETTINGS,
```

Agent-level `model_settings` apply to every run, including one where `run(model=...)` overrides the model. `GenerateQuizBody.model_pref` (`models/__init__.py:65`) accepts `"smart"`, and `routes/quiz.py:124` resolves that to `gemini-2.5-pro`. This codebase already knows Pro rejects a zero budget — `agents/flashcard.py:49`: "Flash accepts thinking_budget=0 (unlike Pro)" — and `agents/chat_tutor.py:14-20` states the exact reason not to put this on the agent: *"the Pro thinking cap is applied at the route layer (`routes.learn._build_pro_model_settings`), not on the agent itself, because the same agent instance also serves Lite runs (via the 'fast' model_pref override)."* `routes/learn.py:86-102` is the pattern to mirror; it uses `_PRO_THINKING_BUDGET = 2048`, not 0. As written, `POST /api/quiz/generate {"model_pref":"smart"}` sends `thinkingBudget: 0` to Pro on both attempts.

**[P1] The claimed frontend fix for the 15-question 422 is not in this PR** — `backend/models/__init__.py:49-57`, `backend/tests/test_quiz_routes.py:552-554`

```python
    # ... 15-question
    # quizzes are not available at this schema complexity — QuizPanel's
    # COUNT_OPTIONS drops to 5 / 10 to match, so the picker never offers a
    # value the API refuses.
    num_questions: int = Field(default=5, ge=1, le=10)
```

No frontend file is among this PR's 14 changed files, and `frontend/src/components/QuizPanel.tsx` still declares `COUNT_OPTIONS` as `5 / 10 / 15`. Picking "15 questions" remains an unconditional 422 for every user. The same revert left the test class docstring asserting the opposite of the two tests directly beneath it:

```python
    The bound is 15 because that is the largest count QuizPanel offers.
    While it sat at 10, picking "15 questions" in the UI was an
    unconditional 422 — the picker offered a value the API refused.
    """

    def test_num_questions_over_cap_rejected(self):
        """POST with num_questions=11 should return 422, not silently truncate."""
```

#### P2

**[P2] Two comments reference functions that do not exist** — `backend/agents/quiz.py:36-37`, `backend/routes/quiz.py:233`

```python
# Output-validation retry budget, read back by _on_final_attempt so the
# gates below know when they are out of moves.
```
```python
        # Read by quiz_agent's _enforce_requested_count output validator.
```

`_on_final_attempt`, `_enforce_requested_count` and `_enforce_worked_ratio` return zero matches across `backend/`. The surviving validator is `_select_requested_quiz`, whose contract is the opposite of what the first comment describes — it "MUST NOT RAISE" (`agents/quiz.py:514`) and has no notion of a final attempt.

**[P2] The routing message instructs the model about a schema field that was removed** — `backend/routes/quiz.py:251-254`

```python
    routing_msg = (
        f"Generate {ask_for} {difficulty} questions for the student. "
        f"At most {allowance} of them may be kind='conceptual' — the rest "
        f"must be worked problems with concrete values. "
```

`QuizQuestion` (`agents/quiz.py:63-90`) has no `kind` field — it was reverted in `1f83afa`, and `is_worked_problem` now infers the classification from the stem. Constraining a key the model cannot emit is wasted prompt at best; RULE 2 in the system prompt already states the same rule in terms the schema supports.

**[P2] The design doc added by this PR documents the design that was reverted** — `docs/superpowers/specs/2026-08-11-tutor-grounding-and-practical-quizzes-design.md:143-176`

```
The count, the answerability, and the ratio are now **output validators**
on `quiz_agent`, reading `num_questions` off `SaplingDeps`. Each raises
`ModelRetry` naming exactly what to fix.
...
The array bound is **gone**, not raised.
...
So `QuizQuestion` gained `kind: "worked_problem" | "conceptual"`,
self-declared, and `_enforce_worked_ratio` counts it.
```

All three statements are false at head: the single validator must not raise, `Quiz.questions` is `Field(min_length=1, max_length=10)` (`agents/quiz.py:117`), and there is no `kind` field. `_enforce_worked_ratio` (line 163) and `_on_final_attempt` (line 176) do not exist. The file is headed `Status: implemented (PR #534, stacked on #533)`, so the next reader will trust it — and §3 still states the `ceil(2N/3)` rule that `conceptual_allowance` replaced.

**[P2] `select_quiz_questions` mixes value-equality with identity, so duplicate questions shorten the quiz** — `backend/agents/quiz.py:485-506`

```python
    if len(chosen) < wanted:
        shortfall = wanted - len(chosen)
        extra = [q for q in conceptual if q not in chosen][:shortfall]
...
    order = {id(q): i for i, q in enumerate(questions)}
```

`q not in chosen` uses Pydantic's field-value `__eq__`; the reordering three lines later uses `id()`. If the model emits two conceptual questions with identical fields — precisely the failure the system prompt anticipates ("you run out of distinct angles around question 6 and stop early") — the second is treated as already chosen, the backfill drops it, and the student gets a short quiz while a usable question sat unused in `conceptual`. Every other membership decision in this function is identity-based.

#### P3

**[P3] Stale bound in the model-settings rationale** — `backend/agents/quiz.py:280-283` — "comfortably fits the largest quiz the UI can ask for (15 requested → 17 generated…)". `quiz_ask_size` returns `min(wanted + 2, 10)`, so the largest generation this route can request is 10.

**[P3] A test passes a field that no longer exists** — `backend/tests/test_output_retry_hardening.py:196` — `kind="worked_problem"` on a `QuizQuestion`. Pydantic's default `extra="ignore"` swallows it silently, so the line asserts nothing and reads as if `kind` were still real.

**[P3] The retry also fires on `UsageLimitExceeded`** — `backend/routes/quiz.py:317` — the second attempt reuses the same `ORCHESTRATOR_LIMITS` object, so a run that exceeded the budget is guaranteed to exceed it again. Only the `UnexpectedModelBehavior` case is justified by the comment above the loop.

**[P3] The repaired tool now costs 3 PostgREST round-trips per call** — `backend/agents/tools/chat_context.py:155-172` — `user_offering_ids_for_course` (`services/academics.py:180-195`) issues an unfiltered `course_offerings` read plus an unfiltered `enrollments` read, and unlike its neighbours `offering_course_id` / `_term_for_offering_cached` it carries no `@lru_cache`. This is the chat tutor's per-turn path.

### Stacked-PR risk

- Base is `fix/tutor-course-scope-pr` (#533), itself `OPEN`, based on `main`, `CLEAN`. #534 is `MERGEABLE` / `UNSTABLE`, and the instability is its own red CI, not a conflict. Merge order is stated in the first line of the body — good.
- Intra-stack overlap is real but benign. `backend/agents/chat_tutor.py` and `backend/tests/test_chat_tutor_imports.py` are the only two files both PRs touch, and #534's change is purely additive: a new `COURSE INFORMATION` paragraph at `chat_tutor.py:147-154`, immediately after the `SCOPE` paragraph #533 added at 139-146. Nothing #533 wrote is reverted or rewritten.
- The fallback is undocumented. None of #534's three claims actually depend on #533's code — the retrieval fix, the tool-result guidance and the quiz work are all independent — so if #533 is reworked or dropped, #534 can be retargeted at `main`. But its new paragraph reads as a continuation of #533's SCOPE rule, and `test_chat_tutor_imports.py` would need rebasing. One line in the body ("retargetable at main; only the preamble paragraph needs reconciliation") would close that.
- Side effect worth knowing: because the base is not the default branch, CodeRabbit skipped this PR entirely — *"Auto reviews are disabled on base/target branches other than the default branch."* The stack costs #534 its automated pass, which is part of why a three-line ruff regression got this far.

### What's good

- The retrieval diagnosis is correct and the fix is the right one. `0025_study_integrity.sql` recreates `documents` on `offering_id` with no `course_id`, so the old filter really was a guaranteed 400; routing through `services/academics.user_offering_ids_for_course` matches the existing `routes/flashcards.py` idiom, the #125 user scoping is preserved, and `deleted_at is.null` closes a genuine soft-delete leak into tutor context.
- `TestSearchCourseMaterialsQueryShape` pinning the query *by column name* is exactly the right regression test for a bug whose whole nature was that nothing observed it — as is the candour about why the evals could not have caught it (the ADR 0023 fixture seam).
- The `CourseMaterialsResult` return-type change is a clean cross-boundary edit: `chat_tutor.py`, `note_chat.py`, `retrieval.py` and `chat_stream.py` all check out, no production path indexes the result as a list, and only the two tests that needed updating were updated.
- Replacing `ModelRetry`-based ratio enforcement with over-generate-and-select is the correct call, and `resolve_correct_index`'s two-bar rule (≥0.90 similarity **and** ≥0.10 margin over the runner-up) is a genuinely careful way to absorb a retyping stutter without ever guessing between two numeric options.

**Verdict: request changes.** Green CI comes first, and because ruff gates the job the ~470 new backend test lines have never run. The same-model "escalation" and the Pro `thinking_budget=0` are both live-path bugs, and the reverted-design residue (dead symbol names, `kind='conceptual'` in the routing message, the spec doc, the QuizPanel claim) should be swept in the same pass.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Blocker
- **3 ruff F401 blocked CI before pytest**, so ~470 new backend test lines had never executed. Removed. Running the suite then exposed a second failure hidden behind that gate: `test_budget_is_read_from_the_run_context` read `_max_output_retries`, which only exists on pydantic-ai 1.107+ while the pin is 1.89 — rewritten to probe both names.

### Major
- **The "retry on a different model" retried on the same model.** `_resolve_model_pref("fast")` resolves to `gemini-2.5-flash-lite`, which *is* `_DEFAULTS["quiz"]` — so the escalation doubled the wait to ~280 s before the identical 502, and the log named a model never used. Now escalates to a genuinely different model, carries the `SAPLING_MODEL_MODE` seam, skips when the caller named a model, and logs the model actually used.
- **Agent-level `thinking_budget=0` broke the Pro path.** `model_pref="smart"` resolves to `gemini-2.5-pro`, which rejects a zero budget — and this codebase already documents that (`flashcard.py`) and already solves it at the route layer (`learn.py::_build_pro_model_settings`). The thinking config moved off the agent to a per-run decision; `max_tokens` stays on the agent.
- **The claimed frontend fix for the 15-question 422 was not in the PR.** `QuizPanel`'s `COUNT_OPTIONS` still offered 15 against an `le=10` bound, so picking it was an unconditional 422. Now 5 / 10, and the test docstring that asserted the opposite of its own tests is fixed.

### Minor / nits
Comments referencing `_on_final_attempt` / `_enforce_requested_count` / `_enforce_worked_ratio` (none of which exist) rewritten · routing message no longer constrains the removed `kind` field · the design doc now describes what shipped rather than the reverted design · `select_quiz_questions` uses identity consistently, so two field-identical questions no longer silently shorten the quiz · stale generation-size comment · removed `kind=` from a test where Pydantic silently ignored it · retry narrowed off `UsageLimitExceeded` · course→offerings lookup cached.

**Verification** — `ruff check .` clean · **1556 passed, 32 skipped** · `tsc` clean

AndresL230:
Review notes (read-only review; nothing checked out, no tests run). **Verdict: not mergeable as it stands**, mostly for structural reasons. There is good work here worth landing in pieces.

**Structure**
- The base is `fix/tutor-course-scope-pr` (#533), not main. "Mergeable" means it merges into #533's branch; CodeRabbit skipped it for the same reason. #533 has its own open review notes, so this is blocked behind it.
- Against current main this branch's own commits conflict in `backend/routes/quiz.py` (6 hunks), `backend/agents/deps.py` and `backend/services/academics.py`. Main's quiz backend batch (+1.6k lines in `routes/quiz.py`) landed after this was cut.

**Superseded: drop these**
- `81dac7d9` (the `search_course_materials` column fix). #562 is merged and fixes the same bug with a wider, correct scope: `academics.course_offering_ids`, because uploads are stamped by `resolve_offering`, not by enrollment. `user_offering_ids_for_course` misses every upload made after a term rollover.
- `7ba26ef7` (the `lru_cache` under `user_offering_ids_for_course`). It is invalidated in-process at one site only, skips the 409 lost-race branch, goes stale across workers and ops scripts, and its callers include the no-retake check in `graph_service`. Main already has an uncached `course_offering_ids` doing the same read.

**Findings in what remains**
1. **`resolve_correct_index` fuzzy pass can mark a distractor correct** (`agents/quiz.py:388`, floor 0.90, margin 0.10). When the model omits the true option, which is the case the docstring cites: `"x = -6"` vs `"x = 6"` scores 0.909; `"λ = 2 and λ = 3"` vs `"λ = -2 and λ = 3"` 0.968; `"[[7, 10], [15, 22]]"` vs `"[[7, 10], [15, 20]]"` 0.947. RULE 2 of the prompt asks for exactly these near-miss distractors. Main drops such a question; this serves a wrong answer key and feeds mastery from it. The normalised pass (case, whitespace, trailing punctuation) is safe; remove the fuzzy pass or limit it to non-numeric prose.
2. **The output validator hides drift from the route.** `_select_requested_quiz` drops unanswerable questions and trims to N inside the agent, so the route's `dropped` counter never sees them and the #543 top-up stops firing for partial drift (asked 5 → 7 generated → 4 unanswerable → 3 served, `dropped=0`, no top-up). It also does no stem dedupe, so the route can drop a duplicate after the unique surplus was already discarded. On a top-up run `wanted` is still N, so every top-up logs a spurious "served k of N" WARNING. The new route tests patch `quiz_agent.run`, which bypasses the validator, so nothing covers this interplay.
3. **`_NO_MATERIALS_GUIDANCE` is sent even when the read failed** ("No indexed course documents matched… Do not mention this lookup."). That tells the model to conceal a failure, against the preamble's "don't fabricate context", and it misfires when the student asks what their own slides say. The same tool is registered on `note_chat`, whose prompt says "not a tutoring session" while the guidance says "teach the topic". Keep the ban on volunteering scope commentary; add an exception for questions about the student's documents, and do not assert "no documents matched" on a failed read. #562 added an F5 empty-result signal at that wrapper, so the failed-read case is now distinguishable.
4. **Semantic conflicts with main's quiz route beyond the textual ones:** main's `_run` returns `(Quiz, model_name)` and `_absorb(quiz, model)` stamps E5 provenance; the fallback-model retry here returns a bare `Quiz`, and a naive resolution would mis-stamp exactly the case that provenance exists for. `SaplingDeps(...)` needs main's `feature="quiz"` and `share_class_context=` plus this PR's `num_questions=`. `_resolve_bu_code` collides with main's `CourseRow` / `BuCodeLookup` rewrite.
5. Prompt/code drift: "AT MOST TWO when N is 11 to 15" is dead (the cap is 10); "a short quiz is rejected and regenerated" vs a validator documented "MUST NOT RAISE"; the PR body still describes `ceil(2N/3)`. `is_worked_problem` (3+ digits in the stem) counts a history question with a year as "worked" and warns on every humanities quiz. Over-generation asks for 2 extra questions at N=3 and N=5, about +40% output tokens at N=5, which cuts against the cost work in #643 and makes `docs/quiz-prompt-budget.md` stale.

**Suggested split**
- (a) Tutor guidance + preamble rule (`23d22554`), rebased on main after #533, with the guidance amended per finding 3.
- (b) Quiz reliability: thinking off on Flash / Flash-Lite, `max_tokens`, fallback to a genuinely different model only on `UnexpectedModelBehavior` and not when the student chose a model. This stands on its own and is valuable.
- (c) Quiz practicality: prompt, over-generation and selection, reworked against main's E5/E6 route per findings 1, 2 and 4.

Worth keeping: the diagnosis and the `deleted_at` filter, `CourseMaterialsResult` (delivering guidance at the point of the empty result is a sound technique once the guidance is truthful), the COURSE INFORMATION preamble rule and its test, the per-tier thinking config, and the measured A/B notes on Gemini schema complexity in the `Quiz` field comments.

AndresL230:
Closing — 47 commits behind `main`, and its headline fix has already shipped.

**Superseded:** "`search_course_materials` had never worked" is exactly what #562 fixed and merged (`ea8bb497`). #646 rewrote the same textless-turn paths in `routes/learn.py`. #629/#630/#484 changed visibility, shareability and `chunk_text` encryption under `rag_service.py` and `agents/tools/chat_context.py`.

**Split rather than dropped** — this PR was carrying two unrelated halves, and only one was superseded:

- **Quiz half -> #656.** `agents/quiz.py` (+395) and `routes/quiz.py` (+234): the `ceil(2N/3)` computational-question rule, mistake-derived distractors, and the measured placement result (2/6 -> 3/6 -> **5/6** worked problems by hoisting the rule before the tool workflow and restating it as a FINAL CHECK). Nothing else covers this and #631 does not reach it.
- **Tutor half -> #631.** "An empty lookup is not information about the course" — suppressing course-scope commentary on an empty retrieval. That now also has a home in `services/tool_signals.py::report_empty_result`, which did not exist when this was written.

Not rebased for the same reason as #533: stale eval cassettes mean billed re-records on work that is partly already merged.

The branch is kept so both follow-ups can read the diff.

PR #535, opened 2026-08-12
feat(landing): port the v5 marketing landing onto staging

Darkest-Teddy:
Puts the v5 marketing landing on `staging` **without** merging `feat/landing-v5-port`.

## Why not just merge the v5 branch

`feat/landing-v5-port` is **217 commits** ahead of `origin/staging` and the diff is mostly not the landing page — 179 backend test files, 31 agents, 27 services, 17 routes, and **15 new migrations**, including two colliding `0032_*` and two colliding `0033_*` basenames. It also isn't a fast-forward: `staging` has one commit that branch lacks (`82787ca`, gradescope). None of that is needed to show the new landing, so this ports the public surface only — no backend changes, no migrations.

## What's in here

- `components/landing-v5/**` and `lib/landing/**` — the landing engine, `hero3d`, drag simulation, and the Act sections
- The `(public)` route group: the new `/`, plus **`/team`, `/wiki`, `/gallery`, `/news`, `/faq`** — the v5 navbar links to these and staging didn't have them
- `components/marketing/**` and `components/companion/CompanionShell`
- `globals.css` and `app/layout.tsx` (OG/metadata), `lib/deployGuard.ts` **with its matching test**, `lib/errorMessage.ts`
- `api.ts` gains only `subscribeToNewsletter` — staging's backend already mounts `/api/newsletter`

## One deliberate divergence from upstream

`marketing/SignInModal` calls the **3-arg** `setActiveUser`. Upstream passes a 4th `{ persist }` option that exists only on that branch's `UserContext`, whose wider change (+127/−21) also adds app-wide session recovery and shell-route redirects. Changing auth behaviour for every screen is out of scope for a landing port, so the option is dropped here and should come back in the same commit that brings that `UserContext` over. The call site carries the same note.

`eslint-suppressions.json` gains an entry for the new `components/marketing/SignInModal.tsx` path; staging only suppressed that rule at the old `components/SignInModal.tsx` path.

## Verification

- `tsc --noEmit` — clean
- `next build` with staging's build vars (`DEPLOY_ENV`/`BACKEND_URL`/`NEXT_PUBLIC_API_URL`/`COOKIE_DOMAIN`) — succeeds, emitting `/team`, `/wiki`, `/gallery`, `/news`, `/faq`
- `eslint` — clean **except one pre-existing error**: `react-hooks/exhaustive-deps` in `settings/GradescopeConnect.tsx:65`, which arrives untouched from `origin/staging` (it is neither modified nor suppressed on that branch). CI will likely go red on it until it's fixed separately — deliberately not touched here.
- **vitest was not run locally** — it needs Node 22 and this machine is on 20. CI runs it on Node 22.

## Deploy note

Per `wrangler.toml` and ADR 0020, the staging Workers Build must set **`DEPLOY_ENV=staging` as a build variable**, not just a runtime var — the `/api` rewrite and `NEXT_PUBLIC_*` are baked at build time. That omission is what previously put prod config on the staging worker.

## Summary by CodeRabbit

- **New Features**
  - Added public About, Team, Wiki, Gallery, News, FAQ, and Careers pages.
  - Introduced interactive galleries, article search and filtering, FAQ accordions, career listings, and rich knowledge-base content.
  - Redesigned the landing experience with interactive learning demonstrations, knowledge graphs, tutor modes, quizzes, flashcards, study tools, and beta signup.
  - Added Google sign-in and newsletter subscription flows.

- **Improvements**
  - Standardized navigation, page layouts, accessibility, responsive styling, animations, and site metadata.
  - Improved error messaging and deployment-aware canonical links.


Jose-Gael-Cruz-Lopez:
## Code review — v5 marketing landing onto staging

This ports the v5 marketing landing onto `staging` as a public-surface-only change, and the central claim in the description holds up: I verified there are no `backend/**` files, no migrations, no `middleware.ts`, no `next.config.ts` and no `wrangler.toml` in the diff (74/74 changed files are under `frontend/`), and `git diff --diff-filter=D` is empty — nothing is deleted out from under another surface. The Layer-2 token scoping is done properly (`.public-surface, .landing-page` in `globals.css:828`, with deliberate literal fallbacks on `.card--hero`), and every new `landing-v5`/`marketing`/`companion` component carries `"use client"`. The problems are elsewhere: a dropped auth-error path, staging metadata that points at production, a colour system that bypasses the design tokens entirely, and — the big one — this branch is a *stale snapshot* of a landing that is already merged on `main`.

### Findings

#### P0

**[P0] Both CI jobs are red, and the failures hide the fact that nothing was typechecked or tested** — `.github/workflows/ci.yml:41,90-94`

Backend (`ruff check .`, run 31553863104):
```
F401 [*] `pydantic.Field` imported but unused
  --> models/gradescope.py:12:33
Found 1 error.
##[error]Process completed with exit code 1.
```
Frontend (`npx eslint .`, same run):
```
frontend/src/components/settings/GradescopeConnect.tsx
  65:6  error  React Hook useEffect has a missing dependency: 'refreshAll'. Either include it or
               remove the dependency array  react-hooks/exhaustive-deps
✖ 33 problems (1 error, 32 warnings)
##[error]Process completed with exit code 2.
```
To be clear about attribution: **neither is yours.** Both reproduce verbatim at the base commit `82787ca` — `backend/models/gradescope.py` imports `Field` with zero `Field(` uses there, and `GradescopeConnect.tsx:65` is already `}, []);`. `gh run list --branch staging` shows staging's last green run was `8bce5b02` on 2026-07-10; `82787ca` never ran CI, so it landed red. The reason this is still P0 for *this* PR: the workflow runs `ruff` before `pytest`, and `eslint` before `npx tsc --noEmit` and `npm test`. Both jobs die at the lint step, so **the type checker and both test suites never executed against these 14,954 added lines.** You currently have no signal that the port compiles. Fixing base (or rebasing once it is fixed) is a prerequisite for this PR meaning anything.

#### P1

**[P1] `?error=session_expired` no longer reaches the sign-in modal** — `frontend/src/app/(public)/page.tsx:89`

```tsx
<SignInModal open={signInOpen} onClose={() => setSignInOpen(false)} />
```
`middleware.ts` (untouched here) bounces expired sessions to the landing with a query param:
```ts
// middleware.ts:38
if (errorCode) u.searchParams.set('error', errorCode)
// middleware.ts:52 and :67
if (!session) return redirectToSignin(request, 'session_expired')
```
The page this replaces read that param and threaded it through — base `(public)/page.tsx:90` `const err = params.get('error')`, then `:879` `errorCode={signInError}`. The new page never reads `?error` (no `searchParams`/`URLSearchParams` anywhere in `page.tsx` or `useLanding.ts`), so the prop is never supplied. Meanwhile `components/marketing/SignInModal.tsx` still ships the whole error surface — `ERROR_COPY` at `:11` including `session_expired` at `:16`, and `:122` `const visibleError = localError ?? errorCode ?? null` — all now dead. Net effect: a user kicked out mid-session lands on the marketing page with no explanation of why they were signed out. Restore the `?error=` read and pass it through.

**[P1] The staging deploy will advertise production canonical and OG URLs** — `frontend/src/app/layout.tsx:43`

```tsx
metadataBase: new URL(resolveSiteUrl(process.env)),
…
alternates: { canonical: '/' },
```
`resolveSiteUrl` falls through to production when neither knob is set:
```ts
// deployGuard.ts:98-106
export function resolveSiteUrl(env: EnvSource): string {
  const deployEnv = (env.DEPLOY_ENV ?? '').trim().toLowerCase();
  if (deployEnv && deployEnv in FRONTEND_ENVS) return FRONTEND_ENVS[deployEnv as FrontendEnv].siteUrl;
  const explicit = (env.NEXT_PUBLIC_SITE_URL ?? '').trim();
  if (explicit) return explicit.replace(/\/+$/, '');
  return FRONTEND_ENVS.production.siteUrl;
}
```
`frontend/wrangler.toml` is not modified by this PR, and its `[env.staging.vars]` block sets only `NEXT_PUBLIC_API_URL`, `BACKEND_URL` and `COOKIE_DOMAIN` — **neither `DEPLOY_ENV` nor `NEXT_PUBLIC_SITE_URL`.** So on `frontend-staging` every page resolves `metadataBase` to `https://saplinglearn.com`, making `canonical`, `og:url` and `og:image` all point at production. Since the entire point of this PR is "show the new landing on staging", any staging link shared for review will unfurl with production's card, and staging self-canonicalises to prod. Set `DEPLOY_ENV = "staging"` in `[env.staging.vars]` (which also arms the existing `checkFrontendDeployEnv` lock in `next.config.ts:46`).

**[P1] The landing runs a parallel colour palette that shares no value with the design tokens** — `frontend/src/components/landing-v5/*`

Counting hex literals across `landing-v5` (excluding the `galleryMinis` illustration art) gives ~900 hardcoded colours. The most frequent:
```
79  #0c5638      48  #fdfcf9      46  #0e9e5a      40  #12201a
37  #33443b      34  #9aa5a0      31  #8b9891      30  #8fd9a8
```
Not one of these is a token. The core set in `globals.css` is `--brand-forest: #1B6C42`, `--brand-forest-bright: #2D8F5C`, `--sap-500: #4e873c`, `--ink-0: #faf8f3`, `--ink-800: #1a1814`. Note the near-misses in particular: `#fdfcf9` vs `--ink-0 #faf8f3`, and `#12201a` vs `--ink-800 #1a1814`. They are close enough to look like the brand and different enough to be visibly off it at the seam where a user crosses from the landing into the app shell. The Engineering Style Guide §3/§4 is explicit — "inline `style` objects referencing global CSS-variable tokens … Use the tokens; don't hardcode values." This is doubly odd because the PR itself builds the correct scaffolding: `(public)/layout.tsx` mounts `.public-surface` and `globals.css:839-841` defines `--surface-hero`/`--brand-glow` on it, and then no component uses any of it.

**[P1] This is a stale snapshot of a landing already merged on `main`, and promoting it will fight `main`** — whole `landing-v5` tree

`origin/main` is `4f5a7f9 Merge pull request #524 from SaplingLearn/feat/landing-v5-port` — the v5 landing is already there. `git merge-base origin/staging origin/main` is `57469ad`, which contains **no** `frontend/src/components/landing-v5/`, so the directory is added independently on both sides: promoting `staging` → `main` is an **add/add conflict across ~50 files**, not a fast merge. `main` has taken 19 commits on the landing since the merge-base, and `git diff origin/main 395588da` over the landing tree is only 4 files / +9 −38 — i.e. this branch is *behind*, missing `a3d0371 fix(landing): raster, veil and scroll-coupling polish on the v5 hero` (the `MAG`/`zoom` raster correction in `ActTutor.tsx`, and the re-tuned hero washes in `Hero.tsx`).

The dangerous part is that one of those divergences is invisible to the type checker:
```ts
// this branch — hero3d/index.ts:52-56 (an intensity multiplier, 0..2)
export function startHeroRig(canvas, getMouse, getParallax: () => number)
// origin/main — hero3d/index.ts:76-80 (a scroll offset in pixels)
export function startHeroRig(canvas, getMouse, getHeroShiftPx: () => number)
```
Both are `() => number`, and both `hero3d/index.ts` and `useLanding.ts` are in the conflict set. Resolving that conflict by taking one file from each side compiles cleanly and silently breaks the hero. Please rebase this branch onto `main`'s landing rather than re-adding an older copy, or state explicitly how the promotion merge will be resolved.

#### P2

**[P2] Three of the four new `deployGuard` exports have no production caller** — `frontend/src/lib/deployGuard.ts:52-146`

`resolveFrontendEnv`, `expectedEnvForHost` and `detectHostConfigMismatch` are referenced only by `deployGuard.test.ts`. `resolveSiteUrl` is the only one wired into anything (`layout.tsx:4`). `detectHostConfigMismatch`'s own docstring bills it as "Runtime defence-in-depth for what the build-time guard cannot see", citing "the exact failure behind staging's `session_expired`" — but `middleware.ts` is not in this diff, so nothing calls it and it defends nothing. That is ~100 lines of inert code plus 131 lines of tests landing in a shared lib on a marketing PR. Either wire it into the middleware or drop it from this PR.

**[P2] `components/SignInModal.tsx` is now orphaned, and its replacement is a near-copy** — `frontend/src/components/SignInModal.tsx`

After this PR the only `SignInModal` import in the tree is `(public)/page.tsx:14` → `@/components/marketing/SignInModal`. The original `components/SignInModal.tsx` (347 lines) has **zero importers** but is still in the tree; the new `components/marketing/SignInModal.tsx` is 352 lines and `diff` between the two is 67 lines. Two copies of the OAuth modal, one of them dead, will drift — and the dead one is the one that still gets fed `errorCode` in nobody's code path. The Engineering Style Guide §3 calls this out by name as the reason `components/ui/` exists ("before they landed the app had 5 distinct button shapes and 18+ pill/badge implementations"). Delete the orphan or make the marketing one a thin wrapper.

**[P2] A new eslint suppression was added for code introduced in this PR** — `frontend/eslint-suppressions.json`

```json
"src/components/marketing/SignInModal.tsx": {
  "react-hooks/set-state-in-effect": { "count": 1 }
},
```
Style Guide §7 states the ratchet's purpose as "don't fix the whole backlog to land a change, but never add to it." The baseline exists to grandfather *legacy* violations; `marketing/SignInModal.tsx` is created by this PR, so this is a new violation being suppressed rather than fixed. `set-state-in-effect` is also usually a real render-loop smell worth a look rather than a waiver.

**[P2] `.landing-page` no longer exists, leaving dead CSS and a comment that now says the opposite of the truth** — `frontend/src/app/(public)/layout.tsx:7`

```tsx
// The landing page keeps its own .landing-page class for the mesh-background visual.
export default function PublicLayout({ children }: { children: React.ReactNode }) {
  return <div className="public-surface">{children}</div>;
}
```
No element in the tree carries `className="landing-page"` at this HEAD — the only remaining mentions are three comments. But `globals.css:866-874` still ships the whole mesh block that depends on it:
```css
.landing-page .sapling-mesh-blob { position: absolute; border-radius: 50%; filter: blur(100px); … }
.landing-page .sapling-mesh-blob--1 { top: -10%; left: -10%; … }
```
along with the other `.landing-page`-prefixed rules the diff kept. All unreachable. Either the wrapper should still apply the class (if the mesh backdrop is meant to render) or the rules and the comment should go.

#### P3

**[P3] 1.3 MB of unoptimized PNGs committed** — `frontend/public/`

`journal-founding.png` is 821,816 bytes and `journal-ai-homework.png` is 500,189 bytes. They are rendered through `next/image` with a `sizes` hint (`Journal.tsx:110-114`), so delivery is fine, but the repo and the OpenNext asset bundle both carry the full-size originals. Worth running them through a compressor or shipping WebP/AVIF sources.

### What's good

- The scope discipline is real and verifiable: no backend files, no migrations, no `middleware.ts`, no `wrangler.toml`, and nothing deleted — exactly what the description promises, and a much better call than merging the 217-commit branch with its colliding `0032_*`/`0033_*` migrations.
- The Layer-2 token boundary is set up correctly — `.public-surface` scoping in `(public)/layout.tsx` mirroring `(shell)`, with the literal fallbacks on `.card--hero` and a comment explaining exactly why they are there. That is the right structure; the components just need to start using it.
- `"use client"` is correct on every new hook-using component, and the `deployGuard` additions are genuinely well-tested (131 lines of focused unit tests) even though most of the code they cover is not yet wired up.

**Verdict:** request changes — the auth-error regression and the staging-metadata bug are both user-visible, and the branch needs to be reconciled with the v5 landing already on `main` before it can be promoted.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Blocker
- **Both CI jobs died at the lint step, before tsc / pytest / vitest ever ran** — so none of this port had been typechecked or tested. Both base failures fixed properly: the unused `pydantic.Field` import, and `GradescopeConnect.tsx`'s `exhaustive-deps` (via `useCallback` keyed on `userId`, which also fixes a real bug the empty dep array hid — switching user kept the previous account's state).

### Major
- **`?error=session_expired` never reached the modal.** `middleware.ts` sets it and `marketing/SignInModal` still shipped the whole `ERROR_COPY` surface, all dead. Now read and threaded through, opening the modal on an inbound code and stripping the param on dismiss.
- **Staging advertised production canonical/OG URLs** — `resolveSiteUrl` fell through to prod because `[env.staging.vars]` set neither knob. `DEPLOY_ENV = "staging"` added, which also arms the existing build-time guard.
- `deployGuard.classify` now normalises trailing slashes, so a valid `https://api.saplinglearn.com/` no longer bypasses the runtime guard; `detectHostConfigMismatch` is wired into `middleware.ts` so it actually defends something.
- The orphaned `components/SignInModal.tsx` (zero importers, 67-line drift from its replacement) is deleted.

### CodeRabbit majors
`onLearn` opened the notetaker instead of the tutor (now via the new `galIndexOf` lookup, not a magic index) · OAuth URLs kept same-origin · `DARK_SECTIONS` reduced to the one genuinely dark section · Escape now exits explore mode (was a keyboard trap with `body { overflow: hidden }`) · `sim.destroy()` re-homes clusters so a StrictMode remount no longer kills the drag field permanently · hero-canvas unmount race · `CardsDemo` timer leak, global key capture and keyboard-inoperable flip · Gallery cards keyboard-openable · `BetaModal` single-column below the breakpoint + `type="email"` · hidden nav/pills removed from the tab order.

### Minor / nits
Dead `.landing-page` CSS removed · stylelint `font-family-name-quotes` 5→0 and `keyframes-name-pattern` 42→0 · lab-demo correctness (clamped percentages, partner snapshot, concept count, invalid guide picks) · ARIA roles and labels across FAQ/careers/calendar/quiz/rooms · `localStorage` input validated · journal PNGs recompressed **1.32 MB → 442 KB**.

### Still open (deliberately not done)
This branch is a **stale snapshot of a landing already merged on `main`** (#524) — promoting it is an add/add conflict across ~50 files, and `startHeroRig`'s third parameter changed meaning while keeping an identical `() => number` type, so a file-by-file conflict resolution compiles cleanly and silently breaks the hero. That is a merge-strategy decision, not a code fix.

**Verification** — `ruff check .` clean · `tsc` clean · eslint **0 errors** (base had 1) · **184 frontend tests pass**

Jose-Gael-Cruz-Lopez:

### This branch now carries a backend change, via #565

`staging` picked up the Gradescope repair in #565, so this PR is no longer frontend-only. That fix restored `routes/gradescope.py` and `services/gradescope_service.py` from `main` because the version on staging was unreachable (router prefix declared twice → every endpoint served at `/api/gradescope/api/gradescope/...`), called a DB client that does not exist (`.eq().execute()` on a PostgREST helper), and had no rate limiting or `bu-sso`.

I checked statically that every table the restored route touches exists on staging — `gradescope_credentials`, `gradescope_course_links`, `assignments`, `gradebook_categories` — so there is no missing-table hazard. **But that is a check of the repo's migration files, not of the staging database's actual ledger.** Please confirm on the real database:

```sql
SELECT table_name FROM information_schema.tables
WHERE table_schema = 'public'
  AND table_name IN ('gradescope_credentials','gradescope_course_links','gradescope_links');
```

All three should be present. Then exercise one endpoint end to end and confirm it resolves at a **single** `/api/gradescope/...` prefix rather than 404ing.

### Two things #565 deliberately left behind

```sql
SELECT count(*) FROM gradescope_links;
```

`gradescope_links` (created by staging's `0033_add_gradescope_links.sql`) now has **no reader** — the restored route uses `gradescope_course_links`. `services/gradescope_sync_service.py` has zero importers, and `models/gradescope.py` is imported only by that dead module. Nothing breaks, but if the table holds rows they are stranded, and the two modules are dead code worth removing in a follow-up.

### Ledger divergence between staging and main

Staging carries two migrations `main` does not, at numeric prefixes `main` reuses for different files:

| prefix | staging | main |
|---|---|---|
| 0032 | `0032_add_gradescope_credentials.sql` | `0032_rooms_missing_columns.sql`, `0032_retire_summer_2026.sql` |
| 0033 | `0033_add_gradescope_links.sql` | `0033_offering_section_not_null.sql`, `0033_realtime_publish_room_messages.sql` |

`schema_migrations` keys on the full basename, so these coexist without a primary-key collision — but `main` has **23 migrations staging lacks**, and staging is **239 commits behind**. Whenever these branches are reconciled, both migration sets apply and the relative ordering within `0032`/`0033` is alphabetical by accident rather than by design. Worth settling deliberately rather than discovering it during a promotion.

PR #536, opened 2026-08-12
fix(learn): add a concept back after deleting it, and stop a stale backend answering

Darkest-Teddy:
Split out of #534, which was carrying these three alongside unrelated quiz
work. Based on `main`; no dependency on #533/#534.

Reported as: **"when deleting a concept node, i should be able to add it back
of the same name by add concept."**

### The reported bug was a stale server, not the graph

`POST /api/graph/{user_id}/nodes` returned `404 {"detail":"Not Found"}` in
the browser while returning 200 through TestClient in the worktree. A backend
left running from a *different* working tree — older code, no `create_node` —
was still answering :5000.

Windows lets a second process bind a port another process is already
listening on, and the **first** binder keeps the traffic. Restarts bound a
second socket, logged "Application startup complete", passed `/api/health`,
and served nobody. `Get-NetTCPConnection` listed two owning PIDs; the stale
parent looked dead in `Win32_Process` while its uvicorn **reload child** held
the socket, so killing the parent alone did nothing.

`python main.py` now probes the port first and exits with the kill command.
Dev entrypoint only — containers and Railway import `main:app` and never
reach it.

The tell, for next time: if a route you can read in `routes/` is missing from
`curl -s localhost:5000/openapi.json`, you are talking to a different build.

### Two real Learn bugs found on the way

**Add-concept silently no-oped after a delete.** The composer resolved its
course as `topicNode?.course_id || selectedCourseId || null`. The picker is
"Course (optional)" and starts at `""`, so for anyone who hasn't chosen one
the course id came *entirely* from the focused node — and deleting a concept
clears the focus. `if (!label || !cardCourseId) return` then bailed before
doing anything: no request, no toast, no rollback. `resolveAddConceptCourseId`
falls back to the last course that did resolve, and says so when there has
genuinely never been one. Exported, like `resolveCardCourseId`, so the test
exercises the resolution `addConcept` actually calls rather than a mirror.

**Failures were unreadable.** `.catch(() => toast.error("Couldn't save the
concept — it was removed"))` discarded the `ApiError`, giving a 401, a 404, a
500 and an unreachable backend one identical sentence. Surfacing the status
and body is what produced the 404 + `request_id` that ended the hunt. The
delete path had the same problem and worse: `.catch(() => {})` let the node
vanish from the map while the row survived, so re-adding the same name
quietly *merged* into the undeleted row.

### Verification

- backend suite green on this branch; `ruff check` clean
- port guard tested both ways (occupied → `SystemExit`, free → returns)
- `tsc --noEmit` clean on the identical `Learn.tsx`
- vitest cannot start on the local Node 20 (rolldown `styleText` crash), so
  the new specs are CI-verified only

The backend was never at fault: add → delete → re-add of the same name
creates a fresh row every time — verified against the real project.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concept creation by retaining the previously selected course when focus is cleared.
  * Added clearer error messages when saving concepts or resolving a course fails.
  * Failed concept deletions now restore the affected content and connections when appropriate.
  * Added safeguards to prevent starting the development server when its configured port is already in use.
* **Tests**
  * Added coverage for course resolution after concepts are deleted and re-added.


Jose-Gael-Cruz-Lopez:
## Code review — learn concept re-add + port guard

The diagnosis here is excellent — the "stale process on :5000" write-up is the most useful thing in the PR, and `_refuse_if_port_taken` is correctly scoped (dev `__main__` only; `backend/Dockerfile:37` runs `uvicorn main:app`, so containers never reach it). Replacing the delete path's `.catch(() => {})` with a real restore is the right instinct. But the headline fix does not reach the reported flow: `addConcept`'s new fallback is unreachable, because the composer that calls it is gated on the very value the fallback exists to work around. There is also a new failure mode in the delete restore (a 404 resurrects the node), a hand-rolled error formatter that duplicates `lib/errorMessage.ts` and prints raw JSON bodies into toasts, and ~70 lines of undocumented quiz deep-link feature in a PR whose description says it was split out to *exclude* quiz work.

### Findings

#### P0

**[P0] `addConcept`'s new fallback is unreachable, and the reported bug survives** — `frontend/src/components/screens/Learn.tsx:1754`

```tsx
              {/* Add concept */}
              {cardCourseId && (
                <div style={{ padding: "4px 22px 24px" }}>
```

Both `addConcept` call sites live inside that block — `:1763` (`onKeyDown` Enter) and `:1780` (the Add button) — and there is no other caller in the file. So `cardCourseId` is always truthy when `addConcept` can run, which means `resolveAddConceptCourseId(cardCourseId, lastCourseIdRef.current)` at `:1226` always returns `cardCourseId`. The `lastCourseIdRef` + effect (`:474-477`) and the new `toast.error("Pick a course first…")` branch (`:1227-1230`) are dead code.

Worse, the reported repro still fails. `removeConcept` is only reachable from the focus card — `:1611` gates the button row on `focusConcept`, Remove is `:1663-1683` — so the deleted node is always the focused one. After removal, `activeFocusId = focusedNodeId ?? highlightId` (`:452`) and `highlightId` is `graphNodes.find(n => n.name.toLowerCase() === topic.trim().toLowerCase())?.id` (`:446`) — the node just deleted. With `selectedCourseId` at its default `""` (`:392`/`:397`, `searchParams.get("course") ?? ""`), `resolveCardCourseId` (`:269-274`) returns `null` and the whole "＋ Add concept" affordance unmounts (block `:1754-1815`). The student cannot type the name back in at all, which is exactly "when deleting a concept node, i should be able to add it back of the same name by add concept".

The new specs pass because they call the exported resolver directly and never render the gate. The gate is where the fix belongs: resolve the add-course once and gate the composer on *that*, showing which course the concept will land in. (CodeRabbit raised the gate; this adds the proof that the fallback branch is unreachable and that the original repro is unchanged.)

#### P1

**[P1] A 404 on delete resurrects the node and toasts a false statement** — `frontend/src/components/screens/Learn.tsx:1304-1318`

```tsx
    deleteGraphNode(userId, nodeId).catch((err: unknown) => {
      console.error("[removeConcept] failed", { nodeId, err });
      if (!removedNode || /^(node-new-|stream-)/.test(nodeId)) return;
      setGraphNodes(prev => (prev.some(n => n.id === nodeId) ? prev : [...prev, removedNode]));
      ...
      const reason = err instanceof ApiError ? ` (${err.status})` : "";
      toast.error(`Couldn't delete “${removedNode.name}”${reason} — it's still on your map.`);
```

The restore fires on *any* rejection, and 404 is precisely the status the endpoint returns for "the row is already gone": `backend/routes/graph.py:119-125` raises `HTTPException(status_code=404, …)` whenever `delete_node` returns an error, and `backend/services/graph_service.py:486-492` returns `{"error": "Node not found", "deleted": False}` when the owner-scoped select finds no row. So a second tab, a client-side node that is already stale, or a delete that already succeeded server-side now puts a phantom concept back on the map and tells the user "it's still on your map" when it is not. The old `.catch(() => {})` produced the correct end state in exactly this case.

It compounds: delete X → 404 → user re-adds X (fresh row, new id) → the catch appends the old node object under its dead id and the map shows two "X" nodes, one of which doesn't exist. A 404 should be treated as success — nothing to restore. (The non-404 case is already handled correctly: the `prev.some(n => n.id === nodeId)` guard stops a double-insert when a re-add merged back into the surviving row.)

#### P2

**[P2] Hand-rolled error copy duplicates `lib/errorMessage.ts` and dumps the raw response body into a toast** — `frontend/src/components/screens/Learn.tsx:1279-1285`

```tsx
        const reason =
          err instanceof ApiError
            ? `${err.status}${err.message ? ` — ${err.message.slice(0, 160)}` : ""}`
            : err instanceof Error && err.message
              ? err.message.slice(0, 160)
              : "the request never completed";
        toast.error(`Couldn't save “${label}” (${reason}).`);
```

`ApiError.message` is documented as the raw body — `frontend/src/lib/api.ts:25-26`: "`message` stays the raw body for backward compatibility" — so the toast renders `Couldn't save “Markov Chains” (404 — {"detail":"Not Found"}).`, or 160 characters of a proxy's HTML error page. `frontend/src/lib/errorMessage.ts` exists for exactly this and says so in its header: "Rendering that with `String(err)` dumps JSON into the UI, which is what these helpers exist to prevent." Its `humanizeError(err, fallback)` (`:156-160` — "A short sentence safe to put in a toast. Never returns … a raw JSON body") is already used across Gradebook, Calendar, Social, Study, ProfileView, the notetaker, SignInModal and QuizPanel; `Learn.tsx` imports nothing from it. The Engineering Style Guide's shared-primitive rule applies here the same way it does to `components/ui/`. It also removes the inconsistency CodeRabbit flagged — the delete toast at `:1317` surfaces only the status and never the detail; both paths would read the same through one helper.

**[P2] An undocumented quiz deep-link feature ships in a PR that says it excludes quiz work** — `frontend/src/components/screens/Learn.tsx:288-297`, `:1449-1457`, `:1640-1662`

```tsx
export function quizHref(
  conceptId: string | null | undefined,
  topic: string | null | undefined,
): string {
```

`quizHref` plus two new "Quiz me" buttons (session toolbar `:1449`, focus card `:1640`) plus a five-case spec block (`Learn.graph.test.ts:171-206`) are roughly a quarter of the diff, and neither the title, the description, nor the release-note summary mentions them. The description says the opposite: "Split out of #534, which was carrying these three alongside unrelated quiz work." The code itself checks out — `?concept=`/`?topic=` match `screens/Quiz.tsx:86-94`, `resolveInitialSelection` (`lib/quizSelection.ts:71-75`) degrades to `{null, null}` on an id it can't resolve, `Icon` has `flask` (`Icon.tsx:78`), `.btn--sm` exists (`globals.css:233`), and #534 carries no duplicate — but a new user-facing entry point riding along in a bugfix PR makes this hard to review and impossible to revert cleanly. Either name it in the description or move it out.

#### P3

**[P3] A failed delete restores the node but not the focus** — `frontend/src/components/screens/Learn.tsx:1302`

```tsx
    setFocusedNodeId(cur => (cur === nodeId ? null : cur));
```

The catch at `:1304-1318` undoes the node and edge removals but never the focus clear, so after a failed delete the concept reappears on the map while the rail's focus card is gone — and, per the P0 above, so is the Add-concept composer.

**[P3] The port probe only covers IPv4 loopback** — `backend/main.py:338-341`

```python
    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as probe:
        probe.settimeout(0.4)
        if probe.connect_ex(("127.0.0.1", port)) != 0:
            return
```

`uvicorn.run(..., host="0.0.0.0")` at `:359` binds the IPv4 wildcard so the common case is covered, but a stale server bound to `::1` or to one specific interface passes the probe silently. Given the whole point is "tell the next person", resolving both families would close the gap.

### What's good

- The root-cause write-up in the description (two owning PIDs, the uvicorn reload child holding the socket, "if a route you can read in `routes/` is missing from `/openapi.json`, you are talking to a different build") is worth more than the code — it belongs in `CLAUDE.md` or `docs/local-supabase.md` so it outlives the PR.
- The guard is correctly classified and scoped: it is a dev-entrypoint TCP probe under `if __name__ == "__main__"`, not a production origin/port check. `backend/Dockerfile:37` is `CMD ["uvicorn", "main:app", …]` and `scripts/e2e-up.sh` invokes uvicorn directly, so neither containers nor the e2e lane can trip it.
- The backend claim in the description holds: `delete_node` (`graph_service.py:481-509`) hard-deletes edges in both directions and then the row, with no soft-delete column and no unique-constraint collision on re-add — so add → delete → re-add really does create a fresh row.

**Verdict: request changes** — the P0 leaves the reported bug reproducible on the default path, and the P1 introduces a new map/server divergence in the opposite direction.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Blocker
- **The add-concept fallback was unreachable and the reported bug still reproduced.** The composer was gated on `cardCourseId` — the very value the fallback exists to work around — and both `addConcept` call sites live inside that gate, so `resolveAddConceptCourseId` could never return anything else and the whole "＋ Add concept" affordance unmounted after deleting the focused concept. One resolved `addCourseId` now drives both the gate and `addConcept`, and the destination course is surfaced in the UI instead of being applied silently.

### Major
- **A 404 on delete resurrected the node.** 404 is exactly what the endpoint returns for "already gone" (`routes/graph.py` → `graph_service.delete_node`), so the restore produced a phantom concept and a toast claiming "it's still on your map" when it was not — compounding to two visible "X" nodes after a re-add. New `shouldRestoreFailedDelete` treats 404 as success; the genuine-failure path keeps its double-insert guard.

### Minor / nits
- Both toasts now go through `lib/errorMessage.ts::humanizeError` instead of printing `ApiError.message` — the documented *raw body* — into the UI. Also resolves the status-only vs body asymmetry between the add and delete paths.
- A failed delete restores focus, not just the node and edges.
- The dev port guard probes `::1` and `getaddrinfo` results as well as `127.0.0.1`, and names the answering address.

### Note
The `quizHref` / "Quiz me" deep-link code was left in place — it is correct, but it is not mentioned in the PR description, which says quiz work was split out. Worth naming in the body or moving out.

**Verification** — `ruff check .` clean · **1869 passed, 49 skipped** · `tsc` clean · **638 frontend tests pass** · eslint 0 errors

PR #539, opened 2026-08-12
fix(graph): gate 3D mode on WebGL2 and contain renderer crashes

### LINKED ISSUE #538: Dashboard crashes to the root error fallback on browsers without WebGL when 3D graph mode is persisted
## Symptom

Landing on the dashboard (or Tree / Learn — any surface that renders the knowledge graph) in a browser **without WebGL** replaces the entire app with the root error fallback ("We hit a snag"), with `Error creating WebGL context.` in the console. "Try again" re-throws immediately. Reported on staging; applies everywhere.

Only browsers whose profile has `localStorage["sapling.kg.mode"] = "3d"` are affected — a fresh no-WebGL profile defaults to 2D and works. That's why it "used to work": the stored preference, not a code regression, is what flips the failure on.

## Root cause

1. `frontend/src/components/graph/KnowledgeGraph.tsx` picks 2D vs 3D purely from the persisted localStorage mode — there is **no WebGL capability check anywhere in the frontend**.
2. In 3D mode, `react-force-graph-3d` mounts its engine synchronously in a `useLayoutEffect` (react-kapsule), constructing `new three.WebGLRenderer(...)` (three-render-objects).
3. three.js r163+ **throws** `Error('Error creating WebGL context.')` from the constructor when `canvas.getContext('webgl2')` returns null.
4. The nearest error boundary is the **root** one (`app/layout.tsx`), so the whole app unmounts into `ErrorFallback` — the graph has no local boundary.

## Fix

Both changes in the `KnowledgeGraph` wrapper so Dashboard, Tree, and Learn are all covered:

1. **Capability check** — probe `canvas.getContext('webgl2')` once before honoring the stored `"3d"` mode; without WebGL2, force 2D and disable the 3D toggle (e.g. title "3D requires WebGL"). The stored preference is preserved for when the user is back on a WebGL-capable browser.
2. **Local error boundary** — wrap the graph slot so any renderer failure (e.g. context creation failing despite a successful probe: GPU process crash, context-limit exhaustion) degrades to the 2D graph inline instead of blanking the whole app.

Regression coverage: jsdom has no WebGL2, so an unmocked wrapper mount with `"3d"` stored is a natural unit test for the capability check; an e2e journey can simulate a no-WebGL browser by overriding `HTMLCanvasElement.prototype.getContext` via `addInitScript`.

Workaround meanwhile: `localStorage.removeItem('sapling.kg.mode')` in the affected browser.

AndresL230:
Fixes #538.

## The bug

A browser without WebGL that lands on the dashboard (or Tree/Learn) with `localStorage["sapling.kg.mode"] = "3d"` persisted collapses the **entire app** into the root error fallback ("We hit a snag"), permanently — "Try again" re-throws. Chain:

1. The `KnowledgeGraph` wrapper honoured the persisted `"3d"` unconditionally — no WebGL capability check existed anywhere in the frontend.
2. `react-force-graph-3d` mounts its engine synchronously in a `useLayoutEffect` (react-kapsule), constructing `new three.WebGLRenderer(...)`.
3. three.js r163+ **throws** `Error creating WebGL context.` from the constructor when `canvas.getContext("webgl2")` returns null.
4. The nearest boundary was the root one (`app/layout.tsx`), so the whole tree unmounted.

A profile that toggled 3D on a WebGL-capable browser was locked out of the app the moment WebGL went away (disabled, GPU blocklist, remote desktop, VM). Fresh profiles were unaffected (mode defaults to 2D) — which is why this read as "used to work".

## The fix (both in the wrapper, so Dashboard/Tree/Learn are all covered)

1. **WebGL2 capability gate** (`webgl2Available()`): probe `canvas.getContext("webgl2")` once per wrapper mount (probe context released via `WEBGL_lose_context` so probes never count toward the per-page context cap). Without WebGL2:
   - a persisted `"3d"` is **ignored, not rewritten** — the preference survives for the profile's WebGL-capable browsers;
   - the mode toggle renders disabled with title/aria-label "3D requires WebGL";
   - the storage/custom-event mode-sync paths and the setter are gated identically.
2. **Graph-local error boundary**, keyed by mode: any residual renderer crash (context creation failing despite a successful probe — GPU process crash, context-limit exhaustion) degrades to the 2D graph inline and heals the persisted mode to `"2d"`; the heal flips the boundary key, clearing the error state into a normal 2D render. A (never-observed) 2D crash renders a minimal placeholder instead of remounting the thing that just threw.

## Tests

- **Unit (`KnowledgeGraph.test.tsx`, 6 tests, written first and watched fail)**: fresh-profile default pin; 3d-honoured-with-WebGL pin; forces-2D-without-WebGL2; toggle disabled with explanatory label; toggle enabled with WebGL; crash containment + mode heal (RED run reproduced the exact production throw escaping uncaught).
- **Promoted journey (`e2e/graph-webgl-fallback.spec.ts`)**: init script nulls the webgl/webgl2 branches of `getContext` (exactly the API three probes; 2D canvas stays real) and seeds the poisoned `"3d"` mode → dashboard must render the 2D SVG graph, no root fallback, toggle disabled, preference preserved.

## Verification

- `vitest run`: 72 files / 624+ tests green (against a jsdom that is itself a no-WebGL environment, so every existing screen test now exercises the gated path).
- `tsc --noEmit` clean; `eslint` clean on changed files (no new suppressions).
- `next build` clean.
- **Pre-fix RED run (stack cycle 1)**: with the wrapper fix stashed and the real production build served, the new journey fails — the dashboard collapses into the root fallback exactly as reported. The journey failed at the `"We hit a snag"` count-0 assertion (exit 1) — the root fallback was live on the page.
- **Post-fix full lane (stack cycle 2)**: entire Chapter-1 Playwright lane + deterministic oracles green. 43 passed (1.5m), oracles exit 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graph rendering when WebGL2 is unavailable or 3D rendering fails.
  * Automatically falls back to the 2D graph while preserving the saved mode preference.
  * Disabled the 3D toggle when unsupported, with an explanatory message.
  * Added recovery and retry handling for temporary graph rendering failures.
  * Prevented graph errors from disrupting the rest of the dashboard.

* **Tests**
  * Added coverage for WebGL detection, fallback behavior, mode synchronization, toggle states, and renderer recovery.
  * Added end-to-end coverage for 2D fallback and supported 3D environments.

### DIFF
--- a/docs/frontend-testids.md
+++ b/docs/frontend-testids.md
@@ -207,6 +207,8 @@ route:
 | `graph-add-concept` | Tree toolbar: "＋ Add concept" opener (#330) — rendered only when a single course pill is selected (the "all" filter gives no course to attribute the node to) |
 | `graph-add-concept-input` | the concept-name `<input>` (Enter submits, Escape cancels) |
 | `graph-add-concept-submit` | the "Add" button — POSTs create-or-merge, toasts, reloads the graph |
+| `graph-crash-fallback` | graph-local crash placeholder (#538) — rendered by the wrapper's error boundary when a renderer throws; never contains a graph |
+| `graph-crash-retry` | "Try again" button inside the crash placeholder — wired to the boundary's reset |
 
 `graph-node-item` / `graph-node` carry the node id as a separate
 `data-node-id` attribute instead of a testid suffix (the repeated-items rule
@@ -221,6 +223,7 @@ never collide in the DOM.
 | testid | element |
 | --- | --- |
 | `app-shell` | authed shell root — the scrolling `<main id="main-content">` in `ShellFrame.tsx`, present in both (top-nav and sidebar) layout variants; the Playwright harness smoke spec (#385) anchors on it as the "authed shell mounted" signal |
+| `error-fallback` | root `ErrorFallback` surface (`ErrorBoundary.tsx`) — the whole-app "We hit a snag" page; journeys assert its ABSENCE (#538) so the check survives copy rewording |
 
 ### `social`
 
--- a/frontend/e2e/graph-webgl-fallback.spec.ts
+++ b/frontend/e2e/graph-webgl-fallback.spec.ts
@@ -1,40 +1,54 @@
 /**
- * Journey #538 — a browser without WebGL lands on the dashboard without
- * crashing, even when the persisted graph mode is "3d".
+ * Journey #538 — WebGL capability and the knowledge-graph mode toggle.
  *
  * THE BUG: the KnowledgeGraph wrapper honoured localStorage's
  * `sapling.kg.mode = "3d"` unconditionally. three.js r163+ THROWS from the
  * WebGLRenderer constructor when `canvas.getContext("webgl2")` returns null,
  * the throw happens in a mount-time layout effect, and the only error
  * boundary above it was the ROOT one (app/layout.tsx) — so the entire app
- * unmounted into the "We hit a snag" fallback. A profile that toggled 3D on
- * a WebGL-capable browser was permanently locked out of the dashboard the
+ * unmounted into the root error fallback. A profile that toggled 3D on a
+ * WebGL-capable browser was permanently locked out of the dashboard the
  * moment WebGL went away (disabled, GPU blocklisted, remote desktop, VM).
  *
- * THE FIX (KnowledgeGraph.tsx): probe webgl2 before honouring the persisted
- * "3d" (ignore it — do NOT rewrite it — when unavailable), disable the mode
- * toggle with an explanatory label, and contain any residual renderer crash
- * in a graph-local error boundary that degrades to the 2D graph.
+ * THE FIX (KnowledgeGraph.tsx): the wrapper derives the EFFECTIVE mode from
+ * the persisted wish + a WebGL2 capability probe — ignoring (never
+ * rewriting) a "3d" wish the browser can't honour — marks the toggle
+ * aria-disabled with the reason reachable, and contains residual renderer
+ * crashes in a graph-local boundary.
  *
- * Simulation: an init script nulls the webgl/webgl2/experimental-webgl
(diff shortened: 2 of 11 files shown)


AndresL230:
Applied all 15 code-review findings in d6cfaea7 (design rework where the review asked for one, not point patches):

- **Crash containment redesigned (findings 1–5)**: the persisted mode is now never rewritten. The wrapper derives one EFFECTIVE mode from the raw persisted wish + a module-memoized WebGL2 probe; a 3D crash re-probes capability (the only signal the cached verdict may be stale) — capability really gone flips the effective mode and re-keys the boundary so 2D mounts fresh; transient crashes show a static placeholder with the boundary's reset wired to "Try again". The fallback never renders another graph component (a throw in a boundary's own fallback render is uncatchable by that boundary), and renderer crashes now `console.error` in production.
- **A11y (11)**: `aria-disabled` + `aria-describedby` replace native `disabled`; the toggle stays focusable, keeps its action name, and the "3D requires WebGL" reason is reachable.
- **Simplification (9, 10)**: the gate rule has one render-time encoding (`effective = webglOk ? mode : "2d"`); the probe is memoized with an explicit `force` re-probe path.
- **Test hardening (6, 7, 8, 12, 13, 14, 15)**: crash sentinels throw from a layout effect (the real crash phase); `vi.spyOn` getContext stub actually restores; exported storage-key/sync-event constants (journey mirrors by comment); direct SYNC_EVENT gating coverage; journey negatives anchor on the new `error-fallback` testid with an end-state re-assert; a real-browser positive-path test pins the toggle enabled under genuine WebGL2; the four divergent `next/dynamic` passthrough mocks collapsed into one shared hook-safe helper.

Verification: unit suite 627/627, eslint 0 errors (stale suppressions pruned), tsc clean, full e2e lane 44/44 (both #538 journeys) + oracles clean, run from a fresh worktree via one flock'd up→test→down cycle.

PR #562, opened 2026-08-14
fix(tutor): stop the tutor repeating itself, and let it actually read course materials

Darkest-Teddy:
Two independent bugs in the tutor chat, found while debugging "why does the AI repeat the message instead of responding to it". One commit each.

## 1. The tutor replayed its previous reply verbatim

`stream_agent_turn` preferred `run_result.output` over the streamed chunks:

```python
reply = final_output if final_output is not None else joined
```

`.output` resolves out of the run's message list, and that list includes `message_history`. So a turn whose model response carries **no text part** — the model ended its turn after tool calls — handed back the *previous* turn's assistant message. Fully formed and non-blank, so the blank-reply ladder immediately below never saw it. The route persisted it, and the student got a byte-identical copy of the last answer.

Confirmed in two live sessions two days apart: same sha256, same length, including a 1757-character reply reproduced exactly against a completely different question.

**Fix:** trust `final_output` only when this turn actually streamed text. Text always arrives as `PartStart`/`PartDelta` events, so "nothing streamed" means "this turn produced no text" — degrade through the existing blank-reply ladder rather than replaying history.

**Evidence.** Replaying a real session turn against `gemini-2.5-flash-lite`, the correlation is exact:

| streamed `token/reply` | outcome | count |
|---|---|---|
| 0 | repeated the prior answer | 2/8 |
| >0 | answered correctly | 6/8 |

After the fix: 0 repeats in 10 runs.

**What this does not fix:** the model still ends ~40% of these turns without text. Those now take the Rung-1 JSON fallback (a real answer), or — if graph writes already landed — show the honest "interrupted, please retry". A silent wrong answer becomes a correct one or a visible error. Why the model does this at all is unexplored and worth a follow-up.

## 2. The tutor never read any course document

`search_course_materials` filtered `documents` on `course_id`. That table keys on `offering_id` (0025) and has no `course_id` column, so PostgREST 400s on every call — and the tool's deliberate degrade-silently `except` turned that into `[]`.

The tutor answered from base knowledge alone, which presents as "it's generic about my class" rather than as a failure. Nothing was logged at the user's level.

**Fix:** resolve the abstract course to the user's offerings via `academics.user_offering_ids_for_course`, matching the convention that the API boundary keeps the abstract course while documents key on the offering. The `#125` user_id scope is unchanged — documents stay user-scoped *within* a shared offering.

## Tests

Both regression tests were written first and watched fail for the right reason:

- `test_textless_turn_never_replays_the_previous_turns_reply` — failed with `on_complete` receiving the stale prior-turn text.
- `test_documents_are_fetched_by_offering_not_course_id` — failed with `column documents.course_id does not exist`.

The second one needed a schema-faithful fake. The existing mocks accept any filter, which is precisely how bug 2 survived them. The two older `search_course_materials` test classes now stub the offering lookup, which became a real dependency of that function.

Full backend suite: **1999 passed, 56 skipped**, exit 0. `ruff check` clean on all four files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Course material searches now return documents associated with the user’s course offerings while maintaining access restrictions.
  * Prevented tool-only chat turns from displaying or saving stale assistant responses from earlier messages.
  * Ensured empty or non-text streaming responses use a fresh fallback reply.

* **Tests**
  * Added coverage for offering-based material searches, access scoping, and stale-response prevention.


Jose-Gael-Cruz-Lopez:
## Code review — tutor repetition + course-material retrieval

Two genuinely independent bugs, both diagnosed properly, both with regression tests written first. The `documents.course_id` finding is correct — migration `0025_study_integrity.sql` recreates `documents` with `offering_id TEXT NOT NULL REFERENCES course_offerings(id)` and no `course_id` column, so the old filter really did 400 into the tool's degrade-silently `except`. The `chat_stream` guard is a one-line narrowing that is safe for this agent (`chat_tutor` is `output_type=str`, `agents/chat_tutor.py:174`, so its reply always arrives as text parts — there is no output-tool channel the guard could discard). Three things need attention before merge: the now-working documents query no longer honours the soft-delete convention every other reader follows, the offering resolution it picked can only *shrink* the user's own document set relative to the writer, and the identical stale-`run_result.output` read on the JSON path — which is exactly the fallback this PR now routes textless turns into — was left unguarded, so the repeat symptom can still be produced.

### Findings

#### P0

**[P0] CI is red — `Backend (pytest)` fails (pre-existing on `main`, not a regression from this PR)** — run [31763538674](https://github.com/SaplingLearn/Sapling/actions/runs/31763538674)

```
FAILED tests/test_quiz_preflight_a.py::TestQuizErrorEnvelope::test_method_not_allowed_gets_generic_code - AttributeError: '_IncludedRouter' object has no attribute 'path'
1 failed, 1965 passed, 56 skipped, 13 warnings in 42.25s
```

```
            if match == Match.PARTIAL:
>               route = starlette_route.path
                        ^^^^^^^^^^^^^^^^^^^^
E               AttributeError: '_IncludedRouter' object has no attribute 'path'
/opt/hostedtoolcache/Python/3.13.14/x64/lib/python3.13/site-packages/opentelemetry/instrumentation/fastapi/__init__.py:495: AttributeError
```

Not this PR's fault. The identical failure is on `main` — run 31975087459 (`4f5a7f90`, "Merge pull request #524 from SaplingLearn/feat/landing-v5-port"): same test, same `AttributeError`, `1 failed, 1963 passed, 56 skipped`. It is dependency drift in `opentelemetry-instrumentation-fastapi._get_route_details`, which assumes every `Match.PARTIAL` route object has `.path`; FastAPI's `_IncludedRouter` does not. None of this PR's four files touch routing or quiz. It still blocks the merge queue, so it needs a pin/fix on `main` first — worth noting because the PR body claims "Full backend suite: 1999 passed, 56 skipped, exit 0", which does not match CI.

#### P1

**[P1] Soft-deleted documents now reach the tutor** — `backend/agents/tools/chat_context.py:154-164`

```python
            return (
                table("documents").select(
                    "id,file_name,summary,concept_notes",
                    filters={
                        # #125 user scope is preserved: documents are
                        # user-scoped WITHIN a shared offering.
                        "offering_id": f"in.({','.join(offering_ids)})",
                        "user_id": f"eq.{user_id}",
                    },
                    order="created_at.desc",
                )
```

`documents` is soft-deleted, not hard-deleted — `routes/documents.py:288-293`: *"Soft delete (0025): stamp `deleted_at`; reads filter it out."* Every other reader honours that: `routes/documents.py:241`, `routes/study_guide.py:137`, `routes/flashcards.py:153` and `:163`, `routes/documents.py:1230` all pass `"deleted_at": "is.null"`. This query does not. Before this PR that was invisible because the query 400'd and returned `[]` unconditionally; this PR is what makes it return rows, so it is this PR that first exposes deleted documents. Concretely: a student deletes a document from their Library, `documents.deleted_at` is stamped, the row stays, and the tutor keeps decrypting its `summary` + `concept_notes` into the LLM context on every `search_course_materials` call. Nothing in `backend/agents/` or the tutor tests references `deleted_at`, so no test catches it.

**[P1] The stale `run_result.output` read is unguarded on the JSON path — the very fallback this PR routes textless turns into** — `backend/routes/learn.py:636-654`

```python
        result = record_agent_usage(
            await agent.run(user_message, **run_kwargs),
            feature="chat_tutor", task="chat_tutor", user_id=deps.user_id,
        )
        reply = result.output  # str — chat_tutor agents return plain Markdown.

        if not reply.strip():
```

`_prepare_chat_run` puts `"message_history": message_history` into `run_kwargs` (`routes/learn.py:577`), so `result.output` here resolves out of the same history-bearing message list the PR indicts. The only guard is `not reply.strip()` — and the PR's whole point is that a replayed prior turn is *"fully formed, non-blank, and therefore invisible to the blank-reply ladder"*. `_chat_turn_json` is both a first-class route (`POST /api/learn/chat`, `routes/learn.py:745-751`) and the streamed route's Rung-1 `nonstream_fallback` (`routes/learn.py:823`). So after this change a textless streamed turn correctly refuses the stale text, degrades to `_chat_turn_json`, and if that re-run is also textless — the PR body measures ~40% — it takes the prior assistant message from `message_history`, persists it via `save_message` (`routes/learn.py:727`) and returns it. The repeat is not eliminated, it is moved one rung down, onto a path this PR sends *more* traffic to.

**[P1] Offering resolution is narrower than the writer's and than every sibling reader's** — `backend/agents/tools/chat_context.py:151-153`

```python
            offering_ids = user_offering_ids_for_course(user_id, course_id)
            if not offering_ids:
                return []
```

`user_offering_ids_for_course` (`services/academics.py:185-200`) intersects the course's offerings with the user's `enrollments` rows. Documents are written with a different resolver — `offering_id = resolve_offering(course_id, create=True)` at `routes/documents.py:572` and `:710` — which picks the *current-term* offering (creating it if absent, else falling back to any offering of the course) and never consults `enrollments`. The two sibling readers use the writer's resolver too: `routes/study_guide.py:136` and `routes/flashcards.py:144` filter on `resolve_offering(...)`. Divergence is reachable across a term boundary: a student enrolled via `routes/onboarding.py:95-112` in Fall-26's offering who uploads in the next term gets `documents.offering_id` = the new term's offering, has no enrollment row for it, and the tutor silently returns `[]` — while the Library still lists the file (`routes/documents.py:239-241` filters on `user_id` only). Note the enrollment intersection buys no safety: `user_id` is already the security boundary (your own `#125` comment says so), so this filter can only remove the user's *own* documents.

#### P2

**[P2] The empty-offering short-circuit is silent — the exact failure mode this PR exists to remove** — `backend/agents/tools/chat_context.py:152-153`

```python
            if not offering_ids:
                return []
```

The PR body's own indictment of the old bug is *"The tutor answered from base knowledge alone, which presents as 'it's generic about my class' rather than as a failure. Nothing was logged at the user's level."* This early return reproduces that precisely: no log, no metric, indistinguishable from "this course genuinely has no materials". The `except` below it does `logger.exception`; this branch gets nothing. Given the P1 above makes it reachable, a log line here is what turns the next occurrence of this class of bug into a five-minute diagnosis instead of a two-month one.

#### P3

**[P3] Two extra uncached PostgREST round-trips per tool call, in the streaming hot path** — `backend/services/academics.py:191-200` via `chat_context.py:151`

```python
    offs = table("course_offerings").select(
        "id", filters={"course_id": f"eq.{course_id}"}
    ) or []
    off_ids = {o["id"] for o in offs}
    if not off_ids:
        return []
    enr = table("enrollments").select(
        "offering_id", filters={"user_id": f"eq.{user_id}"}
    ) or []
```

`user_offering_ids_for_course` is deliberately uncached and the second select pulls *all* of the user's enrollments across every course. `search_course_materials` is an LLM-callable tool that can fire more than once per turn, on the latency-critical SSE path — so this is +2 sequential round-trips per call. The `documents` select that follows is also unbounded (no `limit`, `order="created_at.desc"` only) and every returned row is AES-decrypted before the list is truncated to `limit` (5) in Python at `chat_context.py:213`. Harmless today; both become real once a course accumulates documents.

**[P3] No test pins the newly-reachable textless-turn-with-writes branch** — `backend/tests/test_chat_stream.py:710-750`

```python
        agent = FakeAgent([
            FunctionToolCallEvent("read_graph_neighborhood"),
            FunctionToolResultEvent(),          # no writes landed
            AgentRunResultEvent(PRIOR),         # stale: from message_history
        ])
```

The new test only covers the no-writes path (Rung-1 fallback). But the trigger it describes is "the model ended its turn after tool calls", and every tutor agent registers `apply_graph_update_tool` and `update_mastery_tool` (`agents/chat_tutor.py:161-162`), which populate `deps.graph_updates` / `deps.mastery_changes` (`agents/tools/graph.py:122`, `:174`). So the *likely* shape of a textless turn is textless-with-writes, which lands on the terminal `retryable: False` error at `chat_stream.py:362-368` — a user-visible dead end, and the branch with no new coverage. Worth one more `FakeAgent` case with an `on_fire` write.

### What's good

- The root-cause work is real and checkable: `0025_study_integrity.sql:15-31` confirms `documents` has `offering_id NOT NULL` and no `course_id`, exactly as claimed.
- The new `TestSearchCourseMaterialsOfferingScope` fake raises on a `course_id` filter instead of accepting any filter — the right response to "the existing mocks accept any filter, which is precisely how bug 2 survived them".
- The `chat_stream.py` guard is correctly scoped: `chat_tutor` is `output_type=str` (`agents/chat_tutor.py:174`), so there is no output-tool channel whose reply the `joined.strip()` condition could throw away. Per the Sapling Engineering Style Guide the new sync `user_offering_ids_for_course` call also stays off the event loop — it sits inside `_fetch`, which runs under `asyncio.to_thread` (`chat_context.py:174`).

**Verdict:** request changes — the soft-delete regression and the unguarded `result.output` on the JSON fallback path both need fixing before merge; the CI failure is pre-existing on `main` and should be unblocked there.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Major
- **Soft-deleted documents reached the tutor.** Added `deleted_at is.null` to the `documents` select. This PR is what first exposed them (the query used to 400 into the silent degrade), so a deleted file's decrypted `summary`/`concept_notes` would have kept flowing into LLM context. Filter set is now byte-identical to #534's, making that overlap a trivial pick-either.
- **The stale `run_result.output` read was only narrowed on the streamed path.** `_chat_turn_json` is both a first-class route and the streamed route's Rung-1 fallback — i.e. exactly where this PR now sends textless turns — so the repeat was relocated, not removed. Added `_new_run_text()` (the non-streaming twin of the `joined.strip()` narrowing) and applied it to both history-bearing `.output` sites; the `message_history=[]` site is left alone with a comment saying why.
- **Offering resolution was narrower than the writer's.** Now the union of `user_offering_ids_for_course` and `resolve_offering`, so a document uploaded across a term boundary is still found. Test covers the cross-term case.

### Minor
- The empty-offering short-circuit logs a warning instead of returning `[]` silently — the exact failure mode this PR exists to remove.
- `limit` pushed into the query so rows are no longer AES-decrypted and then discarded.
- Added the missing textless-turn-**with-writes** case (the likely shape, landing on the terminal `retryable: False` path).

### Not fixed here
The red `Backend (pytest)` is the otel `_IncludedRouter` failure that is pre-existing on `main`; #563 carries the shim that fixes it.

**Verification** — `ruff check .` clean · **1973 passed, 56 skipped**

Jose-Gael-Cruz-Lopez:

### No schema change — but the behaviour change is data-dependent

Two fixes here only show their value against real rows, so they are worth confirming on staging.

### 1. How much was actually leaking

The `documents` read had no `deleted_at` filter. Before this PR the query 400'd and returned `[]`, so nothing leaked; **this PR is what makes it return rows**, which is why the filter had to be added in the same change.

```sql
SELECT count(*) FROM documents WHERE deleted_at IS NOT NULL;
```

That count is exactly how many soft-deleted documents the tutor would otherwise have decrypted into LLM context. If it is non-zero, this fix is load-bearing rather than theoretical.

### 2. The cross-term offering gap is real or it isn't

The offering set is now the union of `user_offering_ids_for_course` (enrollment-derived) and `resolve_offering` (the writer's resolver), because documents are written with the latter and read with the former.

```sql
-- documents sitting on an offering the owner has no enrollment row for:
-- these are exactly the files the tutor could not see before this change
SELECT count(*) FROM documents d
WHERE d.deleted_at IS NULL
  AND NOT EXISTS (
    SELECT 1 FROM enrollments e
    WHERE e.user_id = d.user_id AND e.offering_id = d.offering_id);
```

Non-zero means the divergence is already live, not just reachable at the next term boundary.

### 3. Sanity-check the filter set on the wire

The final query filters on exactly `offering_id in (...)`, `user_id eq`, `deleted_at is.null`. Worth eyeballing one real request in logs to confirm no fourth filter crept in — this is the query whose column set was wrong for months without anyone noticing.

AndresL230:
Pushed two things to this branch to get it merge-ready (on the repo owner's instruction): a merge of current `main`, and one commit of review fixes (`4cfe427a`). Both root-cause fixes in this PR were confirmed in review and are unchanged: the replayed reply (pydantic-ai's `_recover_text_from_message_history` walking the history we pass in) and the dead `documents.course_id` filter.

**What the new commit changes**

- **Offering scope.** `_material_offering_ids` (enrolled offerings + today's current-term offering) was still narrower than the writer. Uploads are stamped with `resolve_offering` at upload time, so: enrolled Fall, uploaded Spring, asked in Summer → neither resolver names Spring and those uploads vanish from the tutor while the Library still lists them. It now uses `academics.course_offering_ids`, which #563 added to main for exactly this shape after this branch was cut. One read instead of four or five per tool call; `None` ("could not tell") is logged, never guessed at. `user_id` stays the #125 boundary, so the wider scope can only re-include the student's own uploads.
- **Empty-result signal.** The tool wrapper now routes an empty result through `report_empty_result` (the CLAUDE.md convention, also newer than this branch). New `Expect.HAS_DOCUMENTS` probe, course-scoped through the `documents → course_offerings` FK inside PostgREST rather than by the offering ids the tool just resolved, because a probe that shares the read's keyspace can only ever agree with it. Skipped when a retrieval is injected (evals).
- **Real-DB test.** `tests/integration/test_course_materials_keyspace_db.py` runs the real tool and probe against the real schema: the seeded syllabus is reachable, the rollover-shaped document is found, another student's documents do not leak through the wider scope, soft-deleted stay out, and the embedded probe filter really is course-scoped.

**Verification:** hermetic suite 2,403 passed; `ruff` clean; full flock'd local cycle on this head: 89 Playwright journeys passed (13 skipped), oracles 0 findings, integration lane 82 passed (with the function-mode seam env set). CI is green on the new head.

**Follow-up worth filing before beta (not blocking):** by this PR's own numbers roughly 25–40% of Flash-Lite turns end textless. Those now surface as "The tutor was interrupted" instead of a silent duplicate, and a manual retry re-runs the tools, which can double-apply mastery. A continuation run (`agent.run(nudge, message_history=run_result.all_messages())`) would produce the text without re-executing tools.

Issue #554, opened 2026-08-13
quiz H2: mine answers_json into the digest (distractor profile) + digest schema version


AndresL230:
Workstream H (epic #537). **After #544.**

`quiz_attempts.answers_json` records which distractor a student picked, per question, on every submit — and nothing has ever read it back. Joined against `questions_json` it is a per-concept mistake profile.

- Do the join in the **post-submit digest task** (off the request path, one decrypt each), not at generation.
- Feed targeted distractors and a real mistake profile into the next generation.
- Add a **schema version** to the digest so the next key-drift is caught. (The last drift — the coercer reading `common_errors`/`misconceptions` while the agent wrote `common_mistakes` — was fixed in #548; a version field is what stops the third instance.)

Issue #553, opened 2026-08-13
quiz H1: misconceptions tool filters offering_id with the abstract course id — verify live, then fix + seed test


AndresL230:
Workstream H of the pre-revamp quiz batch (epic #537). **Do not start until #544 (F) is merged.**

`read_misconceptions_for_course` filters `offering_concept_stats.offering_id` using the **abstract course id** (`agents/tools/graph_read.py:412`, fed from `routes/quiz.py`'s `node["course_id"]`), while the stats table is keyed on `course_offerings.id`. Near-certainly zero rows, always — so `use_shared_context` has been a no-op.

**Verify against the live DB before acting.** This is one of two findings in the 2026-08-13 data-inventory audit that are code-verified only. Run the filter both ways against staging/prod and confirm the zero-row result before changing anything.

Then: resolve course → active offering inside the tool path (`services/academics.py` owns that resolution), and land a test asserting **non-zero rows on the rich seed** so it can't silently regress to empty again.

Related: the same silent-empty class is what #529 was, and what the F5 helper in #544 is meant to catch generally.

Issue #546, opened 2026-08-12
quiz: flip include_answer_key default to false, then delete it once the #537 client ships


AndresL230:
Follow-up to the Workstream C answer-key deprecation (pre-revamp quiz batch, epic #537).

Today the generate response ships the full answer key to the browser — per-option `correct` booleans (`routes/quiz.py:85-89`). Workstream C adds `include_answer_key` (default `true`) so the current `QuizPanel.tsx` keeps working, and logs every request that uses it.

Once the revamped client ships and grades exclusively through `POST /api/quiz/attempts/{id}/answer`:

1. Flip the `include_answer_key` default to `false`.
2. After a deprecation window with zero logged uses, delete the parameter and the key-mapping code entirely.

Removing the client-side key is a hard requirement of #537 — any client-graded flow leaks every correct answer to devtools before the student answers.

Blocked by: #541 (adds the flag) and the #537 client rewrite (removes the consumer).

AndresL230:
Still blocked, and worth restating now that the pre-revamp batch is finished.

Both halves of this need the #537 client rewrite, which has not started: flipping `include_answer_key` to `false` breaks today's `QuizPanel.tsx`, which still grades client-side from the per-option `correct` booleans, and deleting the parameter needs a deprecation window with zero logged uses after that.

The server-side half it depends on is done — `POST /api/quiz/attempts/{id}/answer` exists (#541), grading is server-authoritative, and #574 now covers that path against a real database. So this is purely waiting on the client.

Issue #543, opened 2026-08-12
quiz E: scoring/generation correctness — mastery-delta config seam, honest delivered_count, wire validation, concurrency tests


AndresL230:
Part of the pre-revamp quiz repair batch under epic #537.

## Verified today

- Mastery delta is hard-coded: `after = clamp01(before + score*0.03 − (total−score)*0.02)` at `routes/quiz.py:456`. A 10-question hard quiz and a 3-question easy one move mastery identically per item.
- Questions whose `correct_answer` doesn't match an option verbatim are dropped and renumbered silently (`routes/quiz.py:66-118`, filter at `routes/quiz.py:279-283`); the client never learns the quiz is short. All-dropped already 502s (`routes/quiz.py:284-290`).
- Wire validation today is only "exactly one correct option": no min-option, duplicate-option, or duplicate-stem checks (the agent schema pins exactly 4 options — `agents/quiz.py:62` — but the route contract allows drift).

## Work

**E1 — configurable mastery model.** Move `0.03`/`0.02` into a named config with a pedagogy docstring. Evaluate difficulty/count scaling in a written options doc. **Numbers do not change in this PR** — the revamp decides. Any later behaviour change must update the `#393` journey's pinned `+0.09` (`frontend/e2e/quiz.spec.ts:41`) deliberately in the same commit.

**E2 — never silently short-change a quiz.** Return `requested_count` and `delivered_count`; consider one bounded top-up retry when drop rate is high. All-dropped stays 502.

**E3 — question validation on the wire mapping.** ≥2 options, no duplicate option text, exactly one correct, no duplicate stems within one attempt.

**E4 — concurrency tests.** Double-submit (atomic claim at `routes/quiz.py:408-415` should hold), double-answer on one index, generate-while-generating on the same concept.

## Acceptance

- Seam landed with numbers byte-identical; options write-up committed for the revamp decision.
- Real-DB tests for E2–E4; `#393` untouched and green.

PR #563, opened 2026-08-14
feat(quiz): question provenance, repetition guard, and the silent-empty seam (#537 addendum Part 2)

AndresL230:
Part 2 of the #537 addendum: E5–E8 and F5–F7, shipped as one PR because
they all land in `routes/quiz.py` and would otherwise conflict.

## E5 — question identity + provenance

A generated question had no identity: it was written into the encrypted
`questions_json` blob, graded, and forgotten. Nothing could ask "have we
asked this before?", "which prompt wrote it?", or "was it grounded in our
materials?".

- `services/quiz_identity.py` — `question_hash`, a stable SHA-256 over the
  normalized stem + option set. Insensitive to whitespace, case and option
  **order**; sensitive to stem and option content; version-tagged so a
  future normalization change is visibly disjoint rather than quietly
  colliding.
- Each stored question now carries `question_hash` plus a `provenance`
  block: `prompt_version` (the system-prompt hash, previously reachable
  only as agent trace metadata), the **served** model, the grounding chunk
  ids, and `rag_grounded`/`catalog`.
- Chunk ids were resolved and dropped on the floor — `_course_material_block`
  returned a bare string. It now returns a `CourseMaterial` record.
  **`match_course_chunks` already returned `id`**, so this was a local
  refactor, not the schema change the brief flagged as a stop-and-report risk.
- Provenance never reaches the client, on **both** response shapes — the
  keyed branch is still the default until #546 flips it, so guarding only
  the keyless allowlist would have shipped chunk ids to every browser today.

Dedupe: the within-attempt check is keyed on `question_hash` as specified,
**and** retains the stem check. The hash covers stem *and* options, so it is
the narrower of the two — a model re-emitting one stem with reworded options
passes it. Dropping the stem check would have quietly narrowed #543's
duplicate-question guard, which E5 has no need to trade away.

## E6 — repetition guard

Past `questions_json` was never re-read, so a student could be served the
same question repeatedly with nothing able to notice.

- `services/quiz_repetition.py` reads the last ~15 distinct items for this
  (student, concept) and generation is told not to repeat them.
- **Fetched raw, not precomputed into the digest** — noted as asked. The
  audit's end-state (precompute in `quiz_context`) belongs with the digest
  schema work in #554; this is bounded hard (6 attempts scanned, 15 stems out)
  in the meantime.
- Deliberately **not** filtered to completed attempts, unlike the history
  tool: a student who generated a quiz and walked away still saw those
  questions.
- Prompt-side guidance only — repeats are **not** hard-dropped. The
  function-mode seam returns the same 3 questions regardless of prompt, so
  dropping repeats would empty every second quiz in the #393 journey and 502
  it. Stems are delimiter-neutralized (#150) on the way back into the prompt.

## E7 — stop dropping `event_type`

Submit computed correct/partial/confusion from the score ratio and discarded
it at the write. Migration `20260814051517` adds a nullable `event_type` to
`node_mastery_events`; `apply_graph_update` persists it and **omits the key
when absent**, so every non-quiz caller keeps working on an environment that
took the code before the DDL. Applied to staging and verified (26 existing
rows keep NULL).

## E8 — grounding is a decision, not an accident

A coverage check runs when retrieval comes back empty, and `quiz.rag_uncovered`
distinguishes three different problems that used to look identical:
`course_unresolved`, `no_chunks_for_course`, `no_match_for_concept`
(plus `coverage_unknown`). **Generation is never blocked on it.**

## F5 — the general fix for silent-empty

`services/tool_signals.py::report_empty_result`. Three personalization inputs
were empty for months because an empty list is exactly what "this student has
nothing yet" looks like. The helper supplies the missing half — whether the
student *plausibly should* have data (enrolled / has attempts / has a graph,
one owner-scoped indexed read, only on the empty path) — and emits
`quiz.tool_empty` when the two disagree. Feature-agnostic so the tutor's tools
use it too; wired into all three quiz read tools. Never raises; a failed probe
means "can't tell", which is silence.

## F6 — measure the prompt before anyone tunes it

`services/prompt_dimensions.py` captures prompt composition per request and
rides `quiz.started`, which shares a `request_id` with the `llm_usage` row.
The load-bearing detail: `digest_present` is only knowable inside an agent
tool running under `asyncio.to_thread`, so the accumulator mutates a shared
dict rather than rebinding a ContextVar — pinned by test.

**The audit's ~2–4k estimate was low.** Measured via `count_tokens`
(`scripts/bench_quiz_prompt_budget.py`, results in `docs/quiz-prompt-budget.md`):

| | measured | audit |
|---|---:|---:|
| System prompt | **1,317** | ~800 |
| `read_concepts_for_user` @ cap | **1,340** | ~250 |
| Today, grounded, 13-concept graph | **3,992** | 2–4k |
| Worst case | **6,839** | — |

The concepts tool at cap costs more than a five-chunk RAG block at typical
chunk sizes — so "COURSE MATERIAL is the dominant variable cost" is only true
for long chunks, and the proposed ~4–5k redesign budget is approximately the
current bill rather than headroom.

## F7 — taxonomy

`quiz.tool_empty` and `quiz.rag_uncovered` added to `EVENT_TAXONOMY`, the
module docstring table, and the pin test.

## Also

- Documented the ciphertext oracle's deliberate omission of `quiz_responses`
  where the manifest lives (owed from the addendum's Part 1, item 3).
- `agents/usage.py::served_model_name` is public and now coerces to `str` —
  the model name flows into `encrypt_json` via provenance, and a non-string
  would have 502'd a generation that had already succeeded.

## Verification

- Hermetic suite: 2103 passed, 9 skipped (was 1997 — +106 tests).
- `ruff check .` clean.
- Full local E2E cycle: Playwright, oracles, integration lane.
- Migration applied to staging before merge.

## Review round (commit 2)

`/code-review high` returned six findings, all valid, all fixed in `cb56970f`:

1. **The F5 probe blocked the event loop** — a sync Supabase read called
   inline from async tool bodies, while every other read in those tools uses
   `to_thread` for exactly that reason. It fires on the *empty* path, which
   today is the common one. Added `report_empty_result_async`.
2. **+3. The probes asked a broader question than the tools did** —
   `HAS_ATTEMPTS` checked all concepts while the tool read one;
   `HAS_GRAPH` checked the whole graph while the read was course-scoped. Both
   would have flagged ordinary progress (first quiz on a new concept; taking
   two courses) as "silently broken" — the precise alarm-fatigue failure F5
   exists to prevent. Probes now take a `scope`.
3b. **`feature` defaulted to `"quiz"`** on a tool the tutor also registers,
   contradicting the contract this PR added to CLAUDE.md. `SaplingDeps`
   carries `feature` now; default `"unknown"`, since wrong attribution is
   worse than absent.
4. **`grounded` was RAG-only but named "any course material"** — a
   catalog-only course persisted every question as ungrounded. Split into
   `rag_grounded` + `catalog`.
5. The E7 comment implied the omit-when-absent trick made the quiz path safe
   pre-migration. It does not, and the failure loses the graded attempt.
   Comment now states the ordering requirement.
6. CLAUDE.md's "exactly two raw `genai.Client` sites" invariant, made false by
   this PR's bench script.

Re-verified after the fixes: hermetic **2103 passed / 9 skipped**, ruff clean,
Playwright **45**, oracles **0 findings**, integration **47 passed**.

The review-fix commit has **not** itself been through a second review round.

## CI fix (commit 3) — a live 405→500 bug, pre-existing on main

`Backend (pytest)` had been red **on main** since the FastAPI 0.138 lock
(`0effc9ee` fails identically), on one test, with

```
AttributeError: '_IncludedRouter' object has no attribute 'path'
```

from otel's FastAPI instrumentation. Not test-only: otel's `_get_route_details`
guards its FULL-match `.path` read with `except AttributeError` but its
PARTIAL-match branch does not — and a PARTIAL match is exactly a wrong-method
request. So the error escaped the middleware and **every 405 returned 500**.
Staging and prod install the same lock, so that was live behaviour.

Nothing to upgrade to: the unguarded line is in every released
`opentelemetry-instrumentation-fastapi` through 0.65b0 (checked against the
published wheels). `services/otel_fastapi_compat.py` wraps the resolver,
absorbing only `AttributeError` and falling back to `scope["path"]` — otel's
own FULL-branch fallback. Reproduced *and* the fix verified at the locked
versions in a scratch env, since the dev venv (fastapi 0.136) cannot
reproduce it.

## Review round 2 (commit 4)

1. `quiz.rag_uncovered` was `category="error"`, but
   `/api/admin/analytics/errors` scans `category = error` newest-first. Firing
   per-generation would have buried `quiz.context_write_failed` and
   `rag.retrieval_failed` — degrading the surface workstream B just repaired.
   Now `usage`, which is also the honest label for a legitimate mode.
2. The misconceptions probe was the one left unscoped. It now asks whether
   aggregates exist for *this student's offerings of this course* — the only
   formulation that detects #553's keyspace mismatch without firing on every
   class that simply has no aggregates yet. Non-owner-scoped probes now refuse
   to run unscoped.
3. The quiz-history probe **could not detect the failure its own comment
   cites**: #529 presents as an empty digest *while attempts exist*, and the
   helper short-circuits on a non-zero count. Split into a digest-keyed check
   plus the attempt-list one.
4. `_course_chunk_coverage` reported a degraded count as `0`, making E8 assert
   "nothing indexed" about a possibly fully-indexed course.
5. A `node_mastery_events` insert failure could **permanently lose a graded
   quiz** (it runs after submit's atomic `completed_at` claim, before score is
   written, unwrapped). Now retries once without `event_type` then degrades,
   loudly logged.

Final: hermetic **2119 passed / 9 skipped**, ruff clean, Playwright **45**,
oracles **0 findings**, integration **47**, all CI checks green.

## Summary by CodeRabbit

- **New Features**
  - Quiz generation now avoids recently served questions and preserves stable question provenance.
  - Added clearer course-material coverage status when grounding information is unavailable.
  - Mastery activity is recorded with more specific event types for quizzes and tutoring.

- **Bug Fixes**
  - Wrong-method requests now correctly return HTTP 405 responses instead of server errors.
  - Quiz generation continues gracefully when course data, history, or mastery-event storage is unavailable.
  - Improved handling of empty search and quiz-history results.


Jose-Gael-Cruz-Lopez:
## Code review — quiz provenance, repetition guard, silent-empty seam

This PR adds question identity + provenance (E5), a recently-asked repetition read (E6), `event_type` persistence on mastery events (E7), grounding-coverage reporting (E8), a generic silent-empty detector (F5), prompt-composition capture (F6), taxonomy entries (F7), and an otel compat shim that fixes a live 405→500. I read every substantive changed file at HEAD rather than the hunks. All three headline goals are genuinely delivered: `question_hash` is a stable, version-tagged SHA-256 over the normalized stem + sorted option set and is stamped on every emitted question (`routes/quiz.py:433`), and provenance is excluded from **both** client shapes — the keyless allowlist and the new `_INTERNAL_QUESTION_KEYS` denylist for the still-default keyed branch (`routes/quiz.py:447-457`); I checked the other three readers of `questions_json` (`get_attempt`, `answer_question`, `submit_quiz`) and nothing leaks. The repetition read is bounded (6 attempts scanned, 15 stems out, deduped by identity) and prompt-side only, which the description states as a deliberate trade — it cannot loop or exhaust a pool. The empty case is not swallowed: `_quiz_via_agent` raises on empty `wire_questions` and `generate_quiz` turns it into a typed 502 plus `quiz.generation_failed`. Every new `except` was checked; none are silent. DB access is entirely through `db/connection.py::table()`, the migration is additive with a timestamp basename per the **Infrastructure** doc's #509 convention, and the new event payloads carry ids/counts/enums only. One P1 blocks merge, plus two P2s worth folding in.

### Findings

**[P1] E7's "only the quiz supplies `event_type`" premise is false — the tutor already supplies it** — `backend/services/graph_service.py:799-803`

```python
        # Omitted rather than written as an explicit null when absent: every
        # non-quiz caller (tutor tools, the document pipeline, manual adds)
        # supplies none, and naming a column PostgREST's schema cache doesn't
        # have is a hard 400 — so omitting keeps THOSE paths working on an
        # environment that took this code before the migration.
```

There are exactly two production producers of `updated_nodes`: `routes/quiz.py::submit_quiz` and `agents/tools/graph.py::update_mastery_tool`. The second is registered on the chat tutor (`agents/chat_tutor.py:162`) with a system prompt that says to call it in *every* turn where the student demonstrates understanding, and it has always passed an `event_type`:

```python
# backend/agents/tools/graph.py:55
    event_type: Literal["interaction", "correction", "quiz"] = Field(
        default="interaction",
        description="Event category for the mastery-event log.",
    )
# backend/agents/tools/graph.py:141
            "event_type": u.event_type,
```

`apply_graph_update` simply discarded it until now, so it never mattered; this PR turns it on for both callers at once. Three consequences: (1) the pre-migration deploy analysis is wrong about scope — every tutor mastery write also 400s and takes the `_insert_mastery_event` retry, a wasted round-trip plus a WARNING per event on the highest-volume writer, not the quiet no-op the comment describes; (2) the column receives two disjoint vocabularies from day one — `correct|partial|confusion` from the quiz, `interaction|correction|quiz` from the tutor — where `interaction` is a schema **default**, i.e. exactly the "default that would make un-categorised events indistinguishable from confident ones" that `20260814051517_node_mastery_events_event_type.sql:12-16` says it is avoiding, and `event_type='quiz'` can only have come from the tutor because a real quiz submit never writes that string; (3) `test_graph_service.py:748-757`'s docstring asserts the same false fact about the codebase and passes only because it calls `apply_graph_update` directly rather than through the tool. Pick one vocabulary (or namespace them) and make the comments and the test match which callers actually classify.

**[P2] `quiz.tool_empty` is `category="error"` and fires once per generation while #553 is live** — `backend/services/tool_signals.py:182-192`

```python
        log_event(
            "quiz.tool_empty",
            category="error",
            user_id=user_id,
            payload={
                "tool": tool,
                "feature": feature,
                "expect": expect.value,
                **(payload or {}),
            },
        )
```

`/api/admin/analytics/errors` scans `category = error` newest-first (`routes/admin_analytics.py:453-456`) — the exact reasoning that made review round 2 re-file `quiz.rag_uncovered` as `usage`. The volume profile is the same here: the quiz system prompt tells the agent to call `read_misconceptions_for_course` on every run (`agents/quiz.py:93`), the tool still passes the abstract course id into an `offering_id` filter (#553, explicitly not fixed here — `agents/tools/graph_read.py:436`), and `COURSE_HAS_AGGREGATES` returns True for any class that has `offering_concept_stats` rows. So every generation by every enrolled student in such a class emits one error-category event plus a WARNING, indefinitely, burying `quiz.context_write_failed` and `rag.retrieval_failed` under routine traffic. If the loudness is intentional it needs its own category or a de-dupe, not the shared error feed.

**[P2] Misconceptions offering resolution runs on every call, not only the empty path** — `backend/agents/tools/graph_read.py:452-460`

```python
    offering_ids: list[str] = []
    if ctx.deps.course_id:
        try:
            offering_ids = await asyncio.to_thread(
                user_offering_ids_for_course, ctx.deps.user_id, ctx.deps.course_id
            )
        except Exception:
            logger.debug("misconceptions probe: offering resolution failed", exc_info=True)
    if offering_ids:
        await report_empty_result_async(
```

The gate is `ctx.deps.course_id`, not `not out`. `user_offering_ids_for_course` (`services/academics.py:185`) is uncached and issues two unbounded PostgREST reads — every `course_offerings` row for the course, then every one of the user's `enrollments`. That is two extra Supabase round-trips on the request path of *every* quiz generation, including the non-empty path, contradicting `tool_signals.py`'s own contract ("one owner-scoped indexed read, only on the empty path") and the PR description's identical claim. Invisible today only because #553 makes `out` always empty; pure waste the moment #553 lands.

**[P3] E8 labels a failed course lookup as `course_unresolved`** — `backend/routes/quiz.py:676-683`

```python
    if material.bu_code is None:
        reason = "course_unresolved"
    elif material.course_chunks is None:
        reason = "coverage_unknown"
    elif material.course_chunks == 0:
        reason = "no_chunks_for_course"
    else:
        reason = "no_match_for_concept"
```

`_resolve_bu_code` (`:506-511`) returns `None` both for "this course has no BU code" and for "the `courses` read threw", and any raise inside `_course_material` degrades to `_EMPTY_MATERIAL` with `bu_code=None` (`:821-826`). All three land on `course_unresolved`. E8 exists to tell different problems apart, and `coverage_unknown` is already the honest can't-tell label.

**[P3] New log lines print the raw `user_id`** — `backend/services/tool_signals.py:177-181`

```python
        logger.warning(
            "%s returned no rows for user=%s despite %s — a personalization "
            "input may be silently broken (F5)",
            tool, user_id, expect.value,
        )
```

Canopy **Engineering Style Guide** §8: "Don't log request/response bodies, user IDs, emails, names, tokens, or decrypted columns." Same in `services/quiz_repetition.py:90-94`. Not a blocker given how widespread this already is in `routes/quiz.py`, but these are new lines and both already carry `request_id` correlation through the event they emit.

### What's good

- `_insert_mastery_event`'s one-shot retry-without-`event_type` is the right shape for the ordering hazard, and the judgement that the journal is not worth a graded attempt is correct — `submit_quiz` really does call `apply_graph_update` after the atomic `completed_at` claim and before the score write.
- Running `_course_material` and `recent_question_identities` concurrently under `asyncio.gather(return_exceptions=True)` with both results individually inspected: a bare gather would have 502'd a quiz over one unreadable past attempt.
- Splitting `grounded` into `rag_grounded` + `catalog`, with `chunk_count` falling back to `len(chunk_ids)`, keeps provenance from asserting something false about a catalog-only course. I confirmed `format_rag_context` returns `""` only for an empty chunk list, so `k_chunks > 0` really does imply RAG text in the prompt.
- Measuring the prompt with `count_tokens` instead of inheriting the audit's estimate, and publishing the caveat that `llm_usage.prompt_tokens` remains the authority on totals.

**Verdict:** request changes — the E7 finding needs resolving before merge; the two P2s are worth folding in while you are in here.

---

Jose-Gael-Cruz-Lopez:
## Review fixes applied

### Blocker
- **`event_type` vocabulary collision.** The claim in `graph_service.py` that "every non-quiz caller supplies none" was false — `agents/tools/graph.py:55` has always sent one from the chat tutor. Both vocabularies are now namespaced and disjoint (`tutor_interaction|tutor_correction|tutor_quiz` vs `quiz_correct|quiz_partial|quiz_confusion`), the tutor field no longer defaults to a real category (omitted when `None`, mirroring `apply_graph_update`), and the migration + service comments now document the real six-value set and the real pre-migration blast radius. New tests drive the tutor path end to end.

### Major
- `quiz.tool_empty` moved from `category="error"` to `"usage"` — it was firing once per generation into the feed `/api/admin/analytics/errors` scans, burying `quiz.context_write_failed` and `rag.retrieval_failed`.
- Misconceptions probe now gates on `not out`, so `user_offering_ids_for_course` (uncached, two unbounded reads) no longer runs on the non-empty path of every quiz generation.

### Minor / nits
- E8 tells a failed `courses` read apart from "no BU code" (`coverage_unknown`, not `course_unresolved`), via a new tri-state lookup.
- Bench script's raw `google.genai.Client` is now behind a `model_mode()` gate and uses `model_name_for("quiz")`; CLAUDE.md's raw-client inventory corrected (it was already wrong — a fourth site existed).
- `prompt_dimensions.snapshot()` deep-copies; `tool_signals` probe failure logs at `warning`; `k_chunks` pinned explicitly in the grounded fixture; explicit-`None` `event_type` case covered; stale "schema has no event_type column" comment fixed; doc fence language + F6 dimension list corrected; raw `user_id` removed from new log lines; duplicated `sink` fixture moved to `conftest.py`.

**Verification** — `ruff check .` clean · **2093 passed, 56 skipped**

Jose-Gael-Cruz-Lopez:

### 1. Ordering is load-bearing — migration BEFORE code

This PR's own comment says it, and the review confirmed the failure mode: `submit_quiz` calls `apply_graph_update` **after** the atomic `completed_at` claim and **before** the score write, so a pre-migration insert failure loses a graded attempt. `_insert_mastery_event` retries once without `event_type` and degrades loudly, but that is a safety net, not a licence to deploy first.

```sql
-- Must return one row, is_nullable = YES, before any code ships
SELECT column_name, data_type, is_nullable
FROM information_schema.columns
WHERE table_name = 'node_mastery_events' AND column_name = 'event_type';
```

### 2. Existing rows must be untouched

```sql
SELECT event_type, count(*) FROM node_mastery_events GROUP BY 1 ORDER BY 2 DESC;
```

Before deploy: expect a single `NULL` bucket (the PR reports 26 rows on staging). Nothing should have a value yet.

### 3. After deploy — the namespacing must hold

The fix in this PR namespaced the two producers because the tutor's `update_mastery_tool` has *always* sent an `event_type` (`agents/tools/graph.py`), it was simply discarded until now. Re-run the same query after traffic:

- Expected values: `NULL`, `tutor_interaction`, `tutor_correction`, `tutor_quiz`, `quiz_correct`, `quiz_partial`, `quiz_confusion`.
- **Any bare `interaction` / `correction` / `quiz` / `correct` / `partial` / `confusion` means a writer bypassed the namespacing** — that is the exact ambiguity this change exists to prevent, and it should be investigated rather than accepted.

### 4. Volume expectation

The tutor is the higher-volume writer of the two, not the quiz. If `tutor_*` rows are absent after real tutor traffic, `update_mastery_tool` is silently not persisting and the omit-when-`None` path is swallowing it.

AndresL230:
## Third review round + the live-DB checks

Picking this back up. Two things were outstanding: Jose's three fix commits had never themselves been reviewed, and the pre-merge database checks were left unrun because that environment had no credentials.

### Live DB — both pre-merge checks pass on staging

Run through the session-mode pooler (`scripts/pooler_url.py`; the `.env.staging` value is the IPv6-only direct host and is unreachable from here):

| check | result |
|---|---|
| `event_type` column | one row, `text`, `is_nullable = YES`, no default |
| existing rows | single bucket: **26 × NULL** |
| ledger | migration recorded |

So the migration is applied **strictly before** this code ships, which is the ordering the E7 comment calls load-bearing. Checks 3 and 4 (namespacing holds, `tutor_*` rows appear) are post-traffic and stay open as a post-deploy watch.

I also diffed the edited migration: the header comments were rewritten, but the DDL is byte-identical (`ADD COLUMN IF NOT EXISTS event_type TEXT`), and the ledger keys on basename — no immutability violation, nothing re-runs.

### Review of the fix commits — `affde139`

No correctness bugs. Four low findings, all fixed:

1. **`prompt_dimensions.snapshot()` lost the whole payload on one bad value.** The deep copy sits inside a `try` whose `except` returned `{}`, so a single un-deepcopyable value shipped `quiz.started` with *no* dimensions — measuring nothing while looking like a healthy event, which is the exact bug class F6 exists to end. Degrades to a shallow copy now, and warns. Its docstring's rationale was also just wrong (nothing mutates a recorded list in place), so that's corrected rather than left as a false comment.
2. **E8 called a failed retrieval `no_match_for_concept`.** `retrieve_chunks` swallows its own failures and returns `[]` — the same value a clean miss returns. So a course *with* material indexed whose retrieval broke was recorded as "it has material, none of it covers this concept": a claim about data we never read, which is precisely what the reason taxonomy exists to stop. Added `retrieve_chunks_detailed`, which says whether the empty result is a fault or a fact; a fault now reports `coverage_unknown`. The #439 seam skip is deliberately **not** a fault — otherwise every function-mode E2E run would report broken retrieval. `retrieve_chunks` keeps its list contract, so `learn.py` and `benchmark_quiz.py` are untouched.
3. Bench script promised a keyless run "fails loudly here" but still fell back to a dummy key, dying later inside `count_tokens` on an opaque auth error. It now fails where it says it does.
4. `EVENT_TAXONOMY`'s `quiz.started` row and `docs/quiz-prompt-budget.md` disagreed about which dimensions come from the route vs. only when the agent calls the recording tool. Reconciled.

Both behavioural fixes are pinned by tests written to fail first.

### Verification

Hermetic **2128 passed / 9 skipped** (+2), `ruff` clean, oracles **0 findings**, integration **47 passed**, Playwright **47 passed**.

The one Playwright failure is `landing-drag-field.spec.ts:332`, and it is **not from this PR** — it fails identically on `main` (same line, all three retries) and has since the #524 landing-v5 merge on Aug 16, which is when the e2e browser lane on main went red. This PR touches no frontend code. Filing that separately.

AndresL230:
Merged. Filed the unrelated Playwright failure as #566 — the landing drag-field scroll-follow assertion, red on `main` since #524, which is what has kept the e2e browser lane red there.

Remaining from the addendum: Workstream H (#553–#557), plus #545 and #546.

Post-deploy watch still owed on this one — checks 3 and 4 from the DB comment above (namespacing holds; `tutor_*` rows actually appear once real tutor traffic lands). Nothing to run until it deploys.

